PR #29, opened 2021-09-03
Fix snowflake ID in dbt extractor by adding [ch3428]

alyiwang:
### What?

- Add account into dataset id in DBT extractor 

### Checklist

- [ ] I have tested that the changes in this PR work as expected
- [x] I have added/updated tests that exercise the critical code paths in this diff

### DIFF
--- a/metaphor/dbt/extractor.py
+++ b/metaphor/dbt/extractor.py
@@ -40,7 +40,7 @@ class DbtRunConfig(RunConfig):
     catalog: str
 
     # the database service account this DBT project is connected to
-    account: str = None
+    account: Optional[str] = None
 
 
 class DbtExtractor(BaseExtractor):
--- a/metaphor/dbt/extractor.py
+++ b/metaphor/dbt/extractor.py
@@ -39,6 +39,9 @@ class DbtRunConfig(RunConfig):
     manifest: str
     catalog: str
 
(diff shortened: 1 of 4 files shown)


mars-lan:
Fix title?

metaphor/dbt/extractor.py
```diff
@@ -39,6 +39,9 @@ class DbtRunConfig(RunConfig):
     manifest: str
     catalog: str
 
+    # the database service account this DBT project is connected to
+    account: str = None
```
mars-lan, line 43: This is easier but theoretically you can get it directly from `profile.yml`: https://github.com/MetaphorData/dbt/blob/main/profiles.yml#L5

mars-lan:
> Fix title?

Never mind. I misread it.

metaphor/dbt/extractor.py
alyiwang: Right, the `profile` has all those information. But that way we need to bring `profile.yml` to the crawlers, besides the current manifest and catalog. Also, from the customer's perspective, they might prefer not to pass the `profile` as it contains passwords which the crawler doesn't need.

mars-lan: Makes sense.

mars-lan (approved):

PR #33, opened 2021-09-20
Get table statistics in postgres and snowflake [sc-604]

alyiwang:
### What?

- Fetch table statistics for postgres and redshift
- Fix: MB = 10^6
- Tested against local postgres

### Checklist

### DIFF
--- a/metaphor/snowflake/extractor.py
+++ b/metaphor/snowflake/extractor.py
@@ -195,7 +195,7 @@ class SnowflakeExtractor(BaseExtractor):
         if row_count:
             dataset.statistics.record_count = float(row_count)
         if table_bytes:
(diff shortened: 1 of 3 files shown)


metaphor/snowflake/extractor.py
```diff
@@ -195,7 +195,7 @@ def _init_dataset(
         if row_count:
             dataset.statistics.record_count = float(row_count)
         if table_bytes:
-            dataset.statistics.data_size = table_bytes / 1048576
+            dataset.statistics.data_size = table_bytes / 1000000
```
mars-lan, line 198: Consider this for clarify

```suggestion
            dataset.statistics.data_size = table_bytes / (1000*1000)  # in MB
```

PR #38, opened 2021-09-28
Use default placeholder value when column native type is not set

mars-lan:
### Why?

dbt's catalog may sometimes contain `null` as column type. This causes an issue when ingesting the MCEs which expects `nativeType` to be an non-empty string.

### What?

As titled.

### Checklist


alyiwang (approved):
LGTM

PR #42, opened 2021-10-01
Refactor dbt connector to use classes generated from the official JSON schemas

mars-lan:
### Why?

dbt publishes the official JSON schemas for manifest.json & catalog.json file. We should use these to generate data model classes instead of hand-crafted versions. 

### What?

- Replace hand-crafted dbt_model.py with [pydantic](https://pydantic-docs.helpmanual.io/) versions generated by [datamodel-code-generator](https://github.com/koxudaxi/datamodel-code-generator)
- Update extractor code to make use of the new models and fix a few bugs as a result
- Fix all incorrect assumptions in the test data to match the JSON schemas

Also manually tested against complex dbt projects to ensure the correctness.

### Checklist


alyiwang (approved):
LGTM, great work!

mars-lan:
> Hum CI failed for an unrelated test and can't seem to reproduce it locally either. Need to investigate more...

Turned out that the test was broken by `sql-metadata` 2.3.0. Pined it down to `2.2.2` for now to unblock this PR.

alyiwang (approved):

PR #43, opened 2021-10-02
update looker crawler to output looker explores

alyiwang:
### Why?

We have the new VirtualView models, time to fill in the data

### What?

- Update Looker connector to fetch explores and views
- Update dashboard upstream to also point to explores

### Checklist

- [x] I have added/updated tests that exercise the critical code paths in this diff

### DIFF
--- a/metaphor/common/entity_id.py
+++ b/metaphor/common/entity_id.py
@@ -8,7 +8,9 @@ from metaphor.models.metadata_change_event import (
     DatasetLogicalID,
     EntityType,
     GroupID,
+    KnowledgeCardLogicalID,
     PersonLogicalID,
+    VirtualViewLogicalID,
 )
 
 from metaphor.common.event_util import EventUtil
@@ -17,7 +19,14 @@ from metaphor.common.event_util import EventUtil
 @dataclass
 class EntityId:
     type: EntityType
-    logicalId: Union[DatasetLogicalID, DashboardLogicalID, PersonLogicalID, GroupID]
+    logicalId: Union[
+        DatasetLogicalID,
+        DashboardLogicalID,
+        GroupID,
+        KnowledgeCardLogicalID,
+        PersonLogicalID,
+        VirtualViewLogicalID,
+    ]
 
     def __str__(self) -> str:
         json = encode_canonical_json(EventUtil.clean_nones(self.logicalId.to_dict()))
--- a/metaphor/looker/lookml_parser.py
+++ b/metaphor/looker/lookml_parser.py
@@ -128,6 +128,9 @@ def _to_dataset_id(source_name: str, connection: Connection) -> EntityId:
 
 
 def to_virtual_view_id(name: str, virtualViewType: VirtualViewType) -> EntityId:
+    """
+    converts a virtual view name and type into an Virtual View entity ID
+    """
     return EntityId(
         EntityType.VIRTUAL_VIEW,
         VirtualViewLogicalID(name=name, type=virtualViewType),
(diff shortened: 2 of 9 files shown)


metaphor/common/entity_id.py
```diff
@@ -17,7 +19,14 @@
 @dataclass
 class EntityId:
```
mars-lan, line 20: We should probably generate this from models eventually.

metaphor/looker/lookml_parser.py
```diff
@@ -211,14 +311,28 @@ def _load_model(
     if connection is None:
         raise ValueError(f"Model ${model_path} has an invalid connection")
 
-    return Model.from_dict(raw_views, raw_explores, connection)
-
+    return raw_views, raw_explores, connection
 
-def parse_models(base_dir: str, connections: Dict[str, Connection]) -> Dict[str, Model]:
 
+def parse_models(
```
mars-lan, line 317: Maybe rename this to `parse_project` since it's now returning more than just models? Please also add one line description to all public methods that I missed previously. Thanks.

tests/looker/test_lookml_parser.py
```diff
@@ -53,10 +64,36 @@ def test_basic(test_root_dir):
...
+                ],
+                source_datasets=["DATASET~5881AB4C0A42EF1C15F6C02C0D14AD43"],
+            ),
+        ),
+        VirtualView(
+            logical_id=VirtualViewLogicalID(
+                name="explore1", type=VirtualViewType.LOOKER_EXPLORE
+            ),
+            looker_explore=LookerExplore(
+                base_view="VIRTUAL_VIEW~361092A8CFF1EFEFD695B221B21943B3",
```
mars-lan, line 85: Can a Looker explore have only one base view? Is it possible to join mulitple views to form a new explore?

mars-lan: nvm. I see the `joins` field now.

tests/looker/test_lookml_parser.py
```diff
@@ -89,10 +126,62 @@ def test_join(test_root_dir):
...
+        VirtualView(
+            logical_id=VirtualViewLogicalID(
+                name="view1", type=VirtualViewType.LOOKER_VIEW
+            ),
+            looker_view=LookerView(
+                dimensions=[LookerViewDimension(data_type="string", field="country")],
+                measures=[
+                    LookerViewMeasure(field="average_measurement", type="average")
+                ],
+                source_datasets=["DATASET~5881AB4C0A42EF1C15F6C02C0D14AD43"],
```
mars-lan, line 139: Consider building the entity ID from its parts to make the test more readable.

metaphor/looker/lookml_parser.py
mars-lan: Singular? https://docs.looker.com/data-modeling/getting-started/how-project-works

metaphor/looker/lookml_parser.py
```diff
@@ -164,6 +97,175 @@ def from_dict(
...
+    raw_view = raw_views.get(view_name)
+    if raw_view is None:
+        raise ValueError(f"Refer to non-existing view {view_name}")
+
+    if "derived_table" in raw_view:
+        sql = raw_view["derived_table"].get("sql")
+        if sql is not None:
+            return _extract_upstream_datasets_from_sql(sql, raw_views, connection)
+
+        # TODO(1329): Add support for native derived tables
```
mars-lan, line 155: Add `SC` prefix to ticket

metaphor/looker/lookml_parser.py
```diff
@@ -164,6 +97,175 @@ def from_dict(
...
+        label=raw_view.get("label", None),
+        source_datasets=[
+            str(ds) for ds in _get_upstream_datasets(name, raw_views, connection)
+        ],
+    )
+
+    if "extends" in raw_view:
+        view.extends = [
+            str(to_virtual_view_id(view, VirtualViewType.LOOKER_VIEW))
+            for view in raw_view["extends"]
```
mars-lan, line 198: A more pythonic way is to drop the `if` and change this line to

```suggestion
            for view in raw_view.get("extends", [])
```

mars-lan: It's not entirely equivalent as it'll set the `view.extends` field even if `raw_view` has no `extends`.

mars-lan:
Please add https://docs.looker.com/data-modeling/getting-started/how-project-works to the description at the top for future reference.

alyiwang:
> Please add https://docs.looker.com/data-modeling/getting-started/how-project-works to the description at the top for future reference.

done

metaphor/looker/lookml_parser.py
alyiwang: This does reduce one line 👍 . But personally still prefer not to the set the field if there is no entry, more align with LookerML format.

mars-lan: SGTM

mars-lan (approved):

PR #45, opened 2021-10-05
Update Snowflake user creation instructions

mars-lan:
### Why?

The original instructions were incomplete and incorrect.

### What?

As titled. Tested by manually creating the user and run the connector using the new credential.

### Checklist


alyiwang (approved):
Thanks for updating this!

Issue #13, opened 2021-08-13
Breaking changes in 0.6


mars-lan:
- Rename `PostgresqlExtractor` & `PostgresqlRunConfig` to `PostgreSQLExtractor` & `PostgreSQLRunConfig`
- Replace single db filtering with multiple for Snowflake connector
- Drop dataset/virtual view dual writing in Looker connector
- Drop dual writing in dbt connector

PR #46, opened 2021-10-05
Limit Snowflake connector to extract metadata only from the database specified in the config

mars-lan:
### Why?

Snowflake connector currently will scan all tables/views from all databases regardless what's specified in the config.

### What?

As titled. We can introduce a list of allowed databases in the config file as a breaking change in 0.6:  https://github.com/MetaphorData/connectors/issues/13

Note: While the behavior is backward incompatible, it's considered as a bug fix. 

Verified by running it locally:
```
$ python -m metaphor.snowflake ~/Desktop/local/snowflake.yml
2021-10-04 21:18:57,948 INFO - query: [USE ACME]
...
2021-10-04 21:19:31,522 INFO - Fetched 43 entities
```

### Checklist

### DIFF
--- a/metaphor/snowflake/extractor.py
+++ b/metaphor/snowflake/extractor.py
@@ -64,19 +64,14 @@ class SnowflakeExtractor(BaseExtractor):
         with ctx:
             cursor = ctx.cursor()
 
-            databases = self._fetch_databases(cursor, config.database)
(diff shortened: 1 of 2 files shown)


alyiwang (approved):
LGTM

PR #68, opened 2021-10-21
Use falsy instead of None when building url

alyiwang:
### Why?

If strictly checking `None` value, then the default value `""` passed in by the looker-action won't be treated as falsy and will generate invalid URL. 

### 🤓 What?

- Use falsy check to replace none check. 

### 🧪 Tested?

Unit test

### DIFF
--- a/tests/looker/test_lookml_parser.py
+++ b/tests/looker/test_lookml_parser.py
@@ -97,7 +97,7 @@ def test_basic(test_root_dir):
 
 def test_join(test_root_dir):
     models_map, virtual_views = parse_project(
-        test_root_dir + "/looker/join", connection_map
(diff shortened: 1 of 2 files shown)


tests/looker/test_lookml_parser.py
```diff
@@ -97,7 +97,7 @@ def test_basic(test_root_dir):
 
 def test_join(test_root_dir):
     models_map, virtual_views = parse_project(
-        test_root_dir + "/looker/join", connection_map
+        test_root_dir + "/looker/join", connection_map, ""
```
mars-lan, line 100: Prefer f-string over `+`

mars-lan (commented):
Otherwise LGTM

PR #89, opened 2021-10-28
Add a config to write all logs as a zip file

mars-lan:
### 🤔 Why?

To help debug any potential issues during the run.

### 🤓 What?

All logs are written to a temp location before moving to the final destination or upload to s3 bucket as a zip.
Also need to fix the AWS role assuming code to make smart_open to actually make use the assumed credential.

### 🧪 Tested?

Manually tested end-to-end to make sure that it writes but locally and to S3. 

Screen Shot 2021-10-28 at 12 34 23 PM

### DIFF
--- a/metaphor/common/file_sink.py
+++ b/metaphor/common/file_sink.py
@@ -74,7 +74,7 @@ class FileSink(Sink):
 
         prefix = self.path[0:-5]
 
-        zip_file = tempfile.mktemp(suffix=".zip")
+        _, zip_file = tempfile.mkstemp(suffix=".zip")
         arcname = f"{path.basename(prefix)}.log"
         with ZipFile(zip_file, "w", ZIP_DEFLATED) as file:
             file.write(LOG_FILE, arcname=arcname)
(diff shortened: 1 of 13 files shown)


metaphor/common/file_sink.py
```diff
@@ -50,6 +60,26 @@ def _sink(self, messages: List[dict]) -> bool:
...
+        prefix = self.path[0:-5]
+
+        _, zip_file = tempfile.mkstemp(suffix=".zip")
+        arcname = f"{path.basename(prefix)}.log"
+        with ZipFile(zip_file, "w", ZIP_DEFLATED) as file:
+            file.write(LOG_FILE, arcname=arcname)
+
+        with open(zip_file, "rb") as file:
+            write_file(
+                f"{prefix}-log.zip", file.read(), True, s3_session=self.s3_session
```
alyiwang, line 84: Here it's prefix-log.zip, but in your test example, it's bigquery.zip ? Would prefer with `-log`

alyiwang (approved):
Otherwise, LGTM
This is great for debugging!

metaphor/common/file_sink.py
mars-lan: Yeah sorry the screen capture was done before this change.

Issue #76, opened 2021-10-26
Breaking changes in 0.8


mars-lan:
- Unified database/schema/table filter in config files (https://app.shortcut.com/metaphor-data/story/4106)
- Split off redshift to its own connector
- Rename  config `exclude_views` to `include_views` for `snowflake.usage` connector
- Rename config `projectSourceUrl` to `project_source_url` for `looker` & `dbt` connectors

PR #91, opened 2021-10-29
Rename config key for consistency

mars-lan:
### 🤔 Why?

As part of https://github.com/MetaphorData/connectors/issues/76

### 🤓 What?

As titled. Not bumping PyPI version on purpose.

### 🧪 Tested?

Added unit test.

### DIFF
--- a/metaphor/looker/config.py
+++ b/metaphor/looker/config.py
@@ -39,7 +39,7 @@ class LookerRunConfig(RunConfig):
     connections: Dict[str, LookerConnectionConfig]
 
     # Source code URL for the project directory
-    projectSourceUrl: Optional[str] = None
+    project_source_url: Optional[str] = None
 
     verify_ssl: bool = True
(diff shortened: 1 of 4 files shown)


metaphor/looker/config.py
```diff
@@ -39,7 +39,7 @@ class LookerRunConfig(RunConfig):
     connections: Dict[str, LookerConnectionConfig]
 
     # Source code URL for the project directory
-    projectSourceUrl: Optional[str] = None
+    project_source_url: Optional[str] = None
```
alyiwang, line 42: The DBT config has the same key that need to be changed too.

mars-lan: SGTM. Will do it in a follow-up PR as it involved refactoring out the config class.

alyiwang: sounds good

alyiwang (approved):

PR #92, opened 2021-10-29
Split redshift connector into its own class

mars-lan:
### 🤔 Why?

As part of https://github.com/MetaphorData/connectors/issues/76. This will allow us to add redshift-specific code easily in the future.

### 🤓 What?

Create a separate redshift connection, which for now simply inherits pretty much everything from postgresql connector.

### 🧪 Tested?

Against local postgres:

```
$ python -m metaphor.postgresql ~/Desktop/local/postgres.yml 
...
2021-10-29 09:04:15,395 INFO - Fetched 1 entities
2021-10-29 09:04:15,413 INFO - Extractor ended successfully
```


alyiwang (dismissed):
This is good, it will be easier for me to do https://app.shortcut.com/metaphor-data/story/3610/use-svv-table-info-to-get-table-size-for-redshift and https://app.shortcut.com/metaphor-data/story/3968/add-redshift-usage-crawler , not that I don't need to worry to both cases working.

alyiwang (approved):

PR #98, opened 2021-11-01
Capture field length for Redshift Postgres [sc-3506]

alyiwang:
### 🤔 Why?

Surface more column info for user

### 🤓 What?

- Fill in column data size and precision
- Rewrite the fetch_column and fetch_constraints to issue much less queries and improve speed.

### 🧪 Tested?

Local crawler run, diff with previous result.
The MCE different is only in the max_length and precision fields and timestamps.

### DIFF
--- a/metaphor/postgresql/extractor.py
+++ b/metaphor/postgresql/extractor.py
@@ -239,12 +239,18 @@ class PostgreSQLExtractor(BaseExtractor):
 
     @staticmethod
     def _build_field(column) -> SchemaField:
-        field = SchemaField()
-        field.field_path = column["column_name"]
-        field.native_type = column["data_type"]
-        field.nullable = column["is_nullable"] == "YES"
-        field.description = column["description"]
-        return field
+        return SchemaField(
+            field_path=column["column_name"],
+            native_type=column["data_type"],
+            nullable=column["is_nullable"] == "YES",
+            description=column["description"],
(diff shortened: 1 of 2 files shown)


metaphor/postgresql/extractor.py
```diff
@@ -71,12 +73,10 @@ async def _extract(
...
-                # TODO: parallel fetching
-                for schema, name, fullname in tables:
-                    dataset = self._datasets[fullname]
-                    await self._fetch_columns(conn, schema, name, dataset)
-                    if not redshift:
-                        await self._fetch_constraints(conn, schema, name, dataset)
+                await self._fetch_columns(conn, db)
+
+                if not redshift:
+                    await self._fetch_constraints(conn, db)
```
mars-lan, line 79: Consider implementing redshift-specific `_fetch_constraints` in `RedshiftConnector` directly?

alyiwang: There are several places that are postgres-only or need to implement for redshift. I will duplicate the crawler code and do the implementation. Probably in https://app.shortcut.com/metaphor-data/story/3610/use-svv-table-info-to-get-table-size-for-redshift
This PR is for adding field length and some no-change refactoring. Maybe no expand the scope too much.

mars-lan (approved):

PR #99, opened 2021-11-04
Improve Snowflake usage crawler [sc-4639]

alyiwang:
### 🤔 Why?

Usage crawler is taking too long to parse large number of SQL queries

### 🤓 What?

- Use snowflake access_history to get the query table and columns, instead of parsing SQL
- NOTE: since the query joins access_history table with query_history table, each query can be slow (few seconds for our case). Running small batch with many concurrency actually makes it slower, because each query will do the joining computation again. Prefer large batch size with fewer batches, as it's much faster.
- NOTE 2: The table joining stats is not available, as there is no joining and columns info in the access history

### 🧪 Tested?

local test run with 30 days lookback on DEMO_DB and ACME, took 25-30s, mostly DB query time

### DIFF
--- a/metaphor/snowflake/usage/README.md
+++ b/metaphor/snowflake/usage/README.md
@@ -1,6 +1,6 @@
 # Snowflake Usage Statistics Connector
 
-This connector extracts usage statistics from a Snowflake account using [Snowflake Connector](https://docs.snowflake.com/en/user-guide/python-connector.html).
+This connector extracts usage statistics from a Snowflake account using [Snowflake Connector](https://docs.snowflake.com/en/user-guide/python-connector.html). It can only be run against Snowflake Enterprise account or above, and doesn't support regular account yet.
 
 ## Setup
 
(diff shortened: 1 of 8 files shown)


mars-lan (commented):
Great work. Please update README with notes on needing the enterprise edition.

alyiwang:
> Great work. Please update README with notes on needing the enterprise edition.

Updated

metaphor/snowflake/usage/README.md
```diff
@@ -1,6 +1,6 @@
 # Snowflake Usage Statistics Connector
 
-This connector extracts usage statistics from a Snowflake account using [Snowflake Connector](https://docs.snowflake.com/en/user-guide/python-connector.html).
+This connector extracts usage statistics from a Snowflake account using [Snowflake Connector](https://docs.snowflake.com/en/user-guide/python-connector.html). It can only be run against Snowflake Enterprise account or above, and doesn't support regular account yet.
```
mars-lan, line 3:
```suggestion
This connector extracts usage statistics from a Snowflake account using [Snowflake Connector](https://docs.snowflake.com/en/user-guide/python-connector.html). It queries [ACCESS_HISTORY](https://docs.snowflake.com/en/sql-reference/account-usage/access_history.html) and therefore requires Snowflake Enterprise or higher.
```

mars-lan (commented):
Otherwise LGTM

mars-lan (approved):

PR #101, opened 2021-11-05
Automatically tag commit after publishing a new version to PyPI

mars-lan:
### 🤔 Why?

To help identify the commit that's associated with a particular PyPI version. This will also pave the way for automatic release note generation in the future.

### 🤓 What?

As titled.

### 🧪 Tested?

Since we don't have an existing `workflow_dispatch`, this can only be tested after merging.


alyiwang (approved):
Let's give it a try

PR #102, opened 2021-11-09
Use SVV TABLE INFO to get table size for [sc-3610]

alyiwang:
### 🤔 Why?

The table size and row count doesn't work for redshift if using the postgres crawler, the results are always 0

### 🤓 What?

Using the SSV TABLE INFO to get those information https://docs.aws.amazon.com/redshift/latest/dg/r_SVV_TABLE_INFO.html

### 🧪 Tested?

Tested locally with tunnel to data env, both postgres and redshift crawler run successfully, result json passed eye examination.

### DIFF
--- a/metaphor/redshift/extractor.py
+++ b/metaphor/redshift/extractor.py
@@ -17,6 +17,10 @@ class RedshiftExtractor(PostgreSQLExtractor):
     def config_class():
         return RedshiftRunConfig
 
+    def __init__(self):
+        super().__init__()
+        self._platform = DataPlatform.REDSHIFT
+
     async def extract(self, config: RedshiftRunConfig) -> List[MetadataChangeEvent]:
         assert isinstance(config, PostgreSQLExtractor.config_class())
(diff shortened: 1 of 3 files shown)


metaphor/redshift/extractor.py
```diff
@@ -14,4 +18,42 @@ def config_class():
...
-        return await self._extract(config, True)
+        assert isinstance(config, PostgreSQLExtractor.config_class())
+        logger.info(f"Fetching metadata from redshift host {config.host}")
+
+        databases = await self._fetch_databases(config)
+
+        for db in databases:
+            conn = await self._connect_database(config, db)
+            try:
+                await self._fetch_tables(conn, DataPlatform.REDSHIFT)
```
mars-lan, line 29: Instead of passing the platform in, why not add a `get_platform` method or simpy declare the platform as a class variable?

alyiwang: updated

mars-lan (approved):

PR #104, opened 2021-11-11
Handle different versions of dbt manifest [sc-4531]

alyiwang:
### 🤔 Why?

different dbt versions produces use slightly different manifest schema.

### 🤓 What?

- update the imported schema to manifest v3
- test and fix issues to make parser compatible with older version of schema

### 🧪 Tested?

tested using manifest v1, v2, v3. The produced results are identical except some dbt generated timestamp, ids.


mars-lan (commented):
Great work. Please also update https://github.com/MetaphorData/connectors/blob/main/metaphor/dbt/gen_models.sh

alyiwang:
> Great work. Please also update https://github.com/MetaphorData/connectors/blob/main/metaphor/dbt/gen_models.sh

updated

metaphor/dbt/gen_models.sh
```diff
@@ -2,22 +2,52 @@
...
+CATALOG_V1_URL=https://schemas.getdbt.com/dbt/catalog/v1.json
+
 OUTPUT_DIR=generated
+MANIFEST_V1_FILE=$OUTPUT_DIR/dbt_manifest_v1.py
+MANIFEST_V2_FILE=$OUTPUT_DIR/dbt_manifest_v2.py
+MANIFEST_V3_FILE=$OUTPUT_DIR/dbt_manifest_v3.py
+MANIFEST_V4_FILE=$OUTPUT_DIR/dbt_manifest_v4.py
+CATALOG_V1_FILE=$OUTPUT_DIR/dbt_catalog_v1.py
+
+poetry run datamodel-codegen \
```
mars-lan, line 18: Should probably make this a function rather than copy & paste code?

metaphor/dbt/extractor.py
```diff
@@ -34,20 +34,29 @@
...
     ParsedMacro,
+    ParsedModelNode,
+    ParsedSchemaTestNode,
     ParsedSourceDefinition,
 )
 
 logger = get_logger(__name__)
 
 
+MODEL_NODE_TYPE = Union[CompiledModelNode, ParsedModelNode]
```
mars-lan, line 51: Add comments on how this is mapped to different versions.

metaphor/dbt/gen_models.sh
alyiwang: done

metaphor/dbt/extractor.py
alyiwang: I believe those two types are not version dependent, but rather different command or environment generate different node type. Most cases it's "Compiled" Node, which has "compile_sql" field, but sometimes (I can't remember how exactly I got this) it is "Parsed" node without this field. 
Added a comment in code

mars-lan (approved):
Great work!

PR #107, opened 2021-11-16
not all arguments converted during string [sc-4942]

alyiwang:
### 🤔 Why?

The logging syntax used is incorrect. `logger.error` doesn't take Exception object as second param

### 🤓 What?

- Split the error trace logging into a new logger call
- Silence the snowflake connector logging as it logs each query. 

### 🧪 Tested?

Locally tested the logging works


mars-lan (approved):
Thanks for the quick fix

alyiwang:
> Thanks for the quick fix

This is the fix for the logging, we will see the real error after this is deployed.

PR #109, opened 2021-12-14
Format extractor run duration

mars-lan:
### 🤔 Why?

It's currently printing the full precision, which makes it difficult to read, .e.g.

```
Fetched 0 entities, took 1235.893330335617 s
```

### 🤓 What?

As titled.

### 🧪 Tested?

```
Fetched 0 entities, took 1235.9 s
```


alyiwang (approved):
LGTM

PR #110, opened 2021-12-16
Add README files for PostgreSQL & Redshift connectors

mars-lan:
### 🤔 Why?

The instructions on how to use PostgreSQL & Redshift connectors are currently missing.

### 🤓 What?

As titled. Also add a new `redshift` extra in case if its dependencies diverge from that of `postgresql` in the future.

### 🧪 Tested?

N/A


alyiwang (approved):
Thanks for adding

PR #112, opened 2022-01-02
Rename logging.py to logger.py to avoid false-positive LGTM alerts

mars-lan:
### 🤔 Why?

LGTM is incorrectly flagging that `logging` is imported twice using different syntax: https://lgtm.com/projects/g/MetaphorData/connectors/snapshot/efb7c617bd85f0c8e49638468c26b70dc983eef3/files/metaphor/common/file_sink.py?sort=name&dir=ASC&mode=heatmap#x852603541621ff8c:1

### 🤓 What?

As titled.

### 🧪 Tested?

Pre-commit checks.


alyiwang (approved):
LGTM

PR #114, opened 2022-01-05
Ensure dbt crawler outputs unique entity IDs [sc-5857]

elic-eon:
### 🤔 Why?

To fix the crawler output. Currently, the crawler could output duplicate entity ID in `sourceModels`, `sourceDatasets` or `owner` depending on the dbt `manifest.json`, which will break the uniqueness validation in Metaphor.

### 🤓 What?

- Uniquify entity ID preserving order.

### 🧪 Tested?

Uniquify `tests/dbt/data/shopify/results.json` manually and run tests successfully.

### DIFF
--- a/metaphor/dbt/extractor.py
+++ b/metaphor/dbt/extractor.py
@@ -325,17 +325,21 @@ class DbtExtractor(BaseExtractor):
 
             if model.depends_on is not None:
                 if model.depends_on.nodes:
-                    dbt_model.source_models = [
-                        self._get_virtual_view_id(self._virtual_views[n].logical_id)
-                        for n in model.depends_on.nodes
-                        if n.startswith("model.")
-                    ]
+                    dbt_model.source_models = list(
+                        dict.fromkeys(
+                            self._get_virtual_view_id(self._virtual_views[n].logical_id)
+                            for n in model.depends_on.nodes
+                            if n.startswith("model.")
+                        )
+                    )
 
-                    dbt_model.source_datasets = [
-                        str(source_map[n])
-                        for n in model.depends_on.nodes
-                        if n.startswith("source.")
-                    ]
+                    dbt_model.source_datasets = list(
+                        dict.fromkeys(
+                            str(source_map[n])
+                            for n in model.depends_on.nodes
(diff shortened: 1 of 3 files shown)


metaphor/dbt/extractor.py
```diff
@@ -325,17 +325,21 @@ def _parse_manifest_nodes(
...
-                        for n in model.depends_on.nodes
-                        if n.startswith("model.")
-                    ]
-
-                    dbt_model.source_datasets = [
-                        str(source_map[n])
-                        for n in model.depends_on.nodes
-                        if n.startswith("source.")
-                    ]
+                    dbt_model.source_models = list(
```
mars-lan, line 328: Consider encapsulating this in a utility function, e.g. `unique_list`, to improve readability.

mars-lan:
Please also give a more description `Why`, e.g. "Currently the connector outputs duplicated keys in `sourceModels` which can break the uniqueness validation"

mars-lan:
Please also bump up the version: https://github.com/MetaphorData/connectors/blob/main/pyproject.toml#L3

mars-lan (approved):
Great work!

metaphor/dbt/extractor.py
```diff
@@ -480,3 +486,7 @@ def _init_field_doc(dataset: Dataset, column: str) -> FieldDocumentation:
             doc.field_path = column
             dataset.documentation.field_documentations.append(doc)
         return doc
+
+    @staticmethod
+    def _unique_list(vars: List[str]) -> List[str]:
+        return list(dict.fromkeys(vars))
```
alyiwang, line 492: Sorry, late for the review. Maybe `list(set(values))` is simpler (more readable) ?

mars-lan: I believe @elic-eon wants to preserve the order, which is not strictly required but a nice to have property.

elic-eon: Yes, to prevent tests from randomly failing.

mars-lan: oh yeah that's right I forgot about the tests.

PR #117, opened 2022-01-06
Speed up snowflake profiler using approximation [sc-5603]

alyiwang:
### 🤔 Why?

The snowflake profiler is currently doing full table scan, which can be very slow for large tables. We should provide the option to do random sampling on the data. 

### 🤓 What?

- Add the capability and config to do random sampling based on percentage.

### 🧪 Tested?

Tested on dribble data with 1% sampling, finished `ANALYTICS` DB in 30min.

Update: finished dribble `RAW` and `ANALYTICS` in just under 1hr

### DIFF
--- a/metaphor/snowflake/profile/README.md
+++ b/metaphor/snowflake/profile/README.md
@@ -21,7 +21,7 @@ The config file inherits all the required and optional fields from the general S
 
 ### Optional Configurations
 
-#### Exclude_views
+#### Exclude Views
 
 By default, the connector only profile Snowflake "base table", i.e. exclude views and temporary tables. The following config can enable profiling on all tables and views.
 
@@ -29,15 +29,15 @@ By default, the connector only profile Snowflake "base table", i.e. exclude view
 exclude_views: false
 ```
 
-#### Sample Probability
+#### Sampling
 
-For tables with large number of rows, profiling can be inefficient. To speed up the process, we can choose to do percentage based random sampling on the data by setting the config below:
+For tables with large number of rows, it can take a long time to profile. To speed up the process, we can choose to do percentage-based random sampling on the data by setting `sample_percentage`:
 
 ```yaml
-sample_probability: <number between 0 and 100>
+sample_percentage: <number between 0 and 100>
 ```
 
-For example, `sample_probability = 1` means random sampling of 1% rows in the table. Keep in mind that sampling won't apply to smaller tables (with less than 100K rows), in that case, we do complete table profiling.
+For example, `sample_percentage = 1` means random sampling of 1% rows in the table. Keep in mind that sampling won't apply to smaller tables (with less than 100K rows), in that case, we do complete table profiling.
 
 ## Testing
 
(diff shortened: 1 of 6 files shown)


metaphor/snowflake/profile/README.md
```diff
@@ -19,12 +19,26 @@ grant select on future materialized views in database identifier($db) to role me
 
 The config file inherits all the required and optional fields from the general Snowflake connector [Config File](../README.md#config-file).
 
+### Optional Configurations
+
+#### Exclude_views
```
mars-lan, line 24:
```suggestion
#### Exclude_views
```
Exclude Views

metaphor/snowflake/profile/README.md
```diff
@@ -19,12 +19,26 @@ grant select on future materialized views in database identifier($db) to role me
...
+
 By default, the connector only profile Snowflake "base table", i.e. exclude views and temporary tables. The following config can enable profiling on all tables and views.
 
 ```yaml
 exclude_views: false
 ```
 
+#### Sample Probability
```
mars-lan, line 32:
```suggestion
#### Sampling
```

metaphor/snowflake/profile/README.md
```diff
@@ -19,12 +19,26 @@ grant select on future materialized views in database identifier($db) to role me
...
+
+For tables with large number of rows, profiling can be inefficient. To speed up the process, we can choose to do percentage based random sampling on the data by setting the config below:
```
mars-lan, line 34:
```suggestion
For tables with large number of rows, it can take a long time to profile. To speed up the process, we can choose to do a percentage-based random sampling on the data by setting `sample_probability`:
```

metaphor/snowflake/profile/README.md
```diff
@@ -19,12 +19,26 @@ grant select on future materialized views in database identifier($db) to role me
...
+
+```yaml
+sample_probability: <number between 0 and 100>
+```
+
+For example, `sample_probability = 1` means random sampling of 1% rows in the table. Keep in mind that sampling won't apply to smaller tables (with less than 100K rows), in that case, we do complete table profiling.
```
mars-lan, line 40: Is "probability" the right word here? Or should we call it "ratio" or "percentage"?

alyiwang: changed to percentage

metaphor/snowflake/profile/README.md
```diff
@@ -19,12 +19,26 @@ grant select on future materialized views in database identifier($db) to role me
...
 ```yaml
 exclude_views: false
 ```
 
+#### Sampling
+
+For tables with large number of rows, it can take a long time to profile. To speed up the process, we can choose to do percentage-based random sampling on the data by setting `sample_percentage`:
+
+```yaml
+sample_percentage: <number between 0 and 100>
```
mars-lan, line 37: `sampling_percentage` might be a better name.

alyiwang: updated

mars-lan (approved):

PR #119, opened 2022-01-11
Add bigquery usage crawler [sc-3967]

elic-eon:
### 🤔 Why?

To add dataset usage statistics for BigQuery.

### 🤓 What?

* Add bigquery usage crawler.
* Add `metaphor.common.usage_util` for common usage statistic calculation.
* Add `google-cloud-logging` to project dependency

### 🧪 Tested?

Local test passed. Add tests for both extractor and `metaphor.common.usage_util`.


metaphor/bigquery/extractor.py
```diff
@@ -9,7 +9,7 @@
 from smart_open import open
 
 try:
-    from google.cloud import bigquery
+    from google.cloud import bigquery  # type: ignore
```
mars-lan, line 12: Why?

metaphor/bigquery/usage/README.md
```diff
@@ -0,0 +1,51 @@
+# BigQuery Usage Statistics Connector
+
+This connector extract BigQuery usage statistics from a Google cloud project using [Python Client for Cloud Logging](https://googleapis.dev/python/logging/latest/index.html). It calculate usage statistics from BigQuery data read event from [audit log](https://cloud.google.com/logging/docs/audit/services).
+
+## Setup
+
+This connector need an account with permission of `Private Logs Viewers` to view BigQuery log entries.
+
+1. Go to [Service Accounts](https://console.cloud.google.com/iam-admin/serviceaccounts) in Google Cloud Console. Make sure the appropriate project is selected from the project dropdown.
```
mars-lan, line 9: Users of bigquery usage crawler will most likely also use the general bigquery crawler. As such, it might be more natural to add the additional permission to the existing service account, instead of adding a new account. See https://github.com/MetaphorData/connectors/blob/main/metaphor/snowflake/usage/README.md#setup for an example.

metaphor/bigquery/usage/README.md
```diff
@@ -0,0 +1,51 @@
...
+## Config File
+
+The config file inherits all the required and optional fields from the general BigQuery connector [Config File](../README.md#config-file). In addition, you can specify the following configurations:
+
+```yaml
+# (Optional) Number of days to include in the usage analysis. Default to 30.
+lookback_days: <days>
+
+# (Optional) A list of users whose queries will be excluded from the usage calculation 
+excluded_usernames:
```
mars-lan, line 33: Not used?

pyproject.toml
```diff
@@ -37,14 +38,15 @@ all = [
...
+  "google-cloud-logging",
   "lkml",
   "looker-sdk",
   "slack-sdk",
   "snowflake-connector-python",
   "sql-metadata",
   "tableauserverclient",
 ]
-bigquery = ["google-cloud-bigquery"]
+bigquery = ["google-cloud-bigquery", "google-cloud-logging"]
```
mars-lan, line 49: `google-cloud-logger` brought in a few transitive dependencies, which seems okay so far. If it does lead to dependency hell in the future, we may need to split this to a new `bigquery_usage` extra.

metaphor/common/usage_util.py
```diff
@@ -0,0 +1,168 @@
+from datetime import datetime, timedelta
```
mars-lan, line 1: I assume some code here can be leveraged by the existing snowflake usage crawler? If so, please create. follow-up PR to refactor the snowflake usage crawler, thanks.

metaphor/bigquery/usage/README.md
```diff
@@ -0,0 +1,51 @@
...
+## Setup
+
+This connector need an account with permission of `Private Logs Viewers` to view BigQuery log entries.
+
+1. Go to [Service Accounts](https://console.cloud.google.com/iam-admin/serviceaccounts) in Google Cloud Console. Make sure the appropriate project is selected from the project dropdown.
+2. Click `Create Service Account` and use the following settings to create a new account:
+    - Enter a service account name, e.g. `metaphor-bigquery`.
+    - Enter a description, e.g. `Metadata collection for Metaphor app`
+    - Click `CREATE AND CONTINUE`.
+    - Select `Private Logs Viewers` as the role and click `CONTINUE`.
```
mars-lan, line 14: Are you sure we need access to private logs? DH doc (which can be outdated) doesn't seem to suggest so? https://github.com/linkedin/datahub/blob/master/metadata-ingestion/source_docs/bigquery.md

mars-lan: We want to make sure that we only specify the minimal set of permissions needed.

tests/bigquery/usage/test_parser.py
```diff
@@ -0,0 +1,42 @@
+import re
+
+from metaphor.bigquery.usage.extractor import BigQueryUsageExtractor
+from metaphor.common.event_util import EventUtil
+from tests.test_utils import load_json
+
+try:
```
mars-lan, line 7: Can safely assume the test environment will be set up correctly so no need to be defensive here.

mars-lan:
Let's also add tests for the extractor itself. It'd require mocking like https://github.com/MetaphorData/connectors/blob/main/tests/bigquery/test_extractor.py

metaphor/common/usage_util.py
elic-eon: Yes, will create a PR later.

metaphor/bigquery/usage/README.md
elic-eon: Updated. Add excluded_usernames filter to support this option.

metaphor/bigquery/extractor.py
elic-eon: False alarm. Run `pip install mypy` then `mypy` won't complain anymore. See https://github.com/pypa/build/issues/211 and https://github.com/python/mypy/issues/1153.

mars-lan: If you run inside a poetry shell you should have everything installed: https://github.com/MetaphorData/connectors/blob/main/docs/develop.md

elic-eon: Still need a workaround after updating `mypy` to 0.931. Use `import google.cloud.bigquery as bigquery` for now.

metaphor/bigquery/usage/README.md
elic-eon: Updated

metaphor/bigquery/usage/README.md
elic-eon: Need `logging.privateLogEntries.list` to view Data Access audit logs. And updated the doc.

metaphor/bigquery/usage/README.md
```diff
@@ -0,0 +1,47 @@
+# BigQuery Usage Statistics Connector
+
+This connector extract BigQuery usage statistics from a Google cloud project using [Python Client for Cloud Logging](https://googleapis.dev/python/logging/latest/index.html). It calculate usage statistics from BigQuery data read event from [audit log](https://cloud.google.com/logging/docs/audit/services).
```
mars-lan, line 3:
```suggestion
This connector extract BigQuery usage statistics from a Google cloud project using [Python Client for Cloud Logging](https://googleapis.dev/python/logging/latest/index.html). It computes usage statistics from BigQuery data read event from the [audit log](https://cloud.google.com/logging/docs/audit/services).
```

metaphor/bigquery/usage/README.md
```diff
@@ -0,0 +1,47 @@
...
+```text
+logging.logEntries.list
+logging.logs.list
+logging.logServiceIndexes.list
+logging.logServices.list
+logging.privateLogEntries.list
+resourcemanager.projects.get
+```
+
+For further permission, see [Access Control](https://cloud.google.com/logging/docs/access-control#console_permissions)
```
mars-lan, line 18:
```suggestion
See [Access Control](https://cloud.google.com/logging/docs/access-control#console_permissions) for more information.
```

metaphor/bigquery/usage/config.py
```diff
@@ -0,0 +1,26 @@
...
+from typing import List, Optional, Set
+
+from serde import deserialize
+
+from metaphor.common.extractor import RunConfig
+
+
+@deserialize
+@dataclass
+class BigQueryUsageRunConfig(RunConfig):
```
mars-lan, line 12: Actually, we should extend https://github.com/MetaphorData/connectors/blob/main/metaphor/bigquery/extractor.py#L40 instead? (And move that class to a separate file for consistency in a follow-up PR)

metaphor/bigquery/usage/extractor.py
```diff
@@ -0,0 +1,181 @@
...
+            usage=usage,
+            columns=read_event.columns,
+            start_time=read_event.timestamp,
+            utc_now=self._utc_now,
+        )
+
+    @staticmethod
+    def build_filter(config: BigQueryUsageRunConfig, end_time):
+        def to_bq_format(t: datetime) -> str:
+            BQ_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
```
mars-lan, line 169: Since this is just ISO, I'd imagine a simple `t.isoformat()` would suffice?

mars-lan (commented):
Otherwise LGTM!

metaphor/bigquery/usage/config.py
elic-eon: Sure

metaphor/bigquery/usage/extractor.py
elic-eon: Actually, it is `rfc3339` and change to use `google.cloud` helper instead.

metaphor/bigquery/usage/extractor.py
```diff
@@ -0,0 +1,178 @@
...
+        UsageUtil.update_table_and_columns_usage(
+            usage=usage,
+            columns=read_event.columns,
+            start_time=read_event.timestamp,
+            utc_now=self._utc_now,
+        )
+
+    @staticmethod
+    def build_filter(config: BigQueryUsageRunConfig, end_time):
+        start = _datetime_to_rfc3339(end_time - timedelta(days=config.lookback_days))
```
mars-lan, line 169: Would prefer not to depend on private functions (`_` prefix) as it might change from version to version. Given that ISO is a subset of RFC3339, I'd assume using `isoformat` here would still work?

elic-eon: Yes, `isoformat` work here.

mars-lan (approved):
Great work!

metaphor/common/usage_util.py
```diff
@@ -0,0 +1,168 @@
...
+    FieldQueryCount,
+    FieldQueryCounts,
+    QueryCount,
+    QueryCounts,
+)
+
+
+class UsageUtil:
+    @staticmethod
+    def init_dataset(account: str, full_name: str, platform: DataPlatform) -> Dataset:
```
alyiwang, line 20: Some additional comments: 
1. this `init_dataset` can be also leveraged by snowflake usage crawler ?
2. should probably make the `account` as `Optional[str]`, since it only apply to snowflake for now. Give `None` to a `str` field will cause IDE warning.

mars-lan: Agree that `account` should be `Optional[str]`, but we should probably specify GCP project as the value for `account` since you the dataset names seems to be only unique within the project, not globally: https://cloud.google.com/bigquery/docs/datasets-intro

mars-lan: Actually never mind, looks like we already have the project name as the first part of the dataset name so no need for the account.

elic-eon: I will create another PR to fix `init_dataset` and leverage it to snowflake usage crawler.

PR #120, opened 2022-01-12
Move BigQueryRunConfig to a separate file [sc-6062]

elic-eon:
### 🤔 Why?

BigQueryRunConfig should be a module itself for consistency.

### 🤓 What?

Move BigQueryRunConfig to `config.py` from `extractor.py`

### 🧪 Tested?

Local tests passed.


mars-lan (approved):

mars-lan:
Please also bump up the version in pyproject.toml

PR #125, opened 2022-01-15
Tableau Lineage [sc-5919]

alyiwang:
### 🤔 Why?

Lineage info can help user understand the source data of the dashboards

### 🤓 What?

- parse tableau upstream dataset using Metadata API (graphql)

### 🧪 Tested?

unit tests
local run successfully, MCE looks good

### DIFF
--- a/metaphor/tableau/README.md
+++ b/metaphor/tableau/README.md
@@ -8,7 +8,7 @@ We recommend creating a dedicated Tableau user and role with limited permission 
 
 1. Log into your Tableau site as Site Administrator.
 2. Go to `Leftside Bar` > `Users`, click `Add Users` > `Add Users by Email`.
-3. In the pop-up window, select `Tableau` as authentication method, and provide an email for metaphor connector, then choose the `Viewer` role with read-only permission. This should generate an email containing a URL link to register the new user. 
+3. In the pop-up window, select `Tableau` as authentication method, and provide an email for metaphor connector, then choose the `Server Administrator` (Tableau Server) or `Site Administrator Explorer` (Tableau Online) role. This should generate an email containing a URL link to register the new user. The reason we need administrator role is to read the data connection information. 
 4. Follow the URL link to create user and password, and login to tableau.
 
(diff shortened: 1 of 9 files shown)


metaphor/tableau/README.md
```diff
@@ -8,7 +8,7 @@ We recommend creating a dedicated Tableau user and role with limited permission
 
 1. Log into your Tableau site as Site Administrator.
 2. Go to `Leftside Bar` > `Users`, click `Add Users` > `Add Users by Email`.
-3. In the pop-up window, select `Tableau` as authentication method, and provide an email for metaphor connector, then choose the `Viewer` role with read-only permission. This should generate an email containing a URL link to register the new user. 
+3. In the pop-up window, select `Tableau` as authentication method, and provide an email for metaphor connector, then choose the `Server Administrator` (Tableau Server) or `Site Administrator Explorer` (Tableau Online) role. This should generate an email containing a URL link to register the new user. The reason we need administrator role is to read the data connection information.
```
mars-lan, line 11: This is okay as an interim solution, but I wonder if we could ask the user to provide the data source information in the config file to avoid asking for admin permission, similar to what we did for Looker?

alyiwang: Tested with lower permission roles like "viewer" or "explorer", the REST API doesn't return anything 😞 , i.e. empty list of views and workbooks. Has to bump to "site administrator explorer" to start seeing things. Don't need "site administrator creator" though. 
Strangely, the graphql metadata API is working fine with "viewer" role, didn't find the specific requirement for the role in API spec, but I guess the graphql API is read-only, while the REST API can do updates (even though we don't). That's a possible reason for a higher permission role.

alyiwang: Amundson is only using the graphql API, so any role should be fine. But that lacks preview image and view count.

mars-lan: Thanks for the explanation. Having a read-only role still beats read-write.

mars-lan: So Tableau Server doesn't have an "explorer" permission?

mars-lan (approved):

metaphor/tableau/README.md
alyiwang: No, Server only have `server administrator` https://help.tableau.com/current/server/en-us/users_site_roles.htm

Issue #95, opened 2021-10-30
Breaking changes in 0.10


mars-lan:
- Use [pydantic](https://pydantic-docs.helpmanual.io/) to verify configs
- Drop support for JSON-based config files
- Output file path only specify directory
- Use new sampling config for all profilers
- Use new filter config for all connectors

PR #129, opened 2022-01-19
Metabase crawler [sc-6104]

alyiwang:
### 🤔 Why?

To fetch dashboard and lineage metadata from Metabase

### 🤓 What?

- Add Metabase crawler for dashboard and lineage

### 🧪 Tested?

Local run successfully. Produced 4 MCEs, examined the json.


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.8.28"
+version = "0.8.29"
```
mars-lan, line 3: We're onto 0.9 now.

metaphor/metabase/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+This connector extracts technical metadata from Metabase using [Metabase API](https://www.metabase.com/learn/administration/metabase-api.html).
+
+## Setup
+
+We recommend creating a dedicated Metabase user and role with limited permission for the connector to use.
+
+1. Log into Metabase as an Administrator.
+2. Go to `People` tab, click `Create Users` and fill out the name and email. An email should be sent to the user. 
+3. Follow the URL link in the email to activate user and password, and login to metabase.
```
mars-lan, line 11: No permission settings?

pyproject.toml
```diff
@@ -48,6 +48,7 @@ all = [
 ]
 bigquery = ["google-cloud-bigquery", "google-cloud-logging"]
 dbt = ["pydantic"]
+metabase = ["requests"]
```
mars-lan, line 51: `requests` is not made optional, should it be?

tests/metabase/test_config.py
```diff
@@ -0,0 +1,23 @@
+from metaphor.metabase.config import MetabaseRunConfig
```
mars-lan, line 1: Let's also add a story to add tests for the extractor.

mars-lan: By mocking the API response.

metaphor/metabase/README.md
alyiwang: Good point. On Metabase Cloud, there are only two default groups: admin and all-users. If we don't select admin, it defaults to all-user, which is what I did. I will add more instructions on how to create a Group and set permissions.

pyproject.toml
alyiwang: Just pulled

alyiwang: I was thinking of making the crawler metadata change as v0.9, maybe next time we can have the planned major version bump as we did before 😃

mars-lan: Yeah let's use this: https://github.com/MetaphorData/connectors/issues/95

The crawler metadata change isn't backward-incompatible per se?

metaphor/metabase/README.md
alyiwang: Tested the permissions, turns out, we need admin group to be able to do lineage. Otherwise, there is no info about the `database` part in the dataset full name `database.schema.table`. Updated the README accordingly.

tests/metabase/test_config.py
alyiwang: https://app.shortcut.com/metaphor-data/story/6176/add-tests-for-metabase-crawler

mars-lan (approved):

PR #130, opened 2022-01-20
Crawler status API [sc-3989]

alyiwang:
### 🤔 Why?

Each crawler run should produce metadata so we can check the status as well as keep the history.

### 🤓 What?

- update common extractor to output metadata file at the end of each run.
- add try-catch to ensure the extractor won't terminate by exception.

### 🧪 Tested?

local run successfully produced the metadata file.


metaphor/common/file_sink.py
```diff
@@ -83,3 +84,16 @@ def sink_logs(self):
...
+        if not self.write_logs:
+            logger.info("Skip writing metadata")
+            return
+
+        prefix = self.path[0:-5]
+
+        content = json.dumps(metadata.to_dict()).encode()
+
+        write_file(
+            f"{prefix}-run-metadata.json", content, True, s3_session=self.s3_session
```
mars-lan, line 98: Add the output folder to https://github.com/MetaphorData/connectors/issues/95

alyiwang: Already added a task there yesterday

metaphor/common/extractor.py
```diff
@@ -59,13 +60,31 @@ def config_class() -> Type[RunConfig]:
...
+            events = asyncio.run(self.extract(config))
+        except Exception as ex:
+            run_status = Status.FAILURE
+            logger.error(ex)
+            traceback.format_exc()
+
+        end_time = datetime.now()
         logger.info(
-            f"Fetched {len(events)} entities, took {format(time.time() - start_time, '.1f')} s"
+            f"Extractor ended successfully at {end_time}, fetched {len(events)} entities, took {format((end_time - start_time).total_seconds(), '.1f')} s"
```
mars-lan, line 78: Consider adding a local variable to hold `entity_count` to improve readability.

metaphor/common/file_sink.py
mars-lan: Nice. Sorry missed it.

metaphor/common/extractor.py
alyiwang: updated

mars-lan (dismissed):

mars-lan:
Need to fix format errors.

mars-lan (approved):

PR #132, opened 2022-01-21
Derive config class name directly in CLI

mars-lan:
### 🤔 Why?

The extract class already has a method to get the corresponding config class so there's no need to specify it explicitly.

### 🤓 What?

As titled. Also fixed missing dependency for mypy pre-commit hook (See https://github.com/pre-commit/mirrors-mypy/issues/50 for more details).

### 🧪 Tested?

Selectively ran a few connectors locally to verify.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,6 +26,8 @@ repos:
       rev: v0.931
       hooks:
           - id: mypy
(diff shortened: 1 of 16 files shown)


alyiwang (approved):
LGTM

elic-eon (approved):

PR #135, opened 2022-01-25
Refactor Snowflake sampling config [sc-6268]

elic-eon:
### 🤔 Why?

Apply sampling config for general profiling connector.

### 🤓 What?

- Refactor Snowflake profile connector to use new sampling config.

### 🧪 Tested?

Tests passed.

### DIFF
--- a/metaphor/snowflake/profile/extractor.py
+++ b/metaphor/snowflake/profile/extractor.py
@@ -1,11 +1,12 @@
 import logging
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Tuple
 
 from snowflake.connector import SnowflakeConnection
 
 from metaphor.common.event_util import EventUtil
 from metaphor.common.filter import DatasetFilter, include_table
 from metaphor.common.logger import get_logger
+from metaphor.common.sampling import SamplingConfig
 from metaphor.snowflake.auth import connect
 from metaphor.snowflake.extractor import SnowflakeExtractor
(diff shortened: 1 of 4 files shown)


mars-lan:
Could you also run the crawler manually once to make sure that it's working end-to-end?

metaphor/snowflake/profile/extractor.py
```diff
@@ -63,10 +62,10 @@ async def extract(
         self.max_concurrency = config.max_concurrency
         self.include_views = config.include_views
 
-        assert config.sampling_percentage is None or (
-            0 < config.sampling_percentage <= 100
-        ), f"Invalid sample probability ${config.sampling_percentage}, value must be between 0 and 100"
-        self._sampling_percentage = config.sampling_percentage
+        assert config.sampling is not None and (
```
mars-lan, line 65: Can move this common check (`> 0 && <= 100`) to `SamplingConfig` and leverage the logic in all connectors?

elic-eon: Maybe change it after applying `pydantic` validation?

mars-lan: SGTM

mars-lan (approved):

mars-lan:
Will likely need to merge & fix again after https://github.com/MetaphorData/connectors/pull/139

elic-eon:
> Will likely need to merge & fix again after #139

Looks like it can be merge without conflicts

PR #138, opened 2022-01-25
refactor crawlers output config to only specify [sc-6338]

alyiwang:
### 🤔 Why?

Current the file sink config specifies the MCE json file path, but the log file and metadata file path are derived from the json path. This is unclear to the user and unsafe.

Part of https://github.com/MetaphorData/connectors/issues/95

### 🤓 What?

- Change the file sink config to specify the directory instead, and the MCE file, as well as log and metadata, are all written to the same directory.
- Fix issues with pre-commit hook

### 🧪 Tested?

file written to local directory and S3 dev-yi
image

### DIFF
--- a/metaphor/common/file_sink.py
+++ b/metaphor/common/file_sink.py
@@ -22,8 +22,9 @@ logger = get_logger(__name__)
 @deserialize
 @dataclass
 class FileSinkConfig:
-    # Location of the sink file. Can be local file or s3://bucket/object
-    path: str
+    # Location of the sink directory, where the MCE file and logs will be output to.
+    # Can be local file directory, s3://bucket/ or s3://bucket/path/
+    directory_path: str
 
     # Output logs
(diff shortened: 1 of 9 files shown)


metaphor/common/file_sink.py
```diff
@@ -22,8 +22,9 @@
 @deserialize
 @dataclass
 class FileSinkConfig:
-    # Location of the sink file. Can be local file or s3://bucket/object
-    path: str
+    # Location of the sink directory, where the MCE file and logs will be output to.
+    # Can be local file directory, s3://bucket/ or s3://bucket/path/
+    directory_path: str
```
mars-lan, line 27: Maybe just `directory` for short?

alyiwang: updated

mars-lan (approved):

PR #139, opened 2022-01-26
Drop support for JSON-based config files

mars-lan:
### 🤔 Why?

It was a legacy feature that we no longer use since it's easier to write the config files in YAML instead of raw JSON. Part of https://github.com/MetaphorData/connectors/issues/95.

### 🤓 What?

As titled. Also use the opportunity to move the base config class to a separate file and rename it to the more fitting `BaseConfig`.

### 🧪 Tested?

Selectively ran a few connectors locally to verify.


elic-eon (commented):
LGTM

elic-eon (approved):

PR #140, opened 2022-01-26
Support tilde expansion when creating output directories

mars-lan:
### 🤔 Why?

https://stackoverflow.com/questions/2057045/os-makedirs-doesnt-understand-in-my-path

### 🤓 What?

As titled.

### 🧪 Tested?

Manually tested by specifying `~` in the `directory` config.


elic-eon (approved):
LGTM

PR #148, opened 2022-01-27
Migrate to pydantic-based validation for crawler configs [sc-4533]

elic-eon:
### 🤔 Why?

To have better config validation. Also, bump pydantic version to 1.9.0 since pydantic mypy plugin is not working with mypy 0.931

### 🤓 What?

- Use drop-in replacement `dataclass` decorator from pydantic for all config classes.

### 🧪 Tested?

Selectively run some connectors manually.


metaphor/common/base_config.py
```diff
@@ -26,7 +26,7 @@ class BaseConfig:
     All subclasses should add the @dataclass @deserialize decorators
     """
 
-    output: OutputConfig
+    output: Optional[OutputConfig]
```
mars-lan, line 29: Why making this optional? We want the user to always specify an output?

elic-eon: A lot of config in tests use `None`, if we always want the user to specify an output, let me change all these tests.

mars-lan: Got it. For tests we can write the output to some temp location. I think it's better to make it mandatory. We can add a "dry run" more if the user only want to test the core logic.

elic-eon: Ok, will update.

metaphor/bigquery/profile/extractor.py
```diff
@@ -39,7 +39,7 @@ def config_class():
         return BigQueryProfileRunConfig
 
     def __init__(self) -> None:
-        self._sampling = None
+        self._sampling = SamplingConfig()
```
mars-lan, line 42: This is not needed?

metaphor/snowflake/profile/extractor.py
```diff
@@ -50,7 +50,7 @@ def config_class():
     def __init__(self):
         self.max_concurrency = None
         self.include_views = None
-        self._sampling = None
+        self._sampling = SamplingConfig()
```
mars-lan, line 53: ditto

tests/bigquery/config.yml
```diff
@@ -1,6 +1,5 @@
+---
 key_path: key_path
 project_id: project_id
-dataset_filters:
```
mars-lan, line 3: Did pydantic error out on any extra config options by default? If not, can we make it so?

tests/common/test_extractor.py
```diff
@@ -12,21 +12,24 @@
 @deserialize
 @dataclass
 class DummyRunConfig(BaseConfig):
-    dummy_events: List[MetadataChangeEvent]
+    dummy_attr: int
```
mars-lan, line 15: Why changing this?

tests/snowflake/config.yml
```diff
@@ -4,13 +4,14 @@ user: user
...
+    db1:
+      schema1:
+    db2:
+      schema2:
+        - table1
+        - table2
+  excludes:
+    db3:
 output:
+  file:
```
mars-lan, line 17: Is this required?

metaphor/bigquery/profile/extractor.py
elic-eon: No, is it better to keep it the same type for consistency?

tests/common/test_extractor.py
elic-eon: Because of an attribute named `schema` in the MCE, another way is to change the MCE to a `pydantic` dataclass to fix it.

metaphor/bigquery/profile/extractor.py
mars-lan: Sorry, but could you elaborate? Not sure I fully understand what you meant by consistency.

tests/common/test_extractor.py
elic-eon: `schema` is preserved by the pydantic so need to move events out.

tests/snowflake/config.yml
elic-eon: Yes, output cannot be `None`, but can be `{ output: { file: null } }`

tests/bigquery/config.yml
elic-eon: Actually, is `serde` filtering out the extra field from the yaml. Seems there is no option to config throwing exception for that case.

metaphor/bigquery/profile/extractor.py
mars-lan: I see. Since `self.sampling` is assigned to `config.sample`, which has a default value, I'd imagine mypy won't treat it as `Optional`?  In other words, I think it's safe to drop the entire `__init__` method here?

elic-eon: Sure, it looks even better for me.

tests/bigquery/config.yml
mars-lan: Got it. Good enough for now.

mars-lan (approved):
Great refactoring!

tests/bigquery/config.yml
mars-lan: Consider replacing `serde` with `yaml.safe_load` in a follow-up PR?

PR #152, opened 2022-01-28
Attemp to fix pydantic dependency issue

mars-lan:
### 🤔 Why?

Installing from the latest PyPI package does not include `pydantic`.

### 🤓 What?

Dropping `pydantic` from dbt extra to see if it helps.

### 🧪 Tested?

N/A

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.10.6"
+version = "0.10.7"
 description = ""
 authors = ["Metaphor <dev@metaphor.io>"]
 packages = [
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -47,7 +47,7 @@ all = [
   "tableauserverclient",
 ]
 bigquery = ["google-cloud-bigquery", "google-cloud-logging"]
-dbt = ["pydantic"]
+dbt = []
```
alyiwang, line 50: can remove this line?

alyiwang (commented):
A similar fix, with updated dbt README 
https://github.com/MetaphorData/connectors/pull/153

pyproject.toml
mars-lan: This will remove the `dbt` extra

mars-lan:
> A similar fix, with updated dbt README #153

Prefer to keep the empty extra for consistency

mars-lan:
Closing in favor of https://github.com/MetaphorData/connectors/pull/153

PR #154, opened 2022-01-28
BQ lineage crawler needs to filter out duplicated [sc-6440]

alyiwang:
### 🤔 Why?

the referenced tables in big query log may contain duplicates

### 🤓 What?

- remove duplicates and self-referencing when parsing a log

### 🧪 Tested?

unit tests

### DIFF
--- a/tests/bigquery/lineage/data/sample_log.json
+++ b/tests/bigquery/lineage/data/sample_log.json
@@ -118,6 +118,8 @@
                 "billingTier": 1,
                 "referencedTables": [
                   "projects/metaphor-data/datasets/test/tables/yi_tests1",
+                  "projects/metaphor-data/datasets/test/tables/yi_tests2",
+                  "projects/metaphor-data/datasets/test/tables/yi_tests3",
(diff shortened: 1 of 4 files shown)


tests/bigquery/lineage/data/sample_log.json
```diff
@@ -118,6 +118,8 @@
                 "billingTier": 1,
```
mars-lan, line 118: Please also add a test for self reference.

alyiwang: There is self reference, the test3 is the destination dataset

mars-lan: Argh I see. Duplicated self reference :)

mars-lan (approved):

tests/bigquery/lineage/data/sample_log.json
alyiwang: test2 is duplicate, test3 is self-reference table

PR #155, opened 2022-01-30
Add config to disable preview images for Tableau connector

mars-lan:
### 🤔 Why?

Generating preview images can take a long time and result in an excessively large output for a large number of views.

### 🤓 What?

Add a config option to disable the feature.

### 🧪 Tested?

Manually tested against Tableau cloud with preview enabled & disabled.

### DIFF
--- a/metaphor/tableau/extractor.py
+++ b/metaphor/tableau/extractor.py
@@ -192,11 +192,17 @@ class TableauExtractor(BaseExtractor):
 
     def _parse_chart(self, view: ViewItem) -> Chart:
         # encode preview image raw bytes into data URL
-        preview_data_url = (
-            TableauExtractor._build_preview_data_url(view.preview_image, view.name)
-            if self._preview_image
-            else None
-        )
+        try:
+            preview_data_url = (
+                TableauExtractor._build_preview_data_url(view.preview_image)
+                if self._preview_image
+                else None
+            )
+        except Exception as error:
+            logger.error(
+                f"Failed to build preview data URL for {view.name}, error {error}"
+            )
+            return None
 
         view_url = self._build_view_url(view.content_url)
 
@@ -222,11 +228,5 @@ class TableauExtractor(BaseExtractor):
         return f"{self._base_url}/views/{workbook}/{view}" if self._base_url else None
 
     @staticmethod
-    def _build_preview_data_url(preview: bytes, view_name: str) -> Optional[str]:
-        try:
-            return f"data:image/png;base64,{base64.b64encode(preview).decode('ascii')}"
-        except Exception as error:
-            logger.error(
-                f"Failed to encode preview data URL for {view_name}, error {error}"
-            )
-            return None
+    def _build_preview_data_url(preview: bytes) -> Optional[str]:
+        return f"data:image/png;base64,{base64.b64encode(preview).decode('ascii')}"
--- a/metaphor/tableau/extractor.py
+++ b/metaphor/tableau/extractor.py
@@ -46,6 +46,7 @@ class TableauExtractor(BaseExtractor):
         self._views: Dict[str, ViewItem] = {}
         self._dashboards: Dict[str, Dashboard] = {}
         self._snowflake_account = None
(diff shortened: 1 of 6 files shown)


metaphor/tableau/extractor.py
```diff
@@ -189,17 +192,11 @@ def _parse_dataset_id(self, table: Dict) -> Optional[EntityId]:
...
-            )
-        except Exception as error:
-            logger.error(
-                f"failed to fetch preview for chart {view.name}, error {error}"
-            )
+        preview_data_url = (
+            TableauExtractor._build_preview_data_url(view.preview_image, view.name)
+            if self._preview_image
+            else None
+        )
```
alyiwang, line 199: We need the try-catch here. I've seen some cases that calling the `preview_image` throws an error, don't want that to fail the chart processing. Strange that the `populate_preview_image` method doesn't fail but the `preview_image` attribute fails.

mars-lan: Got it. I noticed that `populate_preview_image` isn't an actual field but a getter that invokes some other code and that's probably why it can fail. Moved the try-catch back.

metaphor/tableau/extractor.py
```diff
@@ -57,6 +58,7 @@ async def extract(self, config: TableauRunConfig) -> List[MetadataChangeEvent]:
         logger.info("Fetching metadata from Tableau")
 
         self._snowflake_account = config.snowflake_account
+        self._preview_image = not config.disable_preview_image
```
perepelitca, line 61: nit. Why introduce a new name?
```suggestion
        self._disable_preview_image = not config.disable_preview_image
```

mars-lan: Reverse logic is just a bit harder to read :)

```
if not self._disable_preview_image:
```

vs

```
if self._preview_image:
```

metaphor/tableau/extractor.py
```diff
@@ -225,5 +228,5 @@ def _build_view_url(self, content_url: str) -> Optional[str]:
         return f"{self._base_url}/views/{workbook}/{view}" if self._base_url else None
 
     @staticmethod
-    def _build_preview_data_url(preview: bytes) -> str:
+    def _build_preview_data_url(preview: bytes) -> Optional[str]:
```
alyiwang, line 231: why the `optional`?

metaphor/tableau/extractor.py
```diff
@@ -189,17 +192,17 @@ def _parse_dataset_id(self, table: Dict) -> Optional[EntityId]:
 
     def _parse_chart(self, view: ViewItem) -> Chart:
         # encode preview image raw bytes into data URL
-        preview_data_url = None
         try:
             preview_data_url = (
                 TableauExtractor._build_preview_data_url(view.preview_image)
-                if view.preview_image
+                if self._preview_image
```
alyiwang, line 198: prefer to use the `preview_image` instead of the internal attribute `_preview_image`

metaphor/tableau/extractor.py
mars-lan: Dropped.

metaphor/tableau/extractor.py
mars-lan: Updated.

metaphor/tableau/extractor.py
mars-lan: Actually now it's only used in one place, dropping the class variable.

metaphor/tableau/extractor.py
```diff
@@ -198,8 +198,9 @@ def _parse_chart(self, view: ViewItem) -> Chart:
             )
         except Exception as error:
             logger.error(
-                f"failed to fetch preview for chart {view.name}, error {error}"
+                f"Failed to build preview data URL for {view.name}, error {error}"
             )
+            return None
```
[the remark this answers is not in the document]
mars-lan: My bad. Copy & paste fail.

alyiwang (approved):
LGTM

PR #147, opened 2022-01-27
Bigquery lineage [sc-6365]

alyiwang:
### 🤔 Why?

Linage is critical information for customers

### 🤓 What?

- parse BigQueryAuditMetadata jobChange logs to get data lineage information

PR #156, opened 2022-01-31
Tableau lineage BigQuery use project Id [sc-6500]

alyiwang:
### 🤔 Why?

The Tableau API is returning the BigQuery project name as database name. This is not consistent with BQ crawler which uses project ID.

### 🤓 What?

- Add optional config to allow the user to provide the big query project ID, since it's difficult, hacky and unreliable to parse the info from the Tableau data sources API response.
- Add more logs in zip file to aid the debugging of missing lineage

### 🧪 Tested?

tested on local run, and verified the dataset ID matches the one from BQ crawler

### DIFF
--- a/metaphor/bigquery/lineage/extractor.py
+++ b/metaphor/bigquery/lineage/extractor.py
@@ -2,7 +2,7 @@ import logging
 import re
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
-from typing import Dict, List, Optional, Set
+from typing import Dict, List, Optional
 
 from metaphor.models.metadata_change_event import (
     DataPlatform,
@@ -128,7 +128,6 @@ class BigQueryLineageExtractor(BaseExtractor):
         self._datasets: Dict[str, Dataset] = {}
         self._datasets_pattern: List[re.Pattern[str]] = []
         self._dataset_filter: DatasetFilter = DatasetFilter()
-        self._excluded_usernames: Set[str] = set()
 
     async def extract(
         self, config: BigQueryLineageRunConfig
@@ -169,9 +168,6 @@ class BigQueryLineageExtractor(BaseExtractor):
             destination.project_id, destination.dataset_id, destination.table_id
         ):
             logger.info(f"Skipped table: {destination.table_name()}")
-            return
-
-        if job_change.user_email in self._excluded_usernames:
             return
 
         table_name = destination.table_name()
--- a/metaphor/tableau/README.md
+++ b/metaphor/tableau/README.md
@@ -67,6 +67,14 @@ If one of the data sources is using Snowflake dataset, please provide the Snowfl
 snowflake_account: <account_name>
 ```
 
+#### BigQuery project ID
+
+If one of the data sources is using BigQuery dataset, please provide the BigQuery project ID (NOT project name) as follows,
+
+```yaml
+bigquery_project_id: <project_id>
+```
+
 ## Testing
 
 Follow the [Installation](../../README.md) instructions to install `metaphor-connectors` in your environment (or virtualenv). Make sure to include either `all` or `tableau` extra.
--- a/metaphor/tableau/README.md
+++ b/metaphor/tableau/README.md
@@ -69,10 +69,12 @@ snowflake_account: <account_name>
 
 #### BigQuery project ID
 
-If one of the data sources is using BigQuery dataset, please provide the BigQuery project ID (NOT project name) as follows,
+If one of the data sources is using BigQuery dataset, please provide the BigQuery project name to project ID map as follows,
 
 ```yaml
-bigquery_project_id: <project_id>
+bigquery_project_name_to_id_map:
+  project_name1: prject_id1
+  project_name2: prject_id2
 ```
 
 ## Testing
--- a/metaphor/tableau/README.md
+++ b/metaphor/tableau/README.md
@@ -77,6 +77,8 @@ bigquery_project_name_to_id_map:
   project_name2: prject_id2
 ```
 
+More information about BigQuery project name and project ID can be found [here](https://cloud.google.com/resource-manager/docs/creating-managing-projects#before_you_begin) 
+
 ## Testing
 
 Follow the [Installation](../../README.md) instructions to install `metaphor-connectors` in your environment (or virtualenv). Make sure to include either `all` or `tableau` extra.
--- a/metaphor/tableau/config.py
+++ b/metaphor/tableau/config.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Dict, Optional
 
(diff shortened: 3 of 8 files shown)


metaphor/tableau/README.md
```diff
@@ -67,6 +67,14 @@ If one of the data sources is using Snowflake dataset, please provide the Snowfl
 snowflake_account: <account_name>
 ```
 
+#### BigQuery project ID
+
+If one of the data sources is using BigQuery dataset, please provide the BigQuery project ID (NOT project name) as follows,
+
+```yaml
+bigquery_project_id: <project_id>
```
mars-lan, line 75: This should be a map of project names to project IDs

alyiwang: updated

metaphor/tableau/README.md
```diff
@@ -67,6 +67,16 @@ If one of the data sources is using Snowflake dataset, please provide the Snowfl
 snowflake_account: <account_name>
 ```
 
+#### BigQuery project ID
+
+If one of the data sources is using BigQuery dataset, please provide the BigQuery project name to project ID map as follows,
```
mars-lan, line 72: See if you can find good links for BQ project name & ID?

metaphor/bigquery/lineage/extractor.py
```diff
@@ -171,9 +170,6 @@ def _parse_job_change_entry(self, entry: LogEntry):
             logger.info(f"Skipped table: {destination.table_name()}")
             return
 
-        if job_change.user_email in self._excluded_usernames:
```
mars-lan, line 174: Why removing this?

metaphor/tableau/config.py
```diff
@@ -32,5 +32,8 @@ class TableauRunConfig(BaseConfig):
     access_token: Optional[TableauTokenAuthConfig]
     user_password: Optional[TableauPasswordAuthConfig]
 
-    # snowflake data source account
-    snowflake_account: Optional[str]
+    # Snowflake data source account
+    snowflake_account: Optional[str] = None
+
+    # BigQuery data source project name to project ID map
+    bigquery_project_name_to_id_map: Optional[Dict] = None
```
mars-lan, line 39: Should probably initialize this to an empty map to avoid crashing when doing membership testing?

alyiwang: good point! updated

metaphor/tableau/README.md
alyiwang: added

metaphor/bigquery/lineage/extractor.py
alyiwang: In https://github.com/MetaphorData/connectors/pull/147 we discussed to remove `excluded_usernames` and may add `include_username` in future. I think some changes are left out in that PR, but a `git add -A` added them here...

metaphor/tableau/extractor.py
```diff
@@ -106,6 +114,7 @@ async def extract(self, config: TableauRunConfig) -> List[MetadataChangeEvent]:
             # by the REST api, can NOT use id to match entities.
             resp = server.metadata.query(workbooks_graphql_query)
             resp_data = resp["data"]
+            logger.debug(json.dumps(resp_data))
```
mars-lan, line 117: Add some context for the log as well?

metaphor/tableau/extractor.py
```diff
@@ -57,6 +59,10 @@ async def extract(self, config: TableauRunConfig) -> List[MetadataChangeEvent]:
         logger.info("Fetching metadata from Tableau")
 
         self._snowflake_account = config.snowflake_account
+        if config.bigquery_project_name_to_id_map is not None:
```
mars-lan, line 62: No longer needed?

metaphor/tableau/config.py
mars-lan: Actually you should do something like this: https://github.com/MetaphorData/connectors/blob/main/metaphor/bigquery/config.py#L21

metaphor/tableau/extractor.py
alyiwang: right, removed

metaphor/tableau/extractor.py
```diff
@@ -79,21 +82,25 @@ async def extract(self, config: TableauRunConfig) -> List[MetadataChangeEvent]:
...
             # fetch all workbooks
-            workbooks: List[WorkbookItem] = [item for item in Pager(server.workbooks)]
+            workbooks: List[WorkbookItem] = list(Pager(server.workbooks))
             logger.info(
                 f"\nThere are {len(workbooks)} work books on site: {[workbook.name for workbook in workbooks]}"
             )
             for item in workbooks:
                 server.workbooks.populate_views(item, usage=True)
+                # log workbook item detail for debugging purpose
+                logger.debug(json.dumps(item.__dict__, default=str))
```
mars-lan, line 103: What I meant is the log itself should include some context, e.g. `logger.debug(f'workbook item: {json.dump(...)})`

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.10.9"
+version = "0.10.10"
```
mars-lan, line 3: 11

metaphor/tableau/config.py
alyiwang: Seems the same as what I had here? 😄  ```
dataclasses.field(
        default_factory=dict
    )
    ```

mars-lan: I must have misread it or was reading an older verison.

metaphor/tableau/extractor.py
alyiwang: I prefer not to add this inline context: 1. there is some context above `There are {len(workbooks)} workbooks on site`. 2. Wrap JSON in string causes all the quotes to be eacaped, making it hard to read, and requiring an extra step to unescape the string then prettify it.

pyproject.toml
alyiwang: good catch, merged main

metaphor/tableau/extractor.py
mars-lan: SGTM

tests/tableau/test_config.py
```diff
@@ -13,7 +13,28 @@ def test_yaml_config(test_root_dir):
             token_value="bar",
         ),
         user_password=None,
-        disable_preview_image=True,
```
mars-lan, line 16: add it back?

mars-lan (commented):
Otherwise LGTM

tests/tableau/test_config.py
alyiwang: It's been merged to below, not removed

mars-lan (approved):

PR #159, opened 2022-02-01
Usage percentile doesn t work well when the [sc-6372]

alyiwang:
### 🤔 Why?

The calculated percentile for example like [1, 1, 1, 1, 5] is not reasonable, as all "1"s are 80%

### 🤓 What?

- Use rank mode instead of weak mode in percentile rank calculation. Details see https://www.notion.so/metaphordata/Calculation-of-Percentile-de54e81b70cd46f5a093cb373a72a4cc

### 🧪 Tested?

unit tests for all covered cases


mars-lan (approved):
Great work!

PR #165, opened 2022-02-03
small code refactoring

alyiwang:
### 🤔 Why?

DRY

### 🤓 What?

- use the common `table_fullname` function

### 🧪 Tested?

unit tests passed


mars-lan (commented):
Should we also do the case normalization (lower case everything) in `dataset_fullname`?

alyiwang:
> Should we also do the case normalization (lower case everything) in `dataset_fullname`?

It's already lowercased, `f"{db}.{schema}.{table}".lower()`

mars-lan (approved):

PR #166, opened 2022-02-03
Use "identity" instead of "name" when building dataset ID

mars-lan:
### 🤔 Why?

As described in https://docs.getdbt.com/reference/resource-properties/identifier, `identifier` is the real table name, whereas `name` is just for dbt's internal references. The two are often the same and that's why the connector has worked in many cases in the past.

### 🤓 What?

As titled.

### 🧪 Tested?

Manually tested against a real-world dbt manifest.


alyiwang (approved):
Thanks for the fix

PR #167, opened 2022-02-04
Use upstream table fullName

alyiwang:
### 🤔 Why?

In many cases the Tableau GQL upstreamTables `fullName` field contains the correct dataset fullname, especially for BQ, it's using the correct project id instead of the project name. We can try this field first, with fallback to table `name` and `schema` field

### 🤓 What?

- Update the Tableau upstream table name parsing logic to use `fullName` first, before falling back to `name` and `schema`
- Dedupe the upstream datasets

### 🧪 Tested?

- unit tests
- tested with production data GQL response

### DIFF
--- a/metaphor/tableau/extractor.py
+++ b/metaphor/tableau/extractor.py
@@ -157,18 +157,18 @@ class TableauExtractor(BaseExtractor):
         upstream_datasets = [
             self._parse_dataset_id(table) for table in workbook["upstreamTables"]
         ]
-        source_datasets = set(
-            [
-                str(dataset_id)
-                for dataset_id in upstream_datasets
-                if dataset_id is not None
-            ]
+        source_datasets = list(
+            set(
+                [
+                    str(dataset_id)
+                    for dataset_id in upstream_datasets
+                    if dataset_id is not None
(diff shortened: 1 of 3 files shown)


metaphor/tableau/extractor.py
```diff
@@ -157,17 +157,22 @@ def _parse_dashboard_upstream(self, workbook: Dict) -> None:
         upstream_datasets = [
             self._parse_dataset_id(table) for table in workbook["upstreamTables"]
         ]
-        source_datasets = [
-            str(dataset_id)
-            for dataset_id in upstream_datasets
-            if dataset_id is not None
-        ]
+        source_datasets = dict.fromkeys(
```
mars-lan, line 160: Why not just use `set([...])` to make it easier to understand?

alyiwang: updated

mars-lan: Sorry, should have suggested combining this with line 170, i.e. `list(set([...]))` so it's clear that the goal is to generate a unique list.

mars-lan (approved):

PR #168, opened 2022-02-04
Metabase upstream lineage not working [sc-6577]

alyiwang:
### 🤔 Why?

The previous crawler version doesn't handle the native SQL, so the chart upstream is incomplete

### 🤓 What?

- Add support for native SQL in charts, and parse the upstream datasets

### 🧪 Tested?

Used real-world data to test, the result matches the output of BQ crawler

### DIFF
--- a/metaphor/metabase/extractor.py
+++ b/metaphor/metabase/extractor.py
@@ -1,5 +1,6 @@
+import json
 from dataclasses import dataclass
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Union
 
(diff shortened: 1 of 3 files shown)


mars-lan:
Still WIP or ready for review?

alyiwang:
> Still WIP or ready for review?

Adding some unit tests. The code is tested against real environment and ready for review

mars-lan (approved):
Thanks for the quick fix, refactoring, and tests!

PR #169, opened 2022-02-07
Add redshift profiler [sc-5880]

elic-eon:
### 🤔 Why?

To extracts column-level data profiles from Redshift & PostgreSQL.

### 🤓 What?

- Add Redshift & PostgreSQL profile connectors.

### 🧪 Tested?

Run connectors locally, manually verify MCEs.

### DIFF
--- a/metaphor/snowflake/profile/extractor.py
+++ b/metaphor/snowflake/profile/extractor.py
@@ -3,6 +3,7 @@ from typing import Dict, List, Tuple
 
 from snowflake.connector import SnowflakeConnection
 
+from metaphor.common.entity_id import dataset_fullname
 from metaphor.common.event_util import EventUtil
 from metaphor.common.filter import DatasetFilter
 from metaphor.common.logger import get_logger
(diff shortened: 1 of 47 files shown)


metaphor/snowflake/profile/extractor.py
```diff
@@ -109,7 +110,7 @@ def _fetch_tables(
                 # exclude both view and temporary table
                 continue
 
-            full_name = SnowflakeExtractor.table_fullname(database, schema, name)
+            full_name = dataset_fullname(database, schema, name)
```
mars-lan, line 113: Interesting. How comes this wasn't caught by mypy or linters in https://github.com/MetaphorData/connectors/pull/165? @alyiwang

alyiwang: I guess if a file is not touched, the linters will ignore it?

mars-lan: Actually seems like this is one of the cases that mypy fails to detect an obvious mistake.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.10.19"
+version = "0.10.20"
```
mars-lan, line 3: Please update.

mars-lan (commented):
Otherwise LGTM

mars-lan (approved):
Great work!

PR #170, opened 2022-02-07
Support dbt 1 0 [sc-5881]

alyiwang:
### 🤔 Why?

[dbt 1.0](https://docs.getdbt.com/docs/guides/migration-guide/upgrading-to-1-0-0) introduced some breaking changes, the current parser won't work

### 🤓 What?

- forked the manifest parser to v3 and v4, to support incompatible schema
- refactor the catalog parser to a separate file too

### 🧪 Tested?

The existing parser tests all passed. Added the v4 parser tests using json generated with dbt 1.0

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.10.19"
+version = "0.10.20"
 license = "Apache-2.0"
 description = "A collection of Python-based 'connectors' that extract metadata from various sources to ingest into the Metaphor app."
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.10.19"
+version = "0.10.20"
```
mars-lan, line 3: Update

mars-lan (dismissed):
Thanks for the refactoring. This looks much cleaner now!

mars-lan (approved):

PR #171, opened 2022-02-08
Missing BQ lineage for views [sc-6663]

alyiwang:
### 🤔 Why?

The BigQuery view should have upstream lineage pointing to the queried tables, but the current lineage crawler doesn't handle that.

### 🤓 What?

- Fetch BQ view definition and parse the SQL to get upstream tables. 
- Make two toggles in BQ lineage config for enabling view lineage and log lineage

### 🧪 Tested?

unit tests
tested locally against metaphor BQ instance, view lineage is correct

### DIFF
--- a/metaphor/bigquery/lineage/README.md
+++ b/metaphor/bigquery/lineage/README.md
@@ -22,7 +22,7 @@ See [Access Control](https://cloud.google.com/logging/docs/access-control#consol
 The config file inherits all the required and optional fields from the general BigQuery connector [Config File](../README.md#config-file). In addition, you can specify the following configurations:
 
 ```yaml
-# (Optional) Whether to enable parsing view query to find upstream of the view, default True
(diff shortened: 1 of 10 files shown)


metaphor/bigquery/lineage/README.md
```diff
@@ -22,6 +22,12 @@ See [Access Control](https://cloud.google.com/logging/docs/access-control#consol
 The config file inherits all the required and optional fields from the general BigQuery connector [Config File](../README.md#config-file). In addition, you can specify the following configurations:
 
 ```yaml
+# (Optional) Whether to enable parsing view query to find upstream of the view, default True
```
mars-lan, line 25:
```suggestion
# (Optional) Whether to enable parsing view definition to build view lineage, default True
```

mars-lan (commented):
Otherwise LGTM!

mars-lan (approved):

PR #172, opened 2022-02-08
Add additional permission info for Redshift

mars-lan:
### 🤓 What?

Clarify that permission must be granted in all databases of interest.


alyiwang (commented):
Do we want to also mention how to switch between databases and grant permission?

mars-lan:
> Do we want to also mention how to switch between databases and grant permission?

I thought about the same as well, but since there are multiple ways to connect to Redshift, it'd be difficult to prescribe a specific way to do that. The assumption here is that the admin has basic knowledge of how to connect/switch to different DBs.

alyiwang (approved):
LGTM

PR #173, opened 2022-02-08
Redshift crawler index of out bound issue [sc-6687]

elic-eon:
### 🤔 Why?

When parsing precision, max_length from `fotmat_type` could cause `IndexError` for some reason. According to the document, `character varying` and `character` should be at least 1.

### 🤓 What?

- Add try-catch to safely parse `format_type`.

### 🧪 Tested?

n/a

### DIFF
--- a/metaphor/postgresql/extractor.py
+++ b/metaphor/postgresql/extractor.py
@@ -1,4 +1,4 @@
-from typing import Dict, List, Optional, Tuple
+from typing import Callable, Dict, List, Optional, Tuple
 
 from asyncpg import Connection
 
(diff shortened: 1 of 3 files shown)


metaphor/postgresql/extractor.py
```diff
@@ -306,10 +306,16 @@ def parse_format_type(
             elif native_type == "double precision":
                 precision = 53
```
alyiwang, line 307: Do we need to make `24` and `53` as `24.0` and `53.0`, validation may complain about casting int to float

metaphor/postgresql/extractor.py
```diff
@@ -306,10 +306,16 @@ def parse_format_type(
...
+                    precision = float(type_str.split("(")[1].split(",")[0])
+                except IndexError:
+                    logger.warning(f"Failed to parse precision from {type_str}.")
 
             if native_type == "character varying" or native_type == "character":
-                max_length = float(type_str.split("(")[1].strip(")"))
+                try:
+                    max_length = float(type_str.split("(")[1].strip(")"))
+                except IndexError:
+                    logger.warning(f"Failed to parse max_length from {type_str}.")
```
alyiwang, line 318: can we add a default `else` statement, simply logging the unsupported cases

alyiwang (commented):
Do we have an example of `type_str` value that failed the parsing?

metaphor/postgresql/extractor.py
elic-eon: Do you mean to catch other exceptions?

metaphor/postgresql/extractor.py
```diff
@@ -302,14 +302,20 @@ def parse_format_type(
...
                 precision = 64.0
             elif native_type == "real":
-                precision = 24
+                precision = 24.0
             elif native_type == "double precision":
-                precision = 53
+                precision = 53.0
             elif native_type == "numeric" and type_str != "numeric":
-                precision = float(type_str.split("(")[1].split(",")[0])
+                try:
```
mars-lan, line 309: Let's move this try-catch block to a util method to share code between `numeric` & `char` types.

PR #174, opened 2022-02-08
Redshift crawler still needs permission to actual schema [sc-6686]

elic-eon:
### 🤔 Why?

Cannot run Redshift crawler with minimum privilege.

### 🤓 What?

- `pg_total_relation_size` needs extra permission when parameter is table name. Use object ID of relation to solving this problem.

### 🧪 Tested?

- Run Redshift crawler with an ordinary user successfully
- Run PostgresSQL crawler successfully with the correct result.


mars-lan (approved):
Thanks for finding the root cause and the elegant fix!

elic-eon:
Bump version next PR.

PR #175, opened 2022-02-08
update dev dependency versions

alyiwang:
### 🤔 Why?

Keeping dev tools up-to-date

### 🤓 What?

- update dev dependency versions
- add `bandit` for code security check

### 🧪 Tested?

bandit is working for pre-commit check. Will see if the CI works

### DIFF
--- a/tests/common/test_extractor.py
+++ b/tests/common/test_extractor.py
@@ -1,6 +1,5 @@
 from typing import List, Type
 
-import pytest
 from metaphor.models.metadata_change_event import MetadataChangeEvent
 from pydantic.dataclasses import dataclass
 from serde import deserialize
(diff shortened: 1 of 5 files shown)


tests/common/test_extractor.py
```diff
@@ -28,7 +27,6 @@ async def extract(self, config: BaseConfig) -> List[MetadataChangeEvent]:
         return self._dummy_events
 
 
-@pytest.mark.asyncio
```
mars-lan, line 31: No longer needed?

alyiwang: right, no async execution in this test

mars-lan: But it ultimately called the async `extract`? Will that be an issue?

alyiwang: It calls the `run` which is not async, also underlying is a dummy function that's not async either.

alyiwang: This is a warning thrown by the upgraded `pytest-asyncio`

mars-lan: Got it.

mars-lan (approved):

PR #176, opened 2022-02-08
remove flakehell

alyiwang:
### 🤔 Why?

[flakehell](https://github.com/life4/flakehell) has been archived, and no more updates since a year ago. And it also doesn't work well with the latest flake8, preventing us from updating flake8 version. The included tests pycodestyle  and pyflakes are supported by flake8 out of the box.

### 🤓 What?

- use plain flake8 to replace flakehell, also update its version
- fix flake8 detected issues, e.g. code too complex for dbt parse, restructure the codes
- update pre-commit hook and CI

### 🧪 Tested?

pre-commit hook is working fine and much faster than before, since the flake8 hook checks the updated files, not the entire directory.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,8 @@
+[flake8]
+ignore = E203,E501,W503
+exclude =
+    venv,
+    __pycache__,
+    metaphor/dbt/generated,
+    **/.pytest_cache
+max-complexity = 12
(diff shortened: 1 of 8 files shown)


.flake8
```diff
@@ -0,0 +1,8 @@
+[flake8]
```
mars-lan, line 1: Too bad we can't keep this in pyproject.toml.

mars-lan (approved):
Thanks for updating this!

PR #177, opened 2022-02-08
Ignore system tables for Redshift & Postgres RDS

mars-lan:
### 🤔 Why?

There's nothing to be extracted from the default system databases but would still require granting permission to the users. 

### 🤓 What?

As titled.

### 🧪 Tested?

Verified locally.

### DIFF
--- a/metaphor/postgresql/config.py
+++ b/metaphor/postgresql/config.py
@@ -20,3 +20,8 @@ class PostgreSQLRunConfig(BaseConfig):
     filter: Optional[DatasetFilter] = field(default_factory=lambda: DatasetFilter())
 
     port: int = 5432
+
+    # Ignore RDS Postgres system databases
+    ignored_dbs: set[str] = field(
+        default_factory=lambda: set(["template0", "template1", "rdsadmin"])
+    )
--- a/metaphor/postgresql/extractor.py
+++ b/metaphor/postgresql/extractor.py
@@ -32,14 +32,6 @@ from metaphor.common.filter import DatasetFilter
 
 logger = get_logger(__name__)
 
-_ignored_dbs = ["template0", "template1", "rdsadmin"]
-_ignored_schemas = [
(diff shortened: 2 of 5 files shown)


metaphor/postgresql/config.py
```diff
@@ -20,3 +20,8 @@ class PostgreSQLRunConfig(BaseConfig):
     filter: Optional[DatasetFilter] = field(default_factory=lambda: DatasetFilter())
 
     port: int = 5432
+
+    # Ignore RDS Postgres system databases
+    ignored_dbs: set[str] = field(
+        default_factory=lambda: set(["template0", "template1", "rdsadmin"])
```
alyiwang, line 26: Can probably remove those three dbs since they are RDS specific?

metaphor/postgresql/extractor.py
```diff
@@ -32,14 +32,6 @@
 
 logger = get_logger(__name__)
 
-_ignored_dbs = ["template0", "template1", "rdsadmin"]
-_ignored_schemas = [
-    "pg_catalog",
-    "information_schema",
-    "pg_internal",
-    "catalog_history",
-]
```
alyiwang, line 41: Do we want to keep the capability to filter out the system schemas?

metaphor/postgresql/config.py
mars-lan: Yeah the assumption here is that it's RDS Postgres. Making this a config means user can still override it they choose to (e.g. crawling an non-RDS Postgres DB)

metaphor/postgresql/extractor.py
mars-lan: Seems like the new method already filtered out these system catalogs so that's why they're no longer needed.

mars-lan:
Actually it seems that people actually use these system databases, particularly `dev` in Redshift. Let's shelf this change for now.

PR #178, opened 2022-02-08
Update redshift usage connector with required permissions

mars-lan:
### 🤔 Why?

To explain why and how to give required permissions to the user.


elic-eon (approved):
Thanks for the update.

PR #179, opened 2022-02-09
Make dbt catalog parser optional [sc-6696]

alyiwang:
### 🤔 Why?

In most cases, the newer version of dbt generated manifest json already contains enough data about the sources and models, we don't necessarily require the catalog.json anymore. In case the manifest json doesn't contain certain column information, the catalog can be used to amend that. 

### 🤓 What?

- add documentation parsing in the manifest parser
- make catalog parser optional and only append info when manifest doesn't contain it
- update readme
- update tests

### 🧪 Tested?

Run through all sample jsons, the result MCE mostly only differ on the order of the datasets. Did actually fix a couple missing info in results


alyiwang:
> As per discussion offline, let's shelf this for now since it might cause more confusion as it's unclear when the catalog.json is actually needed.

I will revert the update in README, but I think the code change is still good. It added the column parsing for manifest, while previously we solely rely on catalog parser to do that. Now from the code, the catalog is purely an add-on.

mars-lan (approved):

PR #180, opened 2022-02-10
Bigquery profiler syntax error [sc-6749]

alyiwang:
### 🤔 Why?

Possibly due to column naming being a reserved keyword of BQ

### 🤓 What?

- escape all column names 

### 🧪 Tested?

updated unit tests
run locally against metaphor BQ instance


mars-lan (approved):
Thanks for the quick fix!

PR #181, opened 2022-02-10
Bigquery profiler failed to handle array [sc-6748]

alyiwang:
### 🤔 Why?

BQ array data type doesn't support `DISTINCT`

### 🤓 What?

- Skip `DISTINCT` query on complex data types

### 🧪 Tested?

unit test
added array and struct type in metaphor BQ instance to test out


mars-lan (dismissed):

elic-eon:
Maybe the Reddhift profiler has the same issue.

alyiwang:
> Maybe the Reddhift profiler has the same issue.

yeah, we should take care of that too.

elic-eon (approved):
LGTM. Thank you for refactoring and adding references.

PR #187, opened 2022-02-11
Catch future exception before use it [sc-6765]

elic-eon:
### 🤔 Why?

Crawler is stuck if the client has an error response from API.

### 🤓 What?

- Check future exceptions first.

### 🧪 Tested?

n/a, cannot reproduce an API error

### DIFF
--- a/metaphor/bigquery/profile/extractor.py
+++ b/metaphor/bigquery/profile/extractor.py
@@ -113,7 +113,7 @@ class BigQueryProfileExtractor(BigQueryExtractor):
             # The profiling result should only have one row
             elif job.result().total_rows != 1:
                 logger.warning(
-                    f"Skip {table}, the profiling result is more than one row"
+                    f"Skip {table}, the profiling result has more than one row"
                 )
             else:
                 try:
--- a/metaphor/bigquery/profile/extractor.py
+++ b/metaphor/bigquery/profile/extractor.py
@@ -93,7 +93,7 @@ class BigQueryProfileExtractor(BigQueryExtractor):
             finished = len(datasets) - len(jobs)
             if counter % 10 == 0:
                 logger.info(
-                    f"{finished} job done, waiting for {len(jobs)} jobs to finish ..."
+                    f"{finished} jobs done, waiting for {len(jobs)} jobs to finish ..."
                 )
             counter += 1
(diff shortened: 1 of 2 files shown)


metaphor/bigquery/profile/extractor.py
```diff
@@ -108,13 +108,13 @@ def _profile_table(
     ) -> Dataset:
         def job_callback(job: QueryJob, schema: DatasetSchema, dataset: Dataset):
```
mars-lan, line 109: Maybe it makes more sense to remove the job from `jobs` as soon as you enter the callback, instead of leaving it at the end after a series of if-else + try-catch?

metaphor/bigquery/profile/extractor.py
```diff
@@ -108,13 +108,13 @@ def _profile_table(
...
             exception = job.exception()
+            if exception:
```
mars-lan, line 111: I generally prefer early termination over complex if-else, i.e.

```
if exception:
  ...
  return

if job.results().total_rows != 1: 
  ...
  return
```

metaphor/bigquery/profile/extractor.py
elic-eon: Tried that before but it may cause the last results not processed.

mars-lan: How come? Isn't this callback invoked only when the job is finished? Removing it from `jobs` shouldn't interfere with the result processing?

elic-eon: https://github.com/MetaphorData/connectors/blob/1505376c7918855e64e530abe981cc8926fddb73/metaphor/bigquery/profile/extractor.py#L92-L99
This loop could end first, so the last one might be not updated by `_parse_result` https://github.com/MetaphorData/connectors/blob/1505376c7918855e64e530abe981cc8926fddb73/metaphor/bigquery/profile/extractor.py#L121

metaphor/bigquery/profile/extractor.py
mars-lan: Would be difficult to write it in this form if we can't remove job at the very beginning.

mars-lan (approved):

PR #189, opened 2022-02-14
Handle pagination is S3 list objects [sc-6783]

alyiwang:
### 🤔 Why?

the previous implementation of S3 list-objects doesn't support pagination, so if there are more than 1000 items in the bucket sub-directory, some items won't be fetched.

### 🤓 What?

- add pagination support in `list_files`

### 🧪 Tested?

- tested locally against dev s3 bucket, used small batch size to force pagination, works well.

### DIFF
--- a/metaphor/common/storage.py
+++ b/metaphor/common/storage.py
@@ -94,13 +94,21 @@ class S3Storage(BaseStorage):
     def list_files(self, path: str, suffix: Optional[str]) -> List[str]:
         bucket, key = S3Storage.parse_s3_uri(path)
 
+        objects = []
         resp = self._client.list_objects_v2(
             Bucket=bucket,
             Prefix=key,
         )
-        # TODO: handle pagination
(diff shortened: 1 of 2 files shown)


metaphor/common/storage.py
```diff
@@ -94,13 +94,21 @@ def write_file(
     def list_files(self, path: str, suffix: Optional[str]) -> List[str]:
         bucket, key = S3Storage.parse_s3_uri(path)
 
+        objects = []
         resp = self._client.list_objects_v2(
             Bucket=bucket,
             Prefix=key,
         )
-        # TODO: handle pagination
+        objects.extend(resp.get("Contents", []))
```
mars-lan, line 102: Can probably kept the one line `objects = resp.get("Contents", [])` here instead of predeclaring `objects` on line 97?

alyiwang: good point, updated

mars-lan (approved):

PR #191, opened 2022-02-16
Redshift profiler overflow [sc-6825]

elic-eon:
### 🤔 Why?

Redshift profiler gets `ERROR:metaphor.common.extractor:Numeric data overflow (addition)`. Root cause most likely is calculating the average when the sum of column exceeds `bigint` range.

### 🤓 What?

- Cast to float before calculating the average.
- Handle some SQL errors to keep the connecter finished smoothly.

### 🧪 Tested?

Run connector locally.


mars-lan (approved):
Thanks for the quick fix!

PR #192, opened 2022-02-17
FileNotFoundError for crawlers [sc-6870]

alyiwang:
### 🤔 Why?

Local storage threw an error when the directory doesn't exist, also it doesn't expand `~`

### 🤓 What?

- Add directory expansion and existence check. 

### 🧪 Tested?

Tested locally 
unit test


elic-eon (approved):
LGTM

PR #194, opened 2022-02-17
Do not crawl BQ partitioned tables [sc-6868]

alyiwang:
### 🤔 Why?

Sometimes the BQ logs TableResources contains table names with partition attached, it may also contain temporary tables. Need to do sanitization before we collect the data.

### 🤓 What?

- strip table partition and skip temporary tables in BQ usage crawler
- add logging to show progress

### 🧪 Tested?

unit tests to cover those cases
tested against production env

### DIFF
--- a/metaphor/bigquery/usage/extractor.py
+++ b/metaphor/bigquery/usage/extractor.py
@@ -2,7 +2,7 @@ import logging
 import re
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
-from typing import Dict, List, Set
+from typing import Dict, List, Optional, Set
 
 from metaphor.models.metadata_change_event import (
     DataPlatform,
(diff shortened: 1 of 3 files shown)


metaphor/bigquery/usage/extractor.py
```diff
@@ -41,10 +41,15 @@ def can_parse(entry: LogEntry) -> bool:
             return False
 
     @staticmethod
-    def from_entry(entry: LogEntry) -> "TableReadEvent":
+    def from_entry(entry: LogEntry) -> Optional["TableReadEvent"]:
+        bq_resource = BigQueryResource.from_str(
+            entry.payload["resourceName"]
+        ).remove_extras()
```
mars-lan, line 47: This removes the partition name?

alyiwang: Yes. I forgot I already added this handling for the lineage crawler. https://github.com/MetaphorData/connectors/blob/585e6d2311471c0f9f944ce6ef8a21994d2a437b/metaphor/bigquery/utils.py#L105

mars-lan (approved):

PR #195, opened 2022-02-18
Redshift profiler error: lists can have at most 1664 entries [sc-6885]

elic-eon:
### 🤔 Why?

The targets of profiling SQL may exceed the limit (1664 entries) because we will generate 1-5 entries for a column.

### 🤓 What?

- Catch PostgresError to smooth the profile connector
- Split profiling query into multiple parts if more than 100 targets in a query.

### 🧪 Tested?

- Add a testcase to test new `_build_profiling_query`.
- Create a table with 600 columns to reproduce the issue, run the connector successfully.


metaphor/postgresql/extractor.py
```diff
@@ -357,6 +357,17 @@ def _parse_format_type(
     ) -> Tuple[Optional[float], Optional[float]]:
         precision, max_length = None, None
 
+        excluded_types = [
```
alyiwang, line 360: nit: a bit safer to use tuple `()` instead of mutable list `[]`

metaphor/postgresql/profile/extractor.py
```diff
@@ -96,7 +96,11 @@ async def _profile_dataset(self, pool: asyncpg.Pool, dataset: Dataset) -> None:
                 res = await conn.fetch(sql)
                 assert len(res) == 1
                 self._parse_result(res[0], dataset)
-            except asyncpg.exceptions.InternalServerError as ex:
+            except asyncpg.exceptions.ProgramLimitExceededError as ex:
+                logger.warn(
+                    f"Skip {dataset.logical_id.name}, too many column to profile, err: {ex}"
```
alyiwang, line 101: instead of skipping table when exceeds the limit, can we proactively split the profiling query into two or more queries?

elic-eon: Great suggestion, thanks.

mars-lan (approved):

PR #200, opened 2022-02-22
Drop serde.deserialize in favor of pydantic's validation for configs [sc-6719]

elic-eon:
### 🤔 Why?

As title.

### 🤓 What?

- Use `yaml.safe_load` to load yaml content to object.
- Pass the object as `kwargs` to `pydantic` dataclass because the decorated dataclass does not support the `parse_obj` function.
- Remove `serde.deserialize` decorator from config classes.
- Move the `pyserde` to optional dependency, since snowflake usage connector use it to deserialize access log. 

### 🧪 Tested?

The test cases passed.
BigQuery and Snowflake connectors run manually.


pyproject.toml
```diff
@@ -23,7 +23,7 @@ lkml = { version = "^1.1.0", optional = true }
 looker-sdk = { version = "^21.4.1", optional = true }
 metaphor-models = "^0.6.15"
 pydantic = "^1.9.0"
-pyserde = { extras = ["yaml"], version = "^0.4.0" }
+pyserde = {extras = ["snowflake"], version = "^0.7.0"}
```
mars-lan, line 26: Doesn't seem like it has the `snowflake` extra? https://github.com/yukinarit/pyserde/blob/master/pyproject.toml#L54

I think you want to make this optional and add it to our `snowflake` extra instead?

elic-eon: Yes, making it optional is what I wanted.

pyproject.toml
```diff
@@ -23,7 +23,7 @@ lkml = { version = "^1.1.0", optional = true }
 looker-sdk = { version = "^21.4.1", optional = true }
 metaphor-models = "^0.6.15"
 pydantic = "^1.9.0"
-pyserde = { extras = ["yaml"], version = "^0.4.0" }
+pyserde = { extras = ["yaml"], version = "^0.4.0", optional = true }
```
mars-lan, line 26: Since snowflake usage connector only uses JSON serde, you can safely drop the yaml extra here?

mars-lan (approved):
Thanks for the refactoring

alyiwang:
nvm, I am a bit late 😃 .  I will do it in next PR

elic-eon:
> Since you are at it, we can remove the `pyserde` completely by replacing `pyserde.from_json` to `pydantic.parse_raw_as` in the snowflake usage crawlers

Maybe we cannot because we use the dataclass, dataclass did provide the helper function. But we can refactor these snowflake dataclass to pydantic model.

PR #201, opened 2022-02-22
Airflow connector [sc-972]

elic-eon:
### 🤔 Why?

To extract lineage from Airflow.

### 🤓 What?

- Provide a lineage backend to collect lineage information.

### 🧪 Tested?

Test on [aws-mwaa-local-runner](https://github.com/aws/aws-mwaa-local-runner) (Airflow 2.0.2). Lineage information ingested both via API or s3.

image


mars-lan (commented):
Great work! Please also update the top-level README file.

metaphor/common/file_sink.py
```diff
@@ -16,6 +17,13 @@
 logger = get_logger(__name__)
 
 
+@dataclass
+class S3AuthConfig:
+    aws_access_key_id: Optional[str] = None
+    aws_secret_access_key: Optional[str] = None
+    aws_session_token: Optional[str] = None
```
mars-lan, line 24: Since the user can already specify `assume_role_arn` as part of the config, it makes more sense to use that instead of passing in a session token.

elic-eon: Thank you to point it out. Updated.

mars-lan (commented):
Great work. Can we rename the package to `metaphor.airflow` for consistency?

elic-eon:
> Great work. Can we rename the package to `metaphor.airflow` for consistency?

I was using `metaphor.airflow` in the beginning. But then I couldn't import `airflow` because they have the same module name.

mars-lan:
> > Great work. Can we rename the package to `metaphor.airflow` for consistency?
> 
> I was using `metaphor.airflow` in the beginning. But then I couldn't import `airflow` because they have the same module name.

Could you elaborate more? I'd assume you'd never do

```
from metaphor import airflow
```

so why would it conflict with `airflow`?

elic-eon:
Oh, `mypy` can't tell `apache-airflow` and `metaphor.airflow` because they have the same name.

mars-lan:
> Oh, `mypy` can't tell `apache-airflow` and `metaphor.airflow` because they have the same name.

Got it. In that case let's go with `airflow_plugin`.

elic-eon:
> 

Looks no way to set the `import airflow` to correct one, now just bypass it.

mars-lan:
> > 
> 
> Looks no way to set the `import airflow` to correct one, now just bypass it.

Looks like it confuses LGTM as well. I guess the Python search path is messed up by the same module name. In that case, let's stick with `airflow_lineage` to avoid any potential conflicts? Sorry about the back-and-forth.

elic-eon:
> Looks like it confuses LGTM as well. I guess the Python search path is messed up by the same module name. In that case, let's stick with airflow_lineage to avoid any potential conflicts? Sorry about the back-and-forth.
Reverted. Looks many code quality checking tools can not handle this situation.

mars-lan (approved):

PR #202, opened 2022-02-22
remove Pyserde 

alyiwang:
### 🤔 Why?

After previous PR https://github.com/MetaphorData/connectors/pull/200, there is not need to use pyserde anymore, we can rely only on pydantic to do serde

### 🤓 What?

- refactor pyserde dataclass to pydantic dataclass
- remove pyserde from dependency

### 🧪 Tested?

Local run snowflake usage crawler successfully

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,7 @@ repos:
           - types-freezegun
           - types-requests
           - types-pytz
+          - types-PyYAML
 
   - repo: https://github.com/pycqa/bandit
     rev: 1.7.2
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -33,7 +33,6 @@ repos:
           - types-freezegun
           - types-requests
           - types-pytz
-          - types-PyYAML
```
mars-lan, line 36: No longer needed?

alyiwang: turns out `mypy` needs it, added it back

mars-lan (approved):

PR #205, opened 2022-02-25
Improve BigQuery crawler performance [sc-6022]

elic-eon:
### 🤔 Why?

As title.

### 🤓 What?

- Call google API in parallel

### 🧪 Tested?

- Run `metaphor.bigquery.profile` locally. Speed up from 60 sec to 8.8 secs with 10 concurrencies.
- Run `metaphor.bigquery` locally. Speed up from 22 secs to 3.3 secs with 10 concurrencies.

Both MCE files were manually checked, no different.


mars-lan (approved):
Great work

PR #206, opened 2022-02-25
Add PostgreSQL usage crawler

elic-eon:
### 🤔 Why?

To extract usage metadata from PostgreSQL database.

### 🤓 What?

PostgreSQL did not record queries in the database but it can be configured to log to a file. We did not use the log parsing approach because it is more complicated to specify the log path or the log format.

PostgresSQL has a built-in [statistic collector](https://www.postgresql.org/docs/12/monitoring-stats.html#PG-STAT-ALL-TABLES-VIEW). We can easily extract basic usage metadata from it. 

### 🧪 Tested?

Local run connector successfully.

### DIFF
--- a/metaphor/common/usage_util.py
+++ b/metaphor/common/usage_util.py
@@ -47,9 +47,8 @@ class UsageUtil:
         return dataset
 
     @staticmethod
-    def update_table_and_columns_usage(
+    def update_table_usage(
         usage: DatasetUsage,
-        columns: List[str],
         start_time: datetime,
         utc_now: datetime,
     ):
@@ -67,6 +66,15 @@ class UsageUtil:
 
         if start_time > utc_now - timedelta(365):
             usage.query_counts.last365_days.count += 1
+
+    @staticmethod
+    def update_table_and_columns_usage(
+        usage: DatasetUsage,
+        columns: List[str],
+        start_time: datetime,
+        utc_now: datetime,
+    ):
+        UsageUtil.update_table_usage(usage, start_time, utc_now)
 
         for column_name in columns:
             if start_time > utc_now - timedelta(1):
@@ -108,9 +116,7 @@ class UsageUtil:
             query_counts.append(FieldQueryCount(field=column, count=1.0))
 
     @staticmethod
-    def calculate_statistics(datasets: Collection[Dataset]) -> None:
-        """Calculate statistics for the extracted usage info"""
-
+    def calculate_table_statstics(datasets: Collection[Dataset]) -> None:
         UsageUtil.calculate_table_percentile(
             datasets, lambda dataset: dataset.usage.query_counts.last24_hours
         )
@@ -130,6 +136,11 @@ class UsageUtil:
         UsageUtil.calculate_table_percentile(
             datasets, lambda dataset: dataset.usage.query_counts.last365_days
         )
+
+    @staticmethod
+    def calculate_statistics(datasets: Collection[Dataset]) -> None:
+        """Calculate statistics for the extracted usage info"""
+        UsageUtil.calculate_table_statstics(datasets)
 
         for dataset in datasets:
             UsageUtil.calculate_column_percentile(
--- a/metaphor/common/usage_util.py
+++ b/metaphor/common/usage_util.py
@@ -47,8 +47,9 @@ class UsageUtil:
         return dataset
 
     @staticmethod
-    def update_table_usage(
+    def update_table_and_columns_usage(
         usage: DatasetUsage,
+        columns: List[str],
         start_time: datetime,
         utc_now: datetime,
     ):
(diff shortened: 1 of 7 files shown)


metaphor/common/usage_util.py
```diff
@@ -47,9 +47,8 @@ def init_dataset(
         return dataset
 
     @staticmethod
-    def update_table_and_columns_usage(
+    def update_table_usage(
```
alyiwang, line 50: This seems very similar to the `update_table_and_columns_usage`, maybe just input an empty list as columns, instead f creating a new function?

metaphor/postgresql/usage/extractor.py
```diff
@@ -0,0 +1,66 @@
...
+                    full_name = dataset_fullname(db, schema, table_name)
+
+                    if not dataset_filter.include_table(db, schema, table_name):
+                        logger.info(f"Ignore {full_name} due to filter config")
+                        continue
+
+                    dataset = UsageUtil.init_dataset(
+                        None, full_name, DataPlatform.POSTGRESQL
+                    )
+                    dataset.usage.query_counts.last365_days.count = float(read_count)
```
mars-lan, line 60: So there's no specific time range associated with `seq_scan`?

metaphor/postgresql/usage/README.md
```diff
@@ -0,0 +1,19 @@
+# PostgreSQL Usage Statistics Connector
```
mars-lan, line 1: Any specific permission needed to access `pg_stat_user_tables`?

metaphor/common/usage_util.py
```diff
@@ -108,9 +116,7 @@ def update_field_query_count(query_counts: List[FieldQueryCount], column: str):
             query_counts.append(FieldQueryCount(field=column, count=1.0))
 
     @staticmethod
-    def calculate_statistics(datasets: Collection[Dataset]) -> None:
-        """Calculate statistics for the extracted usage info"""
-
+    def calculate_table_statstics(datasets: Collection[Dataset]) -> None:
```
alyiwang, line 119: Same comments as above, can probably just call the `calculate_statistics`, IIRC if no columns the column stats won't be calculated, you can verify if that's the case.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.10.48"
+version = "0.10.49"
```
alyiwang, line 3: merge main and update version

metaphor/postgresql/usage/extractor.py
alyiwang, line 60: If we can't determine the actual date of the table read count, we should set the `last30days` and 90/365 days. The reason is that the BE/FE is currently using only the last30days to calculate usage level.

metaphor/postgresql/usage/extractor.py
```diff
@@ -0,0 +1,66 @@
...
+
+logger = get_logger(__name__)
+
+
+USAGE_SQL = """
+SELECT schemaname, relname, seq_scan FROM pg_stat_user_tables
+"""
+
+
+class PostgreSQLUsageExtractor(PostgreSQLExtractor):
```
alyiwang, line 20: I am thinking, this can be potentially merged into the PostgreSQL crawler instead of a new crawler, because it uses the same mechanism (query pg tables), and it doesn't add too much extra overhead. Making it one is easier to manage. WDYT @elic-eon @mars-lan ?

mars-lan: One is definitely easier to manage but I'd opt for consistency in this case.

metaphor/postgresql/usage/README.md
elic-eon: I create a user to verify this. But no extra permission is needed.

metaphor/postgresql/usage/extractor.py
elic-eon: Superusers can reset statistics manually then `seq_scan` would start from it.

metaphor/common/usage_util.py
elic-eon: Sounds good, I don't need to touch usage_util

metaphor/postgresql/usage/extractor.py
elic-eon: Updated.

metaphor/postgresql/usage/extractor.py
elic-eon: It could need to merge usage `init_dataset` to the `init_dataset function` PostgreSQL crawler. If we change that so 
 maybe the RedShift crawler would pollute with zero usage metadata.

alyiwang (approved):
LGTM

PR #207, opened 2022-02-25
Update usage crawlers [sc-6907]

alyiwang:
### 🤔 Why?

To optimize the log-parsing crawlers, we introduced the mechanism to do day-by-day log parsing and save the result to metadata history. This PR updates the dataset usage crawlers to add an option of using dataset_usage_history to write the crawler output.

### 🤓 What?

- Add use_history option to crawler config
- Update crawlers to write to dataset_usage_history if configured so, and only fetch 1 day's logs in that case.
- Use UTC start of the day as the measure of datetime

### 🧪 Tested?

- updated unit tests
- tested the three usage crawlers locally


metaphor/common/utils.py
```diff
@@ -0,0 +1,8 @@
+from datetime import datetime, time, timedelta, timezone
+
+
+def start_of_day(daysAgo=0) -> datetime:
```
elic-eon, line 4: Is it better to name something else? `start_of_day` sounds like to get the 12 am DateTime of a day to me.

metaphor/common/usage_util.py
```diff
@@ -18,14 +20,29 @@
 class UsageUtil:
     @staticmethod
     def init_dataset(
```
elic-eon, line 22: Should we also update tests for `usage_utils` changes?

metaphor/common/usage_util.py
```diff
@@ -99,6 +116,15 @@ def update_table_and_columns_usage(
                     column_name,
                 )
 
+    @staticmethod
+    def update_table_and_columns_usage_history(
```
elic-eon, line 120: Also here, to add a new test?

metaphor/common/utils.py
alyiwang: It's exactly `get the 12 am DateTime of a day` in UTC time, by default it's today, but can optionally specify N days ago.

elic-eon: Oh, my bad. There was a `.date()`

metaphor/common/usage_util.py
alyiwang: added some tests in bigquery and redshift usage crawlers

elic-eon (approved):
LGTM

PR #208, opened 2022-02-28
Add manual lineage connector

mars-lan:
### 🤔 Why?

To provide the ability to manually specify lineage via a config file.

### 🤓 What?

As titled. Also emit a placeholder dataset for each unique upstream ID so that "external datasets" that are not associated with a crawler/connector will work properly.

### 🧪 Tested?

Ran locally and verified the output MCE.


metaphor/manual/lineage/extractor.py
```diff
@@ -0,0 +1,49 @@
...
+                )
+
+            datasets.append(
+                Dataset(
+                    logical_id=lineage.dataset.to_logical_id(),
+                    upstream=DatasetUpstream(source_datasets=list(source_datasets)),
+                )
+            )
+
+        datasets.extend(extra_datasets.values())
```
alyiwang, line 47: The `extra_datasets` may contain items that are already in `datasets`. So extending it might create duplicate.

mars-lan: Remove duplicates before appending.

metaphor/manual/lineage/README.md
```diff
@@ -0,0 +1,61 @@
...
+### Required Configurations
+
+```yaml
+lineages:
+  - dataset:
+      platform: <data_platform>
+      name: <dataset_name>
+    upstreams:
+      - platform: <dataset_platform>
+        name: <dataset_name>
```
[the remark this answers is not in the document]
mars-lan: Added

elic-eon (approved):
LGTM

PR #210, opened 2022-03-02
 Extract metrics as entities from dbt 1 0 [sc-6331]

alyiwang:
### 🤔 Why?

dbt crawler should extract metric information from dbt 1.0 manifest

### 🤓 What?

- Add metric parser for dbt manifest v4

### 🧪 Tested?

added metric to trial project and use the generated manifest in updated unit tests


elic-eon (approved):
LGTM

PR #213, opened 2022-03-03
Snowflake usage crawler error [sc-7154]

alyiwang:
### 🤔 Why?

pydantic validation does not allow unexpected fields in a model. Previous python build-in dataclass simply ignores extra fields. 

The Snowflake access history has different object structures https://docs.snowflake.com/en/user-guide/access-history.html , especially when it's copying for a STAGE or a local file. 

### 🤓 What?

- Add potential fields of access history log in the model as optional with default None
- Check for only table/view `objectDomain` before further processing

### 🧪 Tested?

tested against snowflake instances.

### DIFF
--- a/metaphor/snowflake/usage/extractor.py
+++ b/metaphor/snowflake/usage/extractor.py
@@ -1,12 +1,10 @@
 import logging
 import math
-from dataclasses import field
 from datetime import datetime
-from typing import Collection, Dict, List, Optional, Tuple
+from typing import Collection, Dict, List, Tuple
 
 from metaphor.models.metadata_change_event import DataPlatform, Dataset
-from pydantic import parse_raw_as
-from pydantic.dataclasses import dataclass
+from pydantic import BaseModel, Field, parse_raw_as
 
 from metaphor.common.event_util import ENTITY_TYPES
 from metaphor.common.extractor import BaseExtractor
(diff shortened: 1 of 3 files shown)


metaphor/snowflake/usage/extractor.py
```diff
@@ -31,10 +32,13 @@ class AccessedObjectColumn:
...
-    objectName: str
-    objectId: int
-    columns: List[AccessedObjectColumn]
+    objectDomain: str = ""
+    objectName: str = ""
+    objectId: int = 0
+    stageKind: Optional[str] = None
+    columns: List[AccessedObjectColumn] = field(default_factory=lambda: list())
+    location: Optional[str] = None
+    locations: Optional[List[str]] = None
```
elic-eon, line 41: Did Pydantic provide the option to drop extra fields?

mars-lan: +1. It might be more future-proof to ignore extra fields.

alyiwang: Tried this initially last night following the pydantic config https://pydantic-docs.helpmanual.io/usage/model_config/#options, but for a long time I couldn't get it to work and finally gave up. The default `extra=ignore` is already what we want, but its utility function just doesn't take this config and still enforce no extra field. Even explicitly set this config doesn't help

elic-eon: Great, maybe we should also have our configs inherited from the `BaseModel` to have better support.

elic-eon (approved):

PR #214, opened 2022-03-03
Fill in Dashboard SourceInfo URL [sc-6518]

alyiwang:
### 🤔 Why?

FE is using dashboard sourceInfo URL to provide link to the dashboard entity

### 🤓 What?

- Fill in Dashboard SourceInfo URL with the same value as `DashboardInfo.url`

### 🧪 Tested?

run the crawler locally, checked the results


pyproject.toml
```diff
@@ -10,7 +10,7 @@ packages = [
 
 [tool.poetry.dependencies]
 asyncpg = { version = "^0.25.0", optional = true }
-aws-assume-role-lib = "^1.3.0"
+aws-assume-role-lib = "^2.9.0"
```
mars-lan, line 13: Any breaking changes?

alyiwang: no mention of breaking change in github release.

mars-lan (approved):

PR #216, opened 2022-03-09
Add Power BI connector

elic-eon:
### 🤔 Why?

To support Power BI dashboards. Since we can get the Snowflake account from the `DataSource`, did not specify snowflake account in the config.

### 🤓 What?

1. Use [Power BI REST APIs](https://docs.microsoft.com/en-us/rest/api/power-bi/) to get metadata. The most important one is [GetScanResult](https://docs.microsoft.com/en-us/rest/api/power-bi/admin/workspace-info-get-scan-result).
2. Need read-only admin permission.

### 🧪 Tested?

Local run connector, wait for `metaphor-models` PR to verify MCEs.

PR #217, opened 2022-03-11
Bump metaphor models version

elic-eon:
### 🤔 Why?

The `metaphor-models` did not update in #216 

### 🤓 What?

- poetry update metaphor-models

### 🧪 Tested?

Install local build whl and successfully load Power BI extractor.


mars-lan (approved):

mars-lan:
How did the build pass if the model changes were not there?

elic-eon:
> How did the build pass if the model changes were not there?

Maybe we did add an end-to-end test for Power BI. CI does not actually import it. But mypy should complain about it?

mars-lan:
> > How did the build pass if the model changes were not there?
> 
> Maybe we did add an end-to-end test for Power BI. CI does not actually import it. But mypy should complain about it?

Yeah I'd assume mypy should be able to catch that.

PR #221, opened 2022-03-15
Add Snowflake lineage connector

mars-lan:
### 🤔 Why?

To extract lineage information from Snowflake's [ACCESS_HISTORY view](https://docs.snowflake.com/en/user-guide/access-history.html).

### 🤓 What?

- Add Snowflake lineage connector
- Move `AccessedObject` class to its own file so it can be shared between the Snowflake usage & lineage connectors 
- Add `unique_list` util method

### 🧪 Tested?

Manually tested against a production environment and verified the generated MCEs.


elic-eon (dismissed):
LGTM

elic-eon (approved):

PR #225, opened 2022-03-21
Add BigQuery query history extractor

mars-lan:
### 🤔 Why?

To support extracting recent queries for each BigQuery table.

### 🤓 What?

This connector is structurally very similar to [BigQuery lineage connector]() and processes a similar set of information from  [BigQuery Audit Logs](https://cloud.google.com/bigquery/docs/reference/auditlogs). However, there are a few noted differences that prevent the two to be combined into one:
- For the query connector, we're generally not interested in queries issued by service accounts, which can account for up to 90% of the volume. By filtering out these, we can speed up the process significantly.
- For the lineage connector, we're only interested in queries that result in a "write", whereas the query connector cares about all queries.
- The raw queries are attached to each table referenced (selected or written to), whereas lineage is only associated with tables that are written to. Mixing the two in the output can become messy and difficult to debug.
- 

### 🧪 Tested?

Manually ran against production and verified the output.


metaphor/bigquery/query/extractor.py
```diff
@@ -0,0 +1,211 @@
...
+    """BigQuery query history extractor"""
+
+    @staticmethod
+    def config_class():
+        return BigQueryQueryRunConfig
+
+    def __init__(self):
+        self._utc_now = start_of_day()
+        self._datasets: Dict[str, Dataset] = {}
+        self._datasets_pattern: List[re.Pattern[str]] = []
```
elic-eon, line 102: Should we drop this, seems not used.

mars-lan: Dropped. Thanks for catching.

PR #227, opened 2022-03-25
Optimize Snowflake profiler performance

mars-lan:
### 🤔 Why?

The current profiler can take an exceedingly long time to run against a large number of tables.

### 🤓 What?

This PR introduced 2 optimizations:
1. Make profiling of each type of column statistics configurable and disable the expensive statistics (unique count, average, standard deviation)  by default.
2. Batch process the entire `information_schema.columns` instead of issuing individual queries for each table. This reduces the time from 10+ minutes to 10s for 1000 tables.

### 🧪 Tested?

Finished profiling against 1500 tables within 10 mins, instead of hours w/o the optimization.

### DIFF
--- a/metaphor/snowflake/utils.py
+++ b/metaphor/snowflake/utils.py
@@ -86,7 +86,6 @@ def async_execute(
 
             if results_processor is None:
                 results_map[key] = results
-                print(f"result for {key}")
             else:
                 results_processor(key, results)
 
--- a/metaphor/snowflake/utils.py
+++ b/metaphor/snowflake/utils.py
@@ -86,6 +86,7 @@ def async_execute(
 
             if results_processor is None:
                 results_map[key] = results
+                print(f"result for {key}")
             else:
                 results_processor(key, results)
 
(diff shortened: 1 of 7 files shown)


metaphor/snowflake/utils.py
```diff
@@ -86,6 +86,7 @@ def async_execute(
 
             if results_processor is None:
                 results_map[key] = results
+                print(f"result for {key}")
```
elic-eon, line 89: Drop this?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.10.65"
+version = "0.10.66"
```
elic-eon, line 3:
```suggestion
version = "0.10.68"
```
It's `0.10.68` now.

metaphor/snowflake/utils.py
elic-eon, line 89:
```suggestion
```

elic-eon (approved):
LGTM

PR #228, opened 2022-03-26
Performance optimization for Snowflake usage connector

mars-lan:
### 🤔 Why?

The current query is suboptimal and won't scale up when running against a large `ACCESS_HISTORY` table.

### 🤓 What?

`ACCESS_HISTORY` & `QUERY_HISTORY` can potentially be huge and take a long time to query if not filtered properly. As noted in the [doc](https://docs.snowflake.com/en/sql-reference/account-usage/access_history.html#usage-notes), it's important to filter it by `QUERY_START_TIME` to avoid processing all the logs.

Also dropped the unused `QUERY_ID` in projection to reduce the payload.

TODO: Compeltely refactor the connector to pre-aggregate the table/column access counts on Snowflake to achieve orders of magnitude performance improvement.

### 🧪 Tested?

Verified that `SNOWFLAKE.ACCOUNT_USAGE.ACCESS_HISTORY` is properly filtered when `QUERY_START_TIME` is specified

Before (42s):

Screen Shot 2022-03-25 at 5 16 43 PM

After (33s):

Screen Shot 2022-03-25 at 5 17 44 PM

### DIFF
--- a/metaphor/snowflake/usage/extractor.py
+++ b/metaphor/snowflake/usage/extractor.py
@@ -94,16 +94,19 @@ class SnowflakeUsageExtractor(BaseExtractor):
             queries = {
                 x: QueryWithParam(
                     f"""
-                    SELECT q.QUERY_ID, START_TIME, DIRECT_OBJECTS_ACCESSED
+                    SELECT START_TIME, DIRECT_OBJECTS_ACCESSED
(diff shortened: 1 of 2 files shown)


elic-eon (approved):
LGTM

PR #233, opened 2022-03-29
Simplify Power BI connector setup by using read-only admin APIs exclusively

mars-lan:
### 🤔 Why?

[Read-only Power BI admin APIs](https://docs.microsoft.com/en-us/power-bi/enterprise/read-only-apis-service-principal-authentication#method) is a subset of [Power BI REST APIs](https://docs.microsoft.com/en-us/rest/api/power-bi/) that are [specifically designed](https://powerbi.microsoft.com/en-us/blog/announcing-new-admin-apis-and-service-principal-authentication-to-make-for-better-tenant-metadata-scanning/) for metadata extraction. 

By using the admin APIs exclusively, we no longer need to
1. Grant the service principal access to Power BI API
2. Add the service principal to each workspace

It also allows us to extract metadata from personal workspaces (aka "My workspace").

### 🤓 What?

- Replace all existing non-admin API calls with their admin counterpart
- Update README with simplified instructions
- Update tests & fix bugs in original tests

### 🧪 Tested?

Manually tested against a production environment.

### DIFF
--- a/metaphor/power_bi/README.md
+++ b/metaphor/power_bi/README.md
@@ -1,22 +1,23 @@
-# PowerBI Connector
+# Power BI Connector
 
-This connector extracts technical metadata from PowerBI workspaces using [Power BI REST APIs](https://docs.microsoft.com/en-us/rest/api/power-bi/).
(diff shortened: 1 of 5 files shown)


elic-eon (approved):
LGTM

PR #244, opened 2022-04-06
Add dbt cloud connector

mars-lan:
### 🤔 Why?

To extract dbt-related metadata directly from [dbt Cloud](https://docs.getdbt.com/docs/dbt-cloud/cloud-overview).

### 🤓 What?

This connector is essentially a wrapper around the [dbt core connector](https://github.com/MetaphorData/connectors/blob/main/metaphor/dbt/README.md). It uses [dbt Cloud Admin API](https://docs.getdbt.com/docs/dbt-cloud/dbt-cloud-api/admin-cloud-api) to download `manifest.json` generated from the latest successful run for a particular job and pass the file to dbt core connector for further processing.

### 🧪 Tested?

Added unit tests & manually verified.

### DIFF
--- a/metaphor/dbt/cloud/README.md
+++ b/metaphor/dbt/cloud/README.md
@@ -1,0 +1,38 @@
+# dbt Cloud Connector
+
+This connector extracts metadata from a [dbt cloud](https://www.getdbt.com/product/what-is-dbt/) project using the [Administrative API](https://docs.getdbt.com/docs/dbt-cloud/dbt-cloud-api/admin-cloud-api).
+
+## Setup
+
+We recommend creating a dedicated [service account token](https://docs.getdbt.com/docs/dbt-cloud/dbt-cloud-api/service-tokens) for the connector with minimally required permissions. This means `Read-only` permissions for Team plan and `Account Viewer` permissions for Enterprise plan.
+
+## Config File
+
+Create a YAML config file based the following template.
+
+### Required Configurations
+
+```yaml
+account_id: <dbt_cloud_account_id>
+job_id: <dbt_cloud_job_id>
+service_token: <service_account_token>
+output:
+  file:
+    directory: <output_directory>
+```
+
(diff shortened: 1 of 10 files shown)


metaphor/dbt/cloud/README.md
```diff
@@ -0,0 +1,38 @@
...
+```yaml
+account_id: <dbt_cloud_account_id>
+job_id: <dbt_cloud_job_id>
+service_token: <service_account_token>
+output:
+  file:
+    directory: <output_directory>
+```
+
+You can extract `account_id` & `job_id` from a particular dbt job's URL, which has the format `https://cloud.getdbt.com/#/accounts/<account_id>/projects/<project_id>/jobs/<job_id>/`.
```
elic-eon, line 24: Will an account have many jobs and do we need to have many crawlers run parallel?

metaphor/dbt/cloud/extractor.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def get_last_successful_run(self, job_id: int) -> Tuple[int, int]:
+        """Get the project and run IDs of the last successful run for a job"""
+
+        offset = 0
+        page_size = 50
+        while True:
+            # https://docs.getdbt.com/dbt-cloud/api-v2#operation/listRunsForAccount
+            resp = self._get(
+                "runs/", {"order_by": "-id", "limit": page_size, "offset": offset}
```
elic-eon, line 52: Looks [listRunsForAccount](https://docs.getdbt.com/dbt-cloud/api-v2#operation/listRunsForAccount) provides an optional job_definition_id for filtering jobId. Should we use it?

mars-lan: Good point. Use `job_definition_id` to filter instead.

metaphor/dbt/cloud/README.md
mars-lan: A job maps to a particular dbt project in the repo so typically there's only one. If there are multiple dbt projects, it's probably better to set up separate connectors than to mix the output into one.

elic-eon (approved):

PR #245, opened 2022-04-07
Write PowerBI Dataset name to logical id

elic-eon:
### 🤔 Why?

Since `metaphor-model` now support an additional field for virtual view logical ID, it would be simpler to use it directly instead of writing the name into the virtual view itself.

### 🤓 What?

- Write PowerBI Dataset name to logical ID
- Add workspace name to logical ID

### 🧪 Tested?

Change unit tests and manually run Power BI crawler.


elic-eon:
Since the name of the asset or workspace are not stable, we should use a different way to do this.

PR #247, opened 2022-04-19
[sc-7255] Support Looker 4.x API

elic-eon:
### 🤔 Why?

Looker API 3.x has been deprecated.

### 🤓 What?

- Update `looker-sdk` to latest.
- Use `Looker40SDK` in order to use Looker API 4.0

### 🧪 Tested?

Ran connector locally


mars-lan (commented):
Have you verified that the output stays identical?

elic-eon:
> Have you verified that the output stays identical?

Just verified. Version 4.0 will have empty string if there is not description, and 3.x will have `None`. So new MCEs will have description field with empty string. Otherwise, it's identical.

mars-lan:
> > Have you verified that the output stays identical?
> 
> Just verified. Version 4.0 will have empty string if there is not description, and 3.x will have `None`. So new MCEs will have description field with empty string. Otherwise, it's identical.

Got it. Please make sure the field is not annotated as `IsNotWhitespace` which won't accept empty strings.

elic-eon:
> Got it. Please make sure the field is not annotated as IsNotWhitespace which won't accept empty strings.

Manually ingest the MCE successfully. And it in not annotated as IsNotWhitespace

mars-lan (approved):

PR #251, opened 2022-04-26
Remove deprecated dashboard.url

elic-eon:
### 🤔 Why?

We must not write a URL to this field because `DashboardInfo.url` is deprecated.

### 🤓 What?

- Use `SourceInfo.main_url` instead

### 🧪 Tested?

Run `Looker` and `Metabase` crawler successfully


theseyi (approved):
tenor-108070580

PR #258, opened 2022-05-12
Making logging more pythonic

mars-lan:
### 🤔 Why?

There are a few logging calls that are not following Python's best practices.

### 🤓 What?

- Replace `logger.error` + `traceback.print_exc` with `logger.exception` as the latter will include the stack trace by default
- Replace `logger.error` + `logger.exception` with `logger.exception`
- Do not pass the exception as a second argument to `logger.error` as this causes `TypeError: not all arguments converted during string formatting` on Python 3.8.

### 🧪 Tested?

Ran a few connectors locally to verify.


alyiwang (approved):
LGTM
Thanks for fixing!

PR #259, opened 2022-05-12
bump airflow version

alyiwang:
### 🤔 Why?

Dependabot has shown several important/critical issues with airflow dependency.

image

### 🤓 What?

- Update airflow version to latest, run poetry update

### 🧪 Tested?

locally run several crawlers, they work fine


mars-lan (approved):
Thanks for fixing.

PR #262, opened 2022-05-24
Support dbt manifest v5

mars-lan:
### 🤔 Why?

dbt cloud has upgraded to 1.1.0 which generates v5 of manifest.json file which breaks the connector. Fortunately, the only material difference between v4 & v5 is the addition of [`unique_key` config](https://schemas.getdbt.com/dbt/manifest/v5/index.html#nodes_additionalProperties_oneOf_i0_config_unique_key), which doesn't affect the connector.

### 🤓 What?

- Generate v5 models
- Rename `ManifestParserV4` to `ManifestParserV5` and use it to parse both v4 & v5 manifest
- Add v5 manifest test data

### 🧪 Tested?

Tested using v5 manifest generated from dbt 1.1.0.


alyiwang (approved):
Thanks for updating this

PR #268, opened 2022-06-16
Expose dbt Snowflake tags within dataset [sc-9020]

alyiwang:
### 🤔 Why?

Snowflake tags carry important info about a table or column, we should capture those to complement the metaphor governed tags.

### 🤓 What?

- Fetch snowflake tags and insert into dataset schema
- Updated README to grant access to account_usage

NOTE that the tag_referenece views may have up to 2hr delay from snowflake. Newly added tags will not show up immediately.

### 🧪 Tested?

local run successfully, and verified the tags field in MCE


metaphor/snowflake/extractor.py
```diff
@@ -81,6 +81,11 @@ async def extract(self, config: SnowflakeRunConfig) -> Collection[ENTITY_TYPES]:
                 self._fetch_unique_keys(cursor, database)
                 self._fetch_table_info(conn, tables)
 
+            try:
+                self._fetch_tags(cursor)
+            except Exception as e:
+                logger.error(f"Fetch tags error: {e}")
```
mars-lan, line 87: Is this because of non-Enterprise version? Any more specific way to catch/test this?

mars-lan: Yeah please give it a try. If I remember correctly, the non-enterprise edition will return an empty table instead of throwing errors.

alyiwang: you are right, it won't throw error, just empty results. Removed the try-catch

metaphor/snowflake/extractor.py
```diff
@@ -269,6 +270,34 @@ def _fetch_primary_keys(self, cursor: SnowflakeCursor, database: str) -> None:
...
+                    f"Table {table} not found for tag {self._build_tag(key, value)}"
+                )
+                continue
+
+            if not dataset.schema.tags:
+                dataset.schema.tags = []
+            dataset.schema.tags.append(self._build_tag(key, value))
+
+    @staticmethod
+    def _build_tag(tag_key: str, tag_value: str):
```
mars-lan, line 298:
```suggestion
    def _build_tag(tag_key: str, tag_value: str) -> str:
```

mars-lan (commented):
Otherwise LGTM

mars-lan (approved):

PR #271, opened 2022-06-17
Add config-driven meta field mapping for dbt connector

mars-lan:
### 🤔 Why?

To allow customized mapping of owners & tags defined in dbt model's `meta` field.

### 🤓 What?

- Add `get_ownerships_from_meta` & `get_tags_from_meta` util methods to extract ownership & tag info from `meta`
- Update v3 & v5 manifest parser to extract ownership & tags info and populate `OwnershipAssignment` & `TagAssignment` for the materialized dataset.
- Drop legacy & unused model ownership extraction from `meta` 

### 🧪 Tested?

Unit tests & ran against dbt cloud


metaphor/dbt/README.md
```diff
@@ -43,6 +43,71 @@ docs_base_url: <docs_base_url>
...
 
+#### Ownership
+
+You can optionally specify the ownership for the materialized table or view using the [meta config](https://docs.getdbt.com/reference/resource-configs/meta) of a dbt model. For example:
+
+```yaml
+models:
+  - name: users
+    meta:
+      owner: joe@test.com
```
alyiwang, line 54: Do we want to use the new format for dbt 1.0+ as example?

```
config:
  meta:
     owner: xxx
```

metaphor/dbt/README.md
```diff
@@ -43,6 +43,71 @@ docs_base_url: <docs_base_url>
...
+```yaml
+meta_ownerships:
+  - meta_key: owner
+    ownership_type: Data Steward
+    email_domain: test.com
+```
+
+#### Tags
+
+Similar to [Ownership](#ownership), you can optionally specify the tag in meta. For example:
```
alyiwang, line 76: We probably want to mention this "tag" is different than the dbt tags https://docs.getdbt.com/reference/resource-configs/tags which is also defined under `meta`

metaphor/dbt/README.md
```diff
@@ -43,6 +43,71 @@ docs_base_url: <docs_base_url>
...
+
+```yaml
+models:
+  - name: users
+    meta:
```
alyiwang, line 81: Same as above, may want to add `config` at top level

metaphor/dbt/README.md
```diff
@@ -43,6 +43,71 @@ docs_base_url: <docs_base_url>
...
+      pii: true
```
alyiwang, line 82: For example, if value = false, do we still apply PII tag?

mars-lan: Nope. The default value matcher will match only `true`.

alyiwang: We can mention the value matches `true` by default in the readme

metaphor/dbt/config.py
```diff
@@ -1,10 +1,35 @@
...
+@dataclass
+class MetaTag:
+    # Key to match in the "meta" field
+    meta_key: str
+
+    # Type of the tag to assign
+    tag_type: str
+
+    # Regex to match the value
+    meta_value_matcher: str = "True"
```
alyiwang, line 30: Does this mean lower case `true` is not accepted?

metaphor/dbt/README.md
mars-lan: Good point. Updated.

metaphor/dbt/README.md
mars-lan: Done

metaphor/dbt/README.md
mars-lan: Added

metaphor/dbt/README.md
mars-lan: Added.

metaphor/dbt/config.py
mars-lan: The value is "stringified" here: https://github.com/MetaphorData/connectors/pull/271/files#diff-d4c962f232a7502d0070f236dc78552d8388c05ac8ad46d685dfc49b8c3b404cR113, so it'll turn into `True` in Python.

alyiwang (commented):
Otherwise, LGTM
great work!

metaphor/dbt/util.py
```diff
@@ -1,5 +1,5 @@
 import re
-from email.utils import parseaddr
+from asyncio.log import logger
```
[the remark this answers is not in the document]
mars-lan: Thanks for catching. VS Code auto imported the wrong module ;)

alyiwang (approved):

PR #272, opened 2022-06-21
duplicate snowflake tags [sc-9457]

alyiwang:
### 🤔 Why?

Query to snowflake account_usage views contains results for all DBs, so we should only do it once, not for each DB.

### 🤓 What?

- Move snowflake tags query outside each DB iteration, and only run once.

### 🧪 Tested?

Run it successfully at local, checked the output MCE


mars-lan:
Saw some errors in the log on tags not found for specific tables. Is it related?

alyiwang:
> Saw some errors in the log on tags not found for specific tables. Is it related?

Yes, the crawler fetches multiple DBs, and after fetching DB1, it queries tags for all DBs, then the tags for other DBs (not DB1) will not have the dataset initialized, so it logs table not found. Also fixed here.

mars-lan (approved):

PR #273, opened 2022-06-22
Fix Bigquery audit log parsing to handle QUERY jobs without a destination table

mars-lan:
### 🤔 Why?

Turns out not every [Query job](https://cloud.google.com/bigquery/docs/reference/auditlogs/rest/Shared.Types/BigQueryAuditMetadata#BigQueryAuditMetadata.JobConfig.Query) is guaranteed to have a `destinationTable` specified, e.g. a SCRIPT statement (i.e calling a stored procedure).

### 🤓 What?

Ignore log entries that don't have a `desitnationTable` set.

### 🧪 Tested?

Locally ran the connector to verify.


alyiwang (approved):
Thanks for the quick fix

PR #274, opened 2022-06-27
Add redshift recent query crawler [SC-9439]

elic-eon:
### 🤔 Why?

To extract recent query information from Redshift. Also refactor the part of code can be shared between usage and query.

### 🤓 What?

- Add redshift query connector

### 🧪 Tested?

Run the crawler in local and verified the output MCEs.


mars-lan (dismissed):
Thanks for the refactoring.

mars-lan:
Need to re-bump up the version again after merge.

mars-lan (approved):

PR #275, opened 2022-06-27
Thought spot sdk 1.11 breaks the crawler [SC-9482]

elic-eon:
### 🤔 Why?

Upgrade the `thoughtspot-rest-api-sd` to 1.11.0, and fix the version.

### 🤓 What?

- Use new enum type to accommodate the change of upstream.

### 🧪 Tested?

Run in local, and verified the output MCEs.


mars-lan (approved):

alyiwang (approved):
thanks for the fix

PR #278, opened 2022-06-30
Add description and platform to crawler run metadata [sc-7034]

alyiwang:
### 🤔 Why?

Add more information about the crawler

### 🤓 What?

- Add description and platform to crawler run metadata
- update pre-commit hooks version, since the older hooks don't work with the latest pre-commit (on python 3.10)
- also updated some boto libs

### 🧪 Tested?

run crawlers locally and checked the metadata file


metaphor/common/extractor.py
```diff
@@ -28,18 +29,30 @@ def config_class() -> Type[BaseConfig]:
     async def extract(self, config: BaseConfig) -> Collection[ENTITY_TYPES]:
         """Extract metadata and build messages, should be overridden"""
 
+    @abstractmethod
+    def platform(self) -> Optional[Platform]:
```
mars-lan, line 33: Why optional?

alyiwang: We didn't define platform for Slack and google directory, and manual lineage

pyproject.toml
```diff
@@ -31,9 +31,9 @@ requests = "^2.27.1"
 slack-sdk = { version = "^3.5.1", optional = true }
 smart-open = "^5.0.0"
 snowflake-connector-python = { version = "~=2.7.0", optional = true }
-sql-metadata = { version = "2.3.0", optional = true }
+sql-metadata = { version = "^2.6.0", optional = true }
 tableauserverclient = { version = "^0.17.0", optional = true }
-thoughtspot-rest-api-sdk = { version = "1.11.0", optional = true }
+thoughtspot-rest-api-sdk = { version = "~1.11.0", optional = true }
```
elic-eon, line 36: 👍

metaphor/redshift/query/extractor.py
```diff
@@ -20,11 +20,16 @@
 from metaphor.redshift.query.config import RedshiftQueryRunConfig
 
 logger = get_logger(__name__)
-logger.setLevel(logging.INFO)
 
 
 class RedshiftQueryExtractor(PostgreSQLExtractor):
-    """Redshift usage metadata extractor"""
+    """Redshift query extractor"""
```
elic-eon, line 26: Thanks!

elic-eon (approved):
LGTM. Just curious that why some extractor import is not sorted?

metaphor/thought_spot/extractor.py
```diff
@@ -53,6 +54,12 @@
 class ThoughtspotExtractor(BaseExtractor):
     """Thoughtspot metadata extractor"""
 
+    def platform(self) -> Optional[Platform]:
+        return Platform.THOUGHT_SPOT
+
+    def description(self) -> str:
+        return "Thought spot metadata crawler"
```
mars-lan, line 61:
```suggestion
        return "ThoughtSpot metadata crawler"
```

alyiwang:
> LGTM. Just curious that why some extractor import is not sorted?

I think by default `isort` only sorts imports within a group. Import groups are separated by blank lines, so if some imports are put in one group, they won't be automatically sorted to another group. The best practice is to have 3 groups: python native, third-party libraries, and other modules within the project.

elic-eon:
> I think by default `isort` only sorts imports within a group. Import groups are separated by blank lines, so if some imports are put in one group, they won't be automatically sorted to another group. The best practice is to have 3 groups: python native, third-party libraries, and other modules within the project.

Cool, I thought `isort` will automatically group import into three by default but it didn't for some cases.

PR #280, opened 2022-07-01
Extract elapsedTime for queries [SC-9645]

elic-eon:
### 🤔 Why?

Extract the elapsed time of query for `BigQueryQueryExtractor`, `SnowflakeQueryExtractor` and `RedshiftQueryExtractor`.

### 🤓 What?

- Fill elapsedTime for 3 platform query extractor.
- Fix BigQuery usage & query crawler issue. `logEvent. from_entry` return `None` if destination table is not valid in that query.

### 🧪 Tested?

- Run 3 query extractor and BigQuery usage/lineage extractor.


pyproject.toml
```diff
@@ -21,7 +21,7 @@ google-cloud-bigquery = { version = "^2.21.0", optional = true }
 google-cloud-logging = { version = "^2.7.0", optional = true }
```
mars-lan, line 21: Also bump up version?

elic-eon: version 3?

metaphor/bigquery/query/extractor.py
```diff
@@ -95,9 +95,14 @@ def _parse_job_change_entry(self, entry: LogEntry) -> None:
 
             self._table_queries.store_recent_query(
                 table_name,
-                job_change.timestamp,
-                job_change.query,
-                job_change.user_email,
+                QueryInfo(
+                    elapsed_time=(
```
mars-lan, line 99: Make sure this returns float otherwise it'll throw conversion errors

pyproject.toml
mars-lan: I mean the version for the whole project. GitHub doesn't allow me to add comments to non-changed lines.

elic-eon: Done. I thought I have bumped up the version 😵‍💫

metaphor/bigquery/query/extractor.py
elic-eon: Added a guard to make sure both are `datetime`

mars-lan (approved):

PR #281, opened 2022-07-01
Build Docker image directly from source instead of from PyPI

mars-lan:
### 🤔 Why?

There is a delay (sometime hours) before the newly published package on PyPI becomes available. As a result, the current Docker image build process ends up installing the previous version.

### 🤓 What?

Refactor Dockerfile and the CD pipeline to install directly from the latest source and tag it with the correct version number.

### 🧪 Tested?

CD: https://github.com/MetaphorData/connectors/actions/runs/2599490786
Published: https://hub.docker.com/r/metaphordata/connectors/tags

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -19,9 +19,35 @@ defaults:
     shell: bash
 
 jobs:
+  version:
+    name: Get version
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get-version.outputs.version }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+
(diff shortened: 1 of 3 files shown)


alyiwang (commented):
Otherwise, LGTM

.github/workflows/cd.yml
```diff
@@ -19,9 +19,35 @@ defaults:
...
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get-version.outputs.version }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: ${{ env.NODE_VERSION }}
```
alyiwang, line 33: NODE_VERSION: 14 -> 16

mars-lan: Updated

alyiwang (approved):
LGTM

PR #282, opened 2022-07-06
Allow specifying BigQuery credential directly [sc-9738]

alyiwang:
### 🤔 Why?

Provides the option to specify the BigQuery credentials from the key file directly in the config, and bypass the key file upload.

### 🤓 What?

- Update BigQuery configs to take either key_path or credentials
- Update readme
- bump up bigquery library versions

### 🧪 Tested?

tested bigquery and bigquery usage crawlers locally using both config formats


metaphor/bigquery/utils.py
```diff
@@ -25,29 +25,38 @@
...
-        credentials = service_account.Credentials.from_service_account_info(
-            json.load(fin),
-            scopes=["https://www.googleapis.com/auth/cloud-platform"],
-        )
-
-        return logging_v2.Client(
-            credentials=credentials,
-            project=project_id if project_id else credentials.project_id,
-        )
+    raise ValueError(
```
mars-lan, line 57: Might be better to use a pydantic validator? https://pydantic-docs.helpmanual.io/usage/validators/

mars-lan (commented):
Otherwise LGTM

metaphor/bigquery/utils.py
alyiwang: good point, updated

mars-lan: Root validator might be more appropriate here? https://pydantic-docs.helpmanual.io/usage/validators/#root-validators

alyiwang: thanks, updated

PR #283, opened 2022-07-07
Add pydantic validtor to snowflake auth config

alyiwang:
### 🤔 Why?

To ensure either password or private key is set in the config

### 🤓 What?

- add pydantic validator to ensure either password or private key is set in the config

### 🧪 Tested?

unit tests of the configs

### DIFF
--- a/metaphor/snowflake/auth.py
+++ b/metaphor/snowflake/auth.py
@@ -2,7 +2,7 @@ from typing import Optional
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
-from pydantic import validator
+from pydantic import root_validator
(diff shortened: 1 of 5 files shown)


metaphor/snowflake/auth.py
```diff
@@ -45,8 +46,16 @@ class SnowflakeAuthConfig(BaseConfig):
     # the query tags for each snowflake query the connector issues
     query_tag: Optional[str] = METAPHOR_DATA
 
+    @validator("private_key")
```
mars-lan, line 49: Consider using root validator instead: https://pydantic-docs.helpmanual.io/usage/validators/#root-validators

mars-lan (changes requested):

mars-lan (approved):

PR #284, opened 2022-07-07
Install lumigo_opentelemetry when building the Docker image

mars-lan:
### 🤔 Why?

To allow optional monitoring using Lumigo. See https://docs.lumigo.io/docs/containerized-applications for more details.

### 🤓 What?

Also updated the developer doc.

### 🧪 Tested?

Manually tested.


alyiwang (commented):
Do we need any configuration for the tracer to work?

alyiwang (approved):

PR #287, opened 2022-07-10
Tableau crawler preview error [sc-9553]

alyiwang:
### 🤔 Why?

If set `disable_preview_image: true`, we didn't populate the preview image, but when building chart preview URL, it still fetches the `view.preview_image` field, which will throw this error if not populated. This doesn't affect the crawler output, just produces many unnecessary error messages.

### 🤓 What?

- Check if preview disabled when building chart preview URL

### 🧪 Tested?

Locally tested the crawler with preview disable, no errors.


mars-lan (approved):
Thanks for the quick fix!

PR #290, opened 2022-07-12
Add manual ownership & tag connector

mars-lan:
### 🤔 Why?

Allow bulk assigning governance information such as ownership & tag information.

### 🤓 What?

As titled. Also factor out the shared `DeserializableDatasetLogicalID` model between manual lineage and governance connectors.

### 🧪 Tested?

Unit tests

### DIFF
--- a/metaphor/manual/governance/README.md
+++ b/metaphor/manual/governance/README.md
@@ -1,0 +1,85 @@
+# Manual Governance Connector
+
+This connector assigns manually specified ownership and tagging information to tables.
+
+## Setup
+
+No special setup is required.
+
+## Config File
+
+Create a YAML config file based on the following template.
+
+> Note: The ownership types and tags must be first created on Metaphor for the assignment to take effect.
+
+### Required Configurations
+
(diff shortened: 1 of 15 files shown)


metaphor/manual/governance/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+> Note: The ownership types and tags must be first created on Metaphor for the assignment to take effect.
+
+### Required Configurations
+
+```yaml
+datasets:
+  - id:
+      platform: <data_platform>
+      name: <dataset_name>
```
alyiwang, line 21: Do we have to specify each table? If we have 100 tables, maybe a wildcard can be helpful.

alyiwang (approved):
LGTM!

metaphor/manual/governance/README.md
mars-lan: For now it's one dataset at a time as the ingestion doesn't support wildcard matching. We can add that capability in the future if required.

PR #294, opened 2022-07-20
Add optional role config for Snowflake connectors

mars-lan:
### 🤔 Why?

To allow users to connect using a different role, in case the default one doesn't work.

### 🤓 What?

As titled. Also simplify the `connect` method.

### 🧪 Tested?

Locally tested with & without the `role` config set.

### DIFF
--- a/metaphor/snowflake/auth.py
+++ b/metaphor/snowflake/auth.py
@@ -40,6 +40,9 @@ class SnowflakeAuthConfig(BaseConfig):
     # if using key pair authentication
     private_key: Optional[SnowflakeKeyPairAuthConfig] = None
 
+    # role to use when opening a connection
+    role: Optional[str] = None
+
     # database context when opening a connection
     default_database: Optional[str] = None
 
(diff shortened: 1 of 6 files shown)


alyiwang (approved):
LGTM

metaphor/snowflake/auth.py
```diff
@@ -89,17 +77,16 @@ def connect(config: SnowflakeAuthConfig) -> snowflake.connector.SnowflakeConnect
...
-            },
-        )
-
-    raise ValueError(
-        "Invalid Snowflake configuration, please set either password or private key"
+    return snowflake.connector.connect(
+        account=config.account,
+        user=config.user,
+        password=config.password,
+        private_key=pkb,
```
alyiwang, line 84: I just thought of an edge case: if a user provided both password and key, what will happen? Maybe we should add this to the validator too.

mars-lan: Good point. Internally it must prefer one over the other but we should fail loudly in that case. Will create a follow-up PR.

mars-lan: https://github.com/MetaphorData/connectors/pull/296

PR #295, opened 2022-07-20
Skip empty Snowflake databases to avoid crashing

mars-lan:
### 🤔 Why?

To avoid crashing due to invalid SQL in `_fetch_table_info` for empty databases.

### 🤓 What?

As titled. Also proactively catch and log exceptions when running the assembled query in `_fetch_table_info`, since a well-formed query can still fail against a shared database.

### 🧪 Tested?

Manually tested against an empty database and a shared database.


alyiwang (approved):
LGTM

PR #296, opened 2022-07-20
Add must_set_exactly_one util method for config validation

mars-lan:
### 🤔 Why?

Factor out common validation logic and make sure that it checks exactly one, not at least one.

### 🤓 What?

As titled.

### 🧪 Tested?

Unit tests.


alyiwang (approved):
👍

PR #297, opened 2022-07-20
Support using glob patterns in filters

mars-lan:
### 🤔 Why?

Sometimes it's useful to include/exclude databases/schemas/tables using wildcard (glob) patterns. For example, use `staging_*` to filter out all tables with a `staging_` prefix.

### 🤓 What?

- Refactor `filter.include_table` to support glob patterns
- Move filter config doc to `common/docs` for consistency

### 🧪 Tested?

Unit test and manually tested against production.

### DIFF
--- a/metaphor/bigquery/README.md
+++ b/metaphor/bigquery/README.md
@@ -51,7 +51,7 @@ output:
     directory: <output_directory>
 ```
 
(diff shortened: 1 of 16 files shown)


alyiwang (approved):
LGTM

PR #298, opened 2022-07-21
Missing BQ SQL definition statements [sc-9930]

alyiwang:
### 🤔 Why?

The BigQuery REST API only returns DDLs for (materialized) views: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#Table

We need to query against `INFORMATION_SCHEMA.TABLES` to get DDL https://cloud.google.com/blog/topics/developers-practitioners/spring-forward-bigquery-user-friendly-sql

### 🤓 What?

- Fetch table DDL from `INFORMATION_SCHEMA.TABLES`

### 🧪 Tested?

Tested locally against bigquery instance, verified DDL in MCE

### DIFF
--- a/metaphor/bigquery/extractor.py
+++ b/metaphor/bigquery/extractor.py
@@ -55,19 +55,36 @@ class BigQueryExtractor(BaseExtractor):
 
         with ThreadPoolExecutor(max_workers=config.max_concurrency) as executor:
 
-            def list_table(dataset_ref) -> Iterable[Dataset]:
+            def list_table(dataset_ref: bigquery.DatasetReference) -> List[Dataset]:
                 def get_table(table) -> Dataset:
                     bq_table = client.get_table(table)
                     return self._parse_table(client.project, bq_table)
 
-                return executor.map(
-                    get_table,
-                    BigQueryExtractor._list_tables_with_filter(
-                        dataset_ref, client, dataset_filter
-                    ),
-                )
+                datasets = [
+                    d
+                    for d in executor.map(
+                        get_table,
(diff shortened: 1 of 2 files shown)


elic-eon (dismissed):

metaphor/bigquery/extractor.py
```diff
@@ -55,17 +55,32 @@ async def extract(self, config: BigQueryRunConfig) -> Collection[ENTITY_TYPES]:
...
+                        ),
+                    )
+                }
+
+                table_ddl = client.query(
+                    f"select table_name, ddl from `{dataset_ref.project}.{dataset_ref.dataset_id}.INFORMATION_SCHEMA.TABLES`"
+                ).result()
+
+                for table_name, ddl in table_ddl:
+                    dataset = datasets[table_name]
```
mars-lan, line 79: May need to normalize table_name?

metaphor/bigquery/extractor.py
```diff
@@ -55,17 +55,32 @@ async def extract(self, config: BigQueryRunConfig) -> Collection[ENTITY_TYPES]:
...
+                    assert dataset, f"table {table_name} not found for DDL"
```
mars-lan, line 80: Maybe log instead of throwing exceptions?

metaphor/bigquery/extractor.py
alyiwang: good point, updated

metaphor/bigquery/extractor.py
alyiwang: updated

mars-lan (approved):

PR #301, opened 2022-07-22
Extract PowerBI workspace & app info

mars-lan:
### 🤔 Why?

To better organize Power BI dashboards, reports, and datasets based on the workspace & app they belong to.

### 🤓 What?

As titled. Also refactored the tests for improved consistency.

### 🧪 Tested?

Manually tested against a production Power BI deployment.

PR #302, opened 2022-07-23
Add timeout to extractor execution

mars-lan:
### 🤔 Why?

To prevent a stuck or mis-configured connector from running forever.

### 🤓 What?

The original plan was to add a common `timeout` config option with a default value. However, it turns out that doing so [is difficult before Python 3.10](https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses). Since this is more of a fail-safe mechanism than a configuration option, overriding it using an environment variable seems like a perfectly acceptable solution.

### 🧪 Tested?

```
% TIMEOUT=2 python -m metaphor.bigquery ~/Desktop/local/bigquery.yml
...
Traceback (most recent call last):
  File "/Users/marslan/.pyenv/versions/3.9.6/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/marslan/.pyenv/versions/3.9.6/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/marslan/Metaphor/connectors/metaphor/bigquery/__main__.py", line 6, in <module>
    cli_main("BigQuery metadata extractor", BigQueryExtractor)
  File "/Users/marslan/Metaphor/connectors/metaphor/common/cli.py", line 19, in cli_main
    extractor.run(config=config_cls.from_yaml_file(args.config))
  File "/Users/marslan/Metaphor/connectors/metaphor/common/extractor.py", line 52, in run
    raise TimeoutError(f"Process killed after {timeout} seconds")
TimeoutError: Process killed after 2.0 seconds
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.10.123"
+version = "0.10.122"
 license = "Apache-2.0"
 description = "A collection of Python-based 'connectors' that extract metadata from various sources to ingest into the Metaphor app."
 authors = ["Metaphor <dev@metaphor.io>"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.10.121"
+version = "0.10.123"
 license = "Apache-2.0"
 description = "A collection of Python-based 'connectors' that extract metadata from various sources to ingest into the Metaphor app."
 authors = ["Metaphor <dev@metaphor.io>"]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.10.121"
+version = "0.10.123"
```
alyiwang, line 3: 0.10.122 ?

alyiwang (approved):
Otherwise, LGTM

pyproject.toml
mars-lan: I was hoping https://github.com/MetaphorData/connectors/pull/301 will get in first :)

alyiwang (approved):
LGTM

PR #305, opened 2022-07-27
PowerBI: local variable 'platform' referenced before assignment

elic-eon:
### 🤔 Why?

There is one if-else case not assign the variable.

### 🤓 What?

- Throw an exception in that case
- Small refactor on catching errors

### 🧪 Tested?

Verified the crawler locally.


mars-lan (approved):
Thanks for the quick fix.

PR #306, opened 2022-07-31
Remove lumigo dependency

mars-lan:
### 🤔 Why?

To keep the open-source version clean.

### 🤓 What?

As titled. Also added `.dockerignore` to speed up the build and remove source code after installation to reduce image size.

### 🧪 Tested?

Manually built the image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
 FROM python:3.8-slim
 
-ADD . /src 
+COPY . /src
 
 RUN pip install '/src[all]'
-RUN pip install 'lumigo_opentelemetry'
+
+RUN rm -rf /src
 
 CMD ["sh", "-c", "python -m ${PY_MODULE} ${CONFIG_FILE}"]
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,8 +1,9 @@
 FROM python:3.8-slim
 
-ADD . /src 
+COPY . /src
 
 RUN pip install '/src[all]'
-RUN pip install 'lumigo_opentelemetry'
+
+RUN rm -rf /src
```
alyiwang, line 7: why do we need to remove `src`

alyiwang (commented):
LGTM

Dockerfile
mars-lan: To reduce image size. `pip install` has already installed the `metaphor` package under `/usr/local/lib/python3.8/site-packages` so it's safe to remove the source codes.

alyiwang (approved):

PR #307, opened 2022-08-01
Add major & minor version tags to published Docker image

mars-lan:
### 🤔 Why?

So the image can be imported in a sem-ver fashion, e.g. `metaphordata/connectors:0.10`.

### 🤓 What?

As titled.

### 🧪 Tested?

Manually invoked the CD workflow and verified the tags on Docker Hub: https://hub.docker.com/r/metaphordata/connectors/tags


alyiwang (approved):
LGTM
Thanks for adding this

PR #311, opened 2022-08-02
Add workspace id & name to Power BI datasets

mars-lan:
### 🤔 Why?

https://github.com/MetaphorData/connectors/pull/301 missed adding workspace ID & name to Power BI datasets.

### 🤓 What?

As titled. Also dropped `PowerBIExtractor.parse_power_query` as it has been moved to `power_query_parser.py` .

### 🧪 Tested?

Manually tested against product and verified the output.

### DIFF
--- a/metaphor/power_bi/extractor.py
+++ b/metaphor/power_bi/extractor.py
@@ -1,5 +1,4 @@
 import json
-import re
 import tempfile
 from time import sleep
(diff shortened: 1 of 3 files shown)


elic-eon (approved):
LGTM

alyiwang (approved):
Thanks for the quick fix!

PR #314, opened 2022-08-02
Log more crawler run information

alyiwang:
### 🤔 Why?

To add debugging process

### 🤓 What?

- Add logging from config file path
- Add missing snowflake CLI

### 🧪 Tested?

tested in local run


metaphor/snowflake/__init__.py
```diff
@@ -0,0 +1,6 @@
+from metaphor.common.cli import cli_main
```
mars-lan, line 1: Thanks for adding

mars-lan (approved):

PR #316, opened 2022-08-05
Add option to filter self-loops in BigQuery lineage

mars-lan:
### 🤔 Why?

We currently filter out BigQuery query log entries where the destination table appears in the source tables. This can lead to undesirable missing lineage.

### 🤓 What?

Add a config option to control if self-reference should be filtered out, instead of dropping the query log entry altogether.

### 🧪 Tested?

Verified against production.


alyiwang (approved):
LGTM

PR #317, opened 2022-08-05
Remove unused dependencies and extras

mars-lan:
### 🤔 Why?

As we've already dropped slack & google directory connectors, it's time to clean up the dependencies.

### 🤓 What?

As titled.

### 🧪 Tested?

N/A

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.11.5"
+version = "0.11.7"
 license = "Apache-2.0"
 description = "A collection of Python-based 'connectors' that extract metadata from various sources to ingest into the Metaphor app."
 authors = ["Metaphor <dev@metaphor.io>"]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -19,8 +19,6 @@ botocore = "^1.27.20"
 canonicaljson = "^1.4.0"
 fastjsonschema = "^2.15.1"
 GitPython = "^3.1.0"
-google-api-python-client = { version = "^2.36.0", optional = true }
-google-auth-oauthlib = { version = "^0.5.0", optional = true }
```
alyiwang, line 23: Oauth is still used?
```
try:
    import google.cloud.bigquery as bigquery
    from google.cloud import logging_v2
    from google.oauth2 import service_account
except ImportError:
    print("Please install metaphor[bigquery] extra\n")
    raise
```

I guess it might be included in the google-bigquery anyway.

mars-lan: Yup. It'll get pulled in transitively. Verified that bigquery connector still functions as expected after removing this.

alyiwang (approved):

Issue #310, opened 2022-08-02
Breaking changes in 0.12


mars-lan:
- https://github.com/MetaphorData/connectors/pull/520
- https://github.com/MetaphorData/connectors/pull/521
- https://github.com/MetaphorData/connectors/pull/526
- https://github.com/MetaphorData/connectors/pull/522
- https://github.com/MetaphorData/connectors/pull/525
- https://github.com/MetaphorData/connectors/pull/524
- upgrade Pydantic to 1.10
- upgrade airflow

alyiwang:
version 0.12 released

PR #318, opened 2022-08-07
Speed up BigQuery profiler by turning off expensive column statistics by default

mars-lan:
### 🤔 Why?

BigQuery profiler can take a long time to run for large tables. Disable computing of expensive column statistics like we did for Snowflake profiler https://github.com/MetaphorData/connectors/pull/227.

### 🤓 What?

- Move `ColumnStatistics` class to `metaphor.common` package and add README file
- Add `column_statistics` config and construct BigQuery profiling query based on the settings 

### 🧪 Tested?

Tested again production. A previous run that timed out after an hour now completed in 15 minutes.

### DIFF
--- a/metaphor/bigquery/profile/extractor.py
+++ b/metaphor/bigquery/profile/extractor.py
@@ -10,9 +10,9 @@ except ImportError:
     print("Please install metaphor[bigquery] extra\n")
     raise
 
-
 from metaphor.bigquery.extractor import BigQueryExtractor, build_client
 from metaphor.bigquery.profile.config import BigQueryProfileRunConfig, SamplingConfig
+from metaphor.common.column_statistics import ColumnStatistics
 from metaphor.common.event_util import ENTITY_TYPES
 from metaphor.common.extractor import BaseExtractor
 from metaphor.common.filter import DatasetFilter
@@ -46,6 +46,7 @@ class BigQueryProfileExtractor(BaseExtractor):
 
     def __init__(self):
         self._sampling = None
+        self._column_statistics = None
 
     async def extract(
         self, config: BigQueryProfileRunConfig
@@ -55,6 +56,7 @@ class BigQueryProfileExtractor(BaseExtractor):
         logger.info("Fetching usage info from BigQuery")
 
         self._sampling = config.sampling
+        self._column_statistics = config.column_statistics
 
         client = build_client(config)
         tables = self._fetch_tables(client, config)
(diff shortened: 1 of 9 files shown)


metaphor/bigquery/profile/extractor.py
```diff
@@ -202,24 +212,102 @@ def _build_profiling_query(
...
+                    std_dev=std_dev,
+                )
+            )
+
+        # Verify that we've consumed all the elements from results
+        assert index == len(
+            results
+        ), f"Unconsumed elements from results for {dataset.logical_id}"
+
+        """
```
alyiwang, line 289: Removed the commented part?

metaphor/bigquery/profile/extractor.py
```diff
@@ -168,29 +174,33 @@ def _build_profiling_query(
...
 
-            if not BigQueryProfileExtractor._is_complex(data_type):
+            if (
+                column_statistics.unique_count
+                and not BigQueryProfileExtractor._is_complex(data_type)
+            ):
                 query.append(f", COUNT(DISTINCT `{column}`)")
 
-            if nullable:
+            if column_statistics.null_count:
```
alyiwang, line 192: we don't consider if a field is nullable anymore?

metaphor/bigquery/profile/extractor.py
mars-lan: Thanks for catching. Removed.

metaphor/bigquery/profile/extractor.py
mars-lan: That's correcrt. We'll issue the query regardless the nullability of the column since it's cheap to do and avoids extra logic in code.

alyiwang (approved):
LGTM!

metaphor/common/column_statistics.py
```diff
@@ -0,0 +1,22 @@
...
+    min_value: bool = True
+
+    # Compute max value
+    max_value: bool = True
+
+    # Compute average value
+    avg_value: bool = False
+
+    # Compute standard deviation
+    std_dev: bool = False
```
alyiwang, line 22: One potential issue is when we enabled `sampling`. In that case, only avg and std_dev still makes sense, but they are turned off by default

mars-lan: Yeah I plan to drop the whole sampling thing: https://github.com/MetaphorData/connectors/issues/310

PR #319, opened 2022-08-08
Snowflake usage ambiguous column name USER [sc-11092]

alyiwang:
### 🤔 Why?

ambiguous column name 'USER_NAME' when using user filter

### 🤓 What?

- add table qualifier for the 'USER_NAME' column

### 🧪 Tested?

tested against production instance


mars-lan:
Should probably start adding some tests? 😉

mars-lan (approved):

PR #320, opened 2022-08-09
Redshift lineage "AttributeError: 'NoneType' object has no attribute 'rstrip'

elic-eon:
### 🤔 Why?

As title.

### 🤓 What?

- Check if object is a string before calling `rstrip`

### 🧪 Tested?

```
$ metaphor redshift.lineage configs/redshift_lineage.yml
```

Verified the command without error, output MCE is correct.

### DIFF
--- a/metaphor/redshift/lineage/extractor.py
+++ b/metaphor/redshift/lineage/extractor.py
@@ -147,9 +147,7 @@ class RedshiftLineageExtractor(PostgreSQLExtractor):
             source_table_name = f"{db}.{row['source_schema']}.{row['source_table']}"
             target_table_name = f"{db}.{row['target_schema']}.{row['target_table']}"
             query = (
-                row["querytxt"].rstrip()
-                if row.get("querytxt") is not None
-                else None
+                row["querytxt"].rstrip() if row.get("querytxt") is not None else None
             )
 
             source_id = str(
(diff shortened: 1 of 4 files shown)


metaphor/redshift/lineage/extractor.py
```diff
@@ -146,7 +146,11 @@ async def _fetch_lineage(self, sql, conn, db):
         for row in results:
             source_table_name = f"{db}.{row['source_schema']}.{row['source_table']}"
             target_table_name = f"{db}.{row['target_schema']}.{row['target_table']}"
-            query = row["querytxt"].rstrip() if "querytxt" in row else None
+            query = (
+                row["querytxt"].rstrip()
+                if "querytxt" in row and type(row["querytxt"]) == str
```
mars-lan, line 151: I'd imagine the only possibility for "querytext" is either str or None? 

```suggestion
                row.get("querytxt") is not None
```

metaphor/redshift/lineage/extractor.py
```diff
@@ -147,9 +147,7 @@ async def _fetch_lineage(self, sql, conn, db):
             source_table_name = f"{db}.{row['source_schema']}.{row['source_table']}"
             target_table_name = f"{db}.{row['target_schema']}.{row['target_table']}"
             query = (
-                row["querytxt"].rstrip()
-                if row.get("querytxt") is not None
-                else None
+                row["querytxt"].rstrip() if row.get("querytxt") is not None else None
```
mars-lan, line 150: Oops sorry shouldn't have suggested double negation

```suggestion
                None if row.get("querytxt") is None else row["querytxt"].rstrip()
```

elic-eon: Your suggestion was actually correct. I also updated test cases.

mars-lan (approved):

PR #322, opened 2022-08-12
Reduce the default lookback_days to 7 for various connectors

mars-lan:
### 🤔 Why?

A 30-day lookback period seems excessive as a default. 7 days of lineage/usage/queries is sufficient in most cases.

### 🤓 What?

As titled. Also updated tests and READMEs.

### 🧪 Tested?

Unit tests.


alyiwang (approved):
LGTM

PR #323, opened 2022-08-14
BigQuery KeyError snapshot sync configuration [sc-11102]

alyiwang:
### 🤔 Why?

used python `dict[]` by mistake, which throws keyerror if the key is not found. 

### 🤓 What?

- use `dict.get()` since we are checking for `None` value in the next line

### 🧪 Tested?

tested against production instance.

### DIFF
--- a/metaphor/bigquery/extractor.py
+++ b/metaphor/bigquery/extractor.py
@@ -93,7 +93,7 @@ class BigQueryExtractor(BaseExtractor):
 
             for table_name, ddl in table_ddl:
                 table = tables.get(str(table_name).lower())
-                if not table:
+                if table is None:
                     logger.error(f"table {table_name} not found for DDL")
                     continue
                 table.schema.sql_schema.table_schema = ddl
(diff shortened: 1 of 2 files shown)


mars-lan:
Was the `poetry.lock` change necessary? Can we do that in a separate PR?

alyiwang:
> Was the `poetry.lock` change necessary? Can we do that in a separate PR?

poetry complains that my local versions are outdated compared to pyproject.toml, and asks me to do an update. I thought it should just update from poetry lock, but it actually updated all the things. I will revert that part

alyiwang:
> Was the `poetry.lock` change necessary? Can we do that in a separate PR?

cleaned up

metaphor/bigquery/extractor.py
```diff
@@ -92,7 +92,7 @@ def get_table(table: bigquery.TableReference) -> Dataset:
             ).result()
 
             for table_name, ddl in table_ddl:
-                table = tables[str(table_name).lower()]
+                table = tables.get(str(table_name).lower())
                 if not table:
```
mars-lan, line 96:
```suggestion
                if table is None:
```

mars-lan (approved):

PR #324, opened 2022-08-14
Add optional job_project_id config to BigQuery crawlers

mars-lan:
### 🤔 Why?

To allow running BigQuery jobs using a project ID different from the default.

### 🤓 What?

As titled.

### 🧪 Tested?

Manually tested against a production deployment.

### DIFF
--- a/metaphor/bigquery/extractor.py
+++ b/metaphor/bigquery/extractor.py
@@ -88,7 +88,8 @@ class BigQueryExtractor(BaseExtractor):
 
             logger.info(f"Getting table DDL for {dataset_ref}")
             table_ddl = client.query(
-                f"select table_name, ddl from `{dataset_ref.project}.{dataset_ref.dataset_id}.INFORMATION_SCHEMA.TABLES`"
+                f"select table_name, ddl from `{dataset_ref.project}.{dataset_ref.dataset_id}.INFORMATION_SCHEMA.TABLES`",
+                project=config.job_project_id,
             ).result()
 
             for table_name, ddl in table_ddl:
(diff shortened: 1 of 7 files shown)


metaphor/bigquery/profile/extractor.py
```diff
@@ -55,6 +56,7 @@ async def extract(
 
         logger.info("Fetching usage info from BigQuery")
 
+        self._job_project_id = config.project_id
```
alyiwang, line 59: Should use `config.job_project_id` and fallback to `config.project_id`?

metaphor/bigquery/extractor.py
```diff
@@ -88,7 +88,8 @@ def get_table(table: bigquery.TableReference) -> Dataset:
 
             logger.info(f"Getting table DDL for {dataset_ref}")
             table_ddl = client.query(
-                f"select table_name, ddl from `{dataset_ref.project}.{dataset_ref.dataset_id}.INFORMATION_SCHEMA.TABLES`"
+                f"select table_name, ddl from `{dataset_ref.project}.{dataset_ref.dataset_id}.INFORMATION_SCHEMA.TABLES`",
+                project=config.job_project_id,
```
alyiwang, line 92: Should use `config.job_project_id` and fallback to `config.project_id`?

mars-lan: If `job_project_id` is None,`client` will use the project ID it's created with, which is `config.project_id` or the credential's project ID: https://github.com/MetaphorData/connectors/blob/main/metaphor/bigquery/utils.py#L31

metaphor/bigquery/profile/extractor.py
alyiwang, line 59:
```suggestion
        self._job_project_id = config.job_project_id
```

mars-lan: Thanks for catching.

metaphor/bigquery/extractor.py
alyiwang: I see 👍

alyiwang (approved):
LGTM

PR #326, opened 2022-08-16
unsupported manifest schema v6 for dbt cloud [sc-11481]

alyiwang:
### 🤔 Why?

dbt 1.2 use manifest schema v6, which has breaking changes in dbt metrics.

### 🤓 What?

- since the previous version manifest.json can't be parsed by new version schema, have to make a copy of the parse using generated V6 schema.

### 🧪 Tested?

tested with manifest.json generated by dbt1.2


mars-lan (approved):
Thanks for the quick fix!

PR #329, opened 2022-08-18
Extract subfields from BigQuery Record type [sc-11549]

alyiwang:
### 🤔 Why?

We need to capture the nested fields within BIgQuery Record data type

### 🤓 What?

- extract Record data type subfields recursively

### 🧪 Tested?

unit tests
tested against bigquery instance and verified results


metaphor/common/fieldpath.py
```diff
@@ -0,0 +1,30 @@
...
+    MAP = "MAP"
+    UNION = "UNION"
+
+
+def build_field_path(
+    parent_field_path: str, field_name: str, field_type: FieldDataType
+) -> str:
+    path = parent_field_path
+    field_name_encoded = (
+        field_name.replace(".", "%2E").replace("<", "%3C").replace(">", "%3E")
```
mars-lan, line 17: Why not just use URL encoding directly?

metaphor/bigquery/extractor.py
```diff
@@ -192,18 +194,35 @@ def _parse_fields(
...
             # mode REPEATED means ARRAY
-            field_type = (
+            native_type = (
                 f"ARRAY<{field.field_type}>"
                 if field.mode == "REPEATED"
                 else field.field_type
             )
 
+            # See https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#TableFieldSchema.FIELDS.type
+            field_type = (
```
mars-lan, line 204: Let's also add a test for subfield in the context of bigquery extractor?

metaphor/common/fieldpath.py
alyiwang: two reasons, URL encoding doesn't encode `.` since it's not a resourced character, also, it encodes many other characters we don't need to escape

metaphor/bigquery/extractor.py
alyiwang: updated tests

tests/dbt/test_extractor.py
```diff
@@ -93,4 +95,5 @@ async def _test_project(
     extractor = DbtExtractor()
     events = [EventUtil.trim_event(e) for e in await extractor.extract(config)]
 
+    print(json.dumps(events))
```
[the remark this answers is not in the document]
alyiwang: done, thanks for catching

mars-lan (approved):

PR #337, opened 2022-09-01
Fix Power BI DAX parsing for Redshift sources

mars-lan:
### 🤔 Why?

The current parser implementation adds extra spaces for the Redshift database names, e.g. for the following DAX expression

```
let
    Source = AmazonRedshift.Database("url:5439", "db" )
    public = Source{[Name="public"]}[Data]
    table1 = public{[Name="table"]}[Data]
in
    table1'
```

the parser returns a dataset name ` db .public.test`

### 🤓 What?

Strip off the extra spaces.

### 🧪 Tested?

Tested against a production deployment.


elic-eon (approved):
Thanks for the fix.

PR #339, opened 2022-09-05
Add support for PostgreSQL & Redshift view definitions

mars-lan:
### 🤔 Why?

To capture the DDL used to create a view.

### 🤓 What?

Fetch the view definition from `pg_catalog.views` table.

### 🧪 Tested?

Tested locally against a production deployment.


elic-eon (approved):
LGTM

PR #343, opened 2022-09-08
Map Power BI report pages to charts

mars-lan:
### 🤔 Why?

Power BI's [REST API](https://docs.microsoft.com/en-us/rest/api/power-bi/) doesn't include any visualization information (charts) for [reports](https://docs.microsoft.com/en-us/power-bi/fundamentals/service-basic-concepts#reports). However, the name of the pages (tabs) in the report is considered a good proxy for the visualizations and should therefore be mapped to charts.

### 🤓 What?

As titled. Use the [Get Pages in Group](https://docs.microsoft.com/en-us/rest/api/power-bi/reports/get-pages-in-group) API to retrieve the information, which requires additional permissions. 

### 🧪 Tested?

Tested against a production deployment.

### DIFF
--- a/metaphor/power_bi/README.md
+++ b/metaphor/power_bi/README.md
@@ -13,6 +13,7 @@ We recommend creating a dedicated Azure AD Application and a dedicated security 
 2. Follow [this doc](https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/active-directory-groups-create-azure-portal#create-a-basic-group-and-add-members) to create a security group and add the app's service principal as a member.
 
 3. Log into [Power BI Admin Portal](https://app.powerbi.com/admin-portal/tenantSettings) as a Power BI admin, enable the following settings under **Admin API settings** for the security group created in the previous step:
+    - Allow service principals to use Power BI APIs
     - Allow service principals to use read-only Power BI admin APIs
     - Enhance admin APIs responses with detailed metadata
     - Enhance admin APIs responses with DAX and mashup expressions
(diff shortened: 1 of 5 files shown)


metaphor/power_bi/extractor.py
```diff
@@ -1,5 +1,6 @@
 import json
 import tempfile
+from multiprocessing import AuthenticationError
```
elic-eon, line 3: Does this error class show the useful error message?

metaphor/power_bi/extractor.py
```diff
@@ -152,6 +159,21 @@ class WorkspaceInfo(BaseModel):
     dashboards: List[WorkspaceInfoDashboard] = []
 
 
+class NotAuthorizedError(Exception):
```
elic-eon, line 162: Do we use this class?

metaphor/power_bi/README.md
```diff
@@ -21,6 +22,8 @@ For example,
 
 ![](https://docs.microsoft.com/en-us/power-bi/enterprise/media/read-only-apis-service-principal-auth/allow-service-principals-tenant-setting.png)
 
+4. [Add the service principal](https://docs.microsoft.com/en-us/power-bi/developer/embedded/embed-service-principal#step-4---add-the-service-principal-to-your-workspace) to all workspaces of interest, except "My Workspaces". Note that it may take an hour before the permission is fully propagated.
```
elic-eon, line 25: I think there is no other way to get rid of adding the service principal to all workspaces.

elic-eon (commented):
Otherwise, LGTM

metaphor/power_bi/extractor.py
mars-lan: Wrong class :(

metaphor/power_bi/README.md
mars-lan: The read-only Admin APIs don't require this. Unfortunately, they don't return the page information, so we had to fallback to the non-admin Power BI APIs.

PR #344, opened 2022-09-09
Pass connector configs directly into contructors

mars-lan:
### 🤔 Why?

When the connectors were first written, the idea was to pass the config file to the `extract` function so the same class can be invoked multiple times with different configs. In reality, this is never the case, since whenever the CLI is invoked, the connector classes are created and subsequently destroyed without being reused.

By passing the config directly into the constructor, the initialization process can be simplified with improved type checking. It also allows some configs to be stored as instance variables and used across methods without needing to pass them around.

### 🤓 What?

- Pass the config directly into each connector's constructor, instead of the `extract` method.
- Replace abstract `platform` and `description` methods with arguments passed to the `BaseExtractor`'s constructor.
- Replace `config_class` method with `from_config_file` method for parsing connector-specific config file.
- Rename `metaphor/common/extractor.py` to `metaphor/common/base_extractor.py` for consistency.
- Rename `ThoughtspotExtractor` to `ThoughtSpotExtractor`.
- Move the backported `AsyncMock` to `tests/test_utils.py` so it can be shared across tests.

### 🧪 Tested?

Manually ran every connector and compared the output against the output prior to the refactoring. Verified that there's only timestamp and other expected differences (e.g. additional queries & usage counts generated from the previous run).

### DIFF
--- a/metaphor/snowflake/extractor.py
+++ b/metaphor/snowflake/extractor.py
@@ -46,7 +46,7 @@ class SnowflakeExtractor(BaseExtractor):
         self._account = config.account
         self._filter = config.filter.normalize()
 
-        self._conn = connect(config)
+        self._conn = auth.connect(config)
         self._datasets: Dict[str, Dataset] = {}
 
     async def extract(self) -> Collection[ENTITY_TYPES]:
--- a/metaphor/snowflake/extractor.py
+++ b/metaphor/snowflake/extractor.py
@@ -1,25 +1,19 @@
 from datetime import datetime, timezone
 from typing import Collection, Dict, List, Mapping, Optional
 
-from snowflake.connector import SnowflakeConnection
-from snowflake.connector.cursor import DictCursor, SnowflakeCursor
-
-from metaphor.common.entity_id import dataset_fullname
-from metaphor.common.event_util import ENTITY_TYPES
-from metaphor.common.filter import DatasetFilter
-from metaphor.common.logger import get_logger
-from metaphor.models.crawler_run_metadata import Platform
-from metaphor.snowflake.auth import connect
-from metaphor.snowflake.config import SnowflakeRunConfig
-from metaphor.snowflake.utils import DatasetInfo, SnowflakeTableType
-
 try:
-    import snowflake.connector
+    from snowflake.connector.cursor import DictCursor, SnowflakeCursor
+    from snowflake.connector.errors import ProgrammingError
 except ImportError:
     print("Please install metaphor[snowflake] extra\n")
     raise
 
-from metaphor.common.extractor import BaseExtractor
+
+from metaphor.common.base_extractor import BaseExtractor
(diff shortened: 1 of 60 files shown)


metaphor/snowflake/lineage/extractor.py
```diff
@@ -41,50 +41,41 @@
 class SnowflakeLineageExtractor(BaseExtractor):
```
elic-eon, line 41: Would this connector inherit `SnowflakeExtractor` to make life easier?

metaphor/snowflake/extractor.py
```diff
@@ -33,52 +27,45 @@
     SourceInfo,
     SQLSchema,
 )
+from metaphor.snowflake.auth import connect
```
elic-eon, line 30:
```suggestion
from metaphor.snowflake import auth
```
Align with other snowflake connectors?

metaphor/snowflake/extractor.py
```diff
@@ -33,52 +27,45 @@
...
+
+    def __init__(self, config: SnowflakeRunConfig):
+        super().__init__(config, "Snowflake metadata crawler", Platform.SNOWFLAKE)
+        self._account = config.account
+        self._filter = config.filter.normalize()
 
-    def __init__(self):
-        self.account = None
-        self.max_concurrency = None
+        self._conn = connect(config)
```
elic-eon, line 49:
```suggestion
        self._conn = auth.connect(config)
```

tests/redshift/lineage/test_extractor.py
```diff
@@ -8,13 +8,7 @@
...
 from metaphor.redshift.lineage.config import RedshiftLineageRunConfig
 from metaphor.redshift.lineage.extractor import RedshiftLineageExtractor
-from tests.test_utils import load_json
-
-
-# python3.7 do not have AsyncMock
-class AsyncMock(MagicMock):
-    async def __call__(self, *args, **kwargs):
-        return super(AsyncMock, self).__call__(*args, **kwargs)
+from tests.test_utils import AsyncMock, load_json
```
elic-eon, line 11: 👍

elic-eon (commented):
Otherwise, LGTM

metaphor/snowflake/lineage/extractor.py
mars-lan: We plan to combine all Snowflake connectors into one  so this will be less relevant soon.

metaphor/snowflake/extractor.py
mars-lan: Good point. The reason why I import `auth` instead of `connect` is so that it can be monkey-patched in tests. We didn't have any tests for `SnowflakeExtractor` directly so it didn't matter.

elic-eon (approved):
Thanks for refactoring

PR #345, opened 2022-09-10
Add include_self_lineage config options to Snowflake & Redshift lineage connectors

mars-lan:
### 🤔 Why?

To allow self-referencing lineage to be filtered out.

### 🤓 What?

As titled.

### 🧪 Tested?

Unit tests


alyiwang (approved):
LGTM

PR #346, opened 2022-09-11
Extract last refreshed time for Power BI datasets

mars-lan:
### 🤔 Why?

Because it's a critical trust signal.

### 🤓 What?

As titled. Also use the opportunity to move `PowerBIClient` & related classes to a separate file.

### 🧪 Tested?

Verified against a production deployment.


alyiwang (approved):
LGTM
Thanks for refactoring

PR #347, opened 2022-09-11
Always exclude snowflake database & information_schema

mars-lan:
### 🤔 Why?

There's no good reason why the connectors should include the metadata DB & schemas.

### 🤓 What?

As titled. 

### 🧪 Tested?

Unit tests


metaphor/common/filter.py
```diff
@@ -16,6 +17,24 @@ class DatasetFilter:
...
 
+    def extend(self, filter: "DatasetFilter") -> "DatasetFilter":
+        extended = deepcopy(self)
+
+        if filter.includes is not None:
+            extended.includes = (
+                filter.includes
+                if extended.includes is None
+                else {**extended.includes, **filter.includes}
+            )
```
alyiwang, line 28: Can we construct a new object instead of creating a copy and mutating itself? Something like 
```
return DatasetFilter(
  includeds={**(self.includes or { includes: {}}).includes, **(filter.includes or { includes: {}}).includes }
),
  excludes=...
```

I wish python has cascading null operation like `?` in typescript.

alyiwang (commented):
Otherwise, LGTM

metaphor/common/filter.py
mars-lan: Can probably write this in a single line but would greatly reduce the readability. Tried a different approach and add more tests.

alyiwang (approved):
LGTM

elic-eon (approved):

PR #348, opened 2022-09-12
Sql-based Redshift Lineage crawler

elic-eon:
### 🤔 Why?

Extract lineage from `STL_SCAN` table didn't support lineage from Redshift spectrum table. For some use cases, we cannot extract lineage from `INSERT INTO <table> SELECT ... FROM <table2>`. The sql-bases lineage also fix this issue.

### 🤓 What?

- Provide lineage from `CREATE TABLE` and `INSERT INTO` like query
- Change the default method to sql-based

### 🧪 Tested?

Create example in dev environment, and verified the MCE. Manually ingest to Metaphor can verified the lineage showed correctly.


metaphor/redshift/lineage/config.py
```diff
@@ -9,8 +9,11 @@ class RedshiftLineageRunConfig(RedshiftRunConfig):
     # Whether to enable parsing view query to find upstream of the view, default True
     enable_view_lineage: bool = True
 
-    # Whether to enable parsing stl_scan table to find table lineage information, default True
-    enable_lineage_from_stl_scan: bool = True
+    # Whether to enable parsing stl_scan table to find table lineage information, default False
+    enable_lineage_from_stl_scan: bool = False
```
mars-lan, line 13: Can probably remove this altogether given that we plan to support column-level lineage, which can only come from query log parsing?

elic-eon: Got it, I will remove `stl_scan` based code.

mars-lan:
Based on https://docs.aws.amazon.com/redshift/latest/dg/r_STL_QUERY.html it seems that you need super user permission to view all queries? Or is unrestricted syslog access sufficient?

metaphor/redshift/lineage/extractor.py
```diff
@@ -134,6 +105,59 @@ async def _fetch_view_upstream(self, conn: Connection, db: str) -> None:
         """
         await self._fetch_lineage(view_lineage_query, conn, db)
 
+    async def _fetch_lineage_sql_base(self, conn):
```
mars-lan, line 108:
```suggestion
    async def _fetch_lineage_from_stl_query(self, conn):
```

metaphor/redshift/lineage/extractor.py
```diff
@@ -134,6 +105,59 @@ async def _fetch_view_upstream(self, conn: Connection, db: str) -> None:
...
+            query = row["querytxt"]
+            database = row["database"]
+            parser = Parser(query.replace('"', ""))
+
+            tables = [format_table_name(table, database) for table in parser.tables]
+            target, *sources = tables
+
+            if len(sources) == 0:
+                if self._include_self_lineage:
+                    sources = [target]
```
mars-lan, line 147: You should only include `target` if it's actually part of the lineage.

elic-eon: Let me rewrite this part. It is confusing.

elic-eon: Some self lineage would not capture in this implementation.

metaphor/redshift/lineage/extractor.py
```diff
@@ -134,6 +105,62 @@ async def _fetch_view_upstream(self, conn: Connection, db: str) -> None:
...
+                AND aborted = 0
+        ) q
+        ORDER BY q.endtime DESC;
+        """
+        results = await conn.fetch(sql)
+
+        def format_table_name(table: str, database: str) -> str:
+            tokens = table.split(".")
+            if len(tokens) == 1:
+                return ".".join([database, "public", tokens[0]])
```
mars-lan, line 130: Is `public` always the default schema when not specified?

metaphor/redshift/lineage/extractor.py
```diff
@@ -134,6 +105,62 @@ async def _fetch_view_upstream(self, conn: Connection, db: str) -> None:
...
+        sql = """
+        SELECT DISTINCT
+            trim(q.querytxt) AS querytxt,
+            trim(q.database) AS database
+        FROM (
+            SELECT *
+            FROM pg_catalog.stl_query
+            WHERE userid > 1
+                AND (
+                    querytxt ILIKE 'create table %% as select %%'
```
mars-lan, line 118: Are we sure these are the only 2 possible ways to create tables?

metaphor/redshift/lineage/extractor.py
```diff
@@ -134,6 +105,62 @@ async def _fetch_view_upstream(self, conn: Connection, db: str) -> None:
...
+            database = row["database"]
+            parser = Parser(query.replace('"', ""))
+
+            tables = [format_table_name(table, database) for table in parser.tables]
+
+            target, *sources = tables
+
+            # Only one table in a Insert/Create Table query
+            if len(tables) == 1:
+                if self._include_self_lineage:
```
mars-lan, line 148: Still not sure this is the correct logic. Let's chat offline.

metaphor/redshift/lineage/extractor.py
elic-eon: Nope. User can change it, should we skip sql without specifying schema? It's not a reliable way to assume any schema.

mars-lan: Ideally we can disambiguate it by searching all schemas for the table name. For now let's skip it and add a TODO to track that.

metaphor/redshift/lineage/extractor.py
elic-eon: Used `metaphor-sqllineage` and solve the issue.

metaphor/redshift/lineage/extractor.py
elic-eon: Added the skip logic

metaphor/redshift/lineage/extractor.py
elic-eon: https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_AS.html
There are other way to create table, but `CREATE_TABLE_AS` is what we want to capture here, only `CREATE_TABLE_AS` can do create table and insert data from query in the same time.

metaphor/redshift/lineage/extractor.py
```diff
@@ -134,6 +106,64 @@ async def _fetch_view_upstream(self, conn: Connection, db: str) -> None:
...
+                logger.warn(f"Cannot extract lineage for the query: {query}")
+                continue
+
+            if any(
+                [
+                    table.schema.raw_name == Schema.unknown
+                    for table in set(parser.source_tables + parser.target_tables)
+                ]
+            ):
+                continue
```
mars-lan, line 147: Should also log a warning here?

metaphor/redshift/lineage/extractor.py
```diff
@@ -134,6 +106,64 @@ async def _fetch_view_upstream(self, conn: Connection, db: str) -> None:
...
+        """
+        results = await conn.fetch(sql)
+
+        def format_table_name(table: Table, database: str) -> str:
+            print(str(table))
+            return f"{database}.{str(table)}"
+
+        for row in results:
+            query = row["querytxt"]
+            database = row["database"]
```
mars-lan, line 134: From https://docs.aws.amazon.com/redshift/latest/dg/r_STL_QUERY.html:

> The name of the database the user was connected to when the query was issued.

So this is the default DB name, but not necessarily the target DB name. Should only use this when DB name is unknown from the parsing result?

metaphor/redshift/lineage/extractor.py
```diff
@@ -134,6 +106,64 @@ async def _fetch_view_upstream(self, conn: Connection, db: str) -> None:
...
+                    OR querytxt ILIKE 'insert %%'
+                )
+                AND aborted = 0
+        ) q
+        ORDER BY q.endtime DESC;
+        """
+        results = await conn.fetch(sql)
+
+        def format_table_name(table: Table, database: str) -> str:
+            print(str(table))
```
mars-lan, line 129: Remove?

metaphor/redshift/lineage/extractor.py
mars-lan: Actually on second thought, is it possible to create tables outside of the connection's DB for Postgres/Redshit?

elic-eon: I think it would be the correct database name, I don't think user can do anything without a DB connection.

metaphor/redshift/lineage/extractor.py
elic-eon: Removed

metaphor/redshift/lineage/extractor.py
elic-eon: Added

metaphor/redshift/lineage/extractor.py
mars-lan: The question is about cross-DB statements, i.e. can you connect to DB1 and create a table in DB2? My guess is no for Postgres (yes for other systems), but just want to confirm.

elic-eon: Yes. `ERROR:  cross-database reference to database "metaphor" is not supported`. It will show this error.

metaphor/redshift/lineage/extractor.py
```diff
@@ -134,6 +106,65 @@ async def _fetch_view_upstream(self, conn: Connection, db: str) -> None:
...
+                logger.warning(f"Cannot extract lineage for the query: {query}")
+                continue
+
+            if any(
+                [
+                    table.schema.raw_name == Schema.unknown
+                    for table in set(parser.source_tables + parser.target_tables)
+                ]
+            ):
+                # TODO: find the default schema name
```
mars-lan, line 146: Add ticket number (SC-XXXX) as well.

metaphor/redshift/lineage/extractor.py
```diff
@@ -134,6 +106,65 @@ async def _fetch_view_upstream(self, conn: Connection, db: str) -> None:
...
+                logger.warning(f"Skip query with default schema: {query}")
```
mars-lan, line 147:
```suggestion
                logger.warning(f"Skip query missing explicit schema: {query}")
```

metaphor/redshift/lineage/extractor.py
mars-lan: Also any sense how common this is in the real world?

elic-eon: Created a [tickect](https://app.shortcut.com/metaphor-data/story/12243/extract-implicit-schema-name-from-the-query)

mars-lan (approved):
Great work.

PR #353, opened 2022-09-18
Extract tags for ThoughtSpot Liveboards & Answers

mars-lan:
### 🤔 Why?

The tags can be a useful metadata to extract.

### 🤓 What?

As titled. Also added extra debug logs.

### 🧪 Tested?

Manually verified against a production instance.


alyiwang (approved):
LGTM

PR #363, opened 2022-09-30
BigQuery query log crawler [sc-12584]

alyiwang:
### 🤔 Why?

fetch audit logs from big query for server side processing

### 🤓 What?

- fetch audit logs from big query

### 🧪 Tested?

tested locally against BIgQuery instance.

### DIFF
--- a/metaphor/bigquery/README.md
+++ b/metaphor/bigquery/README.md
@@ -101,6 +101,27 @@ The max number of concurrent requests to the google cloud API can be configured 
 max_concurrency: <max_number_of_queries> # Default to 5
 ```
 
+#### Query Logs
+
+By default, the BigQuery connector will fetch a full day's query logs (AuditMetadata) from yesterday, to be analyzed for additional metadata, such as dataset usage and lineage information. To backfill log data, one can set `lookback_days` to the desired value. To turn off query log fetching, set `lookback_days` to 0.  
+
+```yaml
+query_log:
+  # (Optional) Number of days of query logs to fetch. Default to 1. If 0, the no query logs will be fetched.
+  lookback_days: <days>
+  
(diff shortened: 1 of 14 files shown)


metaphor/bigquery/README.md
```diff
@@ -101,6 +101,27 @@ The max number of concurrent requests to the google cloud API can be configured
...
+  # (Optional) Number of days of query logs to fetch. Default to 1. If 0, the no query logs will be fetched.
+  lookback_days: <days>
+  
+  # (Optional) A list of users whose queries will be excluded from the log fetching.
+  excluded_usernames:
+    - <user_name1>
+    - <user_name2>
+
+  # (Optional) Exclude queries issued by service accounts. Default to true.
+  exclude_service_accounts: <boolean>
```
mars-lan, line 119: Probably want to set this to false by default

metaphor/bigquery/README.md
```diff
@@ -101,6 +101,27 @@ The max number of concurrent requests to the google cloud API can be configured
...
+
+  # (Optional) The number of query logs to fetch from BigQuery in one batch. Default to 1000.
```
mars-lan, line 121: I believe max is also 1000?

alyiwang: I didn't find documentation on the max page size. But 1000 seems a reasonable value that works in other bq crawlers.

metaphor/bigquery/README.md
alyiwang: updated

metaphor/bigquery/README.md
mars-lan: There is. Need to look in the rest API doc

mars-lan (approved):

PR #365, opened 2022-09-30
Improve PBI DAX parser for native queries

elic-eon:
### 🤔 Why?

We cannot parse the source table from PBI DAX in some cases. The reason is we didn't handle the control character before parsing the SQL.

### 🤓 What?

- Provide a optional snowflake account in config
- Unescape control char correctly
- Add tests to cover.

### 🧪 Tested?

Tests.

Manually verified the MCEs

### DIFF
--- a/metaphor/power_bi/config.py
+++ b/metaphor/power_bi/config.py
@@ -20,5 +20,5 @@ class PowerBIRunConfig(BaseConfig):
     # (Optional) The ids of Power BI workspace
     workspaces: List[str] = field(default_factory=lambda: list())
 
-    # (Optional) The snowflake account
+    # (Optional) The default snowflake account
     snowflake_account: Optional[str] = None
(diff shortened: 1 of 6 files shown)


metaphor/power_bi/config.py
```diff
@@ -19,3 +19,6 @@ class PowerBIRunConfig(BaseConfig):
 
     # (Optional) The ids of Power BI workspace
     workspaces: List[str] = field(default_factory=lambda: list())
+
+    # (Optional) The snowflake account
```
mars-lan, line 23:
```suggestion
    # (Optional) The default snowflake account
```

mars-lan (commented):
Please also update README

mars-lan (commented):
Otherwise LGTM

mars-lan (approved):

PR #371, opened 2022-10-06
Populate query ID & sql hash fields

mars-lan:
### 🤔 Why?

These fields are useful to capture to help identify and dedupe the queries.

### 🤓 What?

As titled.

### 🧪 Tested?

Manually tested against production instances.


theseyi (dismissed):
tenor-19230247

theseyi (approved):
tenor-19230247

PR #378, opened 2022-10-11
Extract thoughtspot pinboard embed url

elic-eon:
### 🤔 Why?

To embed dashboard from Metaphor.

### 🤓 What?

- Fill out the embed URL field for TS dashboard

### 🧪 Tested?

Verified the embed urls from the MCE is working.


tests/thought_spot/test_extractor.py
```diff
@@ -53,4 +53,8 @@ def create_mock_client():
         extractor = ThoughtSpotExtractor(dummy_config())
         events = [EventUtil.trim_event(e) for e in await extractor.extract()]
 
+        import json
```
mars-lan, line 56: Drop?

elic-eon: Forgot about it, thanks.

mars-lan (approved):

PR #380, opened 2022-10-11
Statement exhausted compilation resources [sc-13029]

alyiwang:
### 🤔 Why?

For a snowflake database with a large number of tables, the query to get table info including DDL can get exceedingly long. 

### 🤓 What?

- Cap the number of tables info to fetch in one query to 100

### 🧪 Tested?

tested against snowflake instance, and verified the output is the same as before.

### DIFF
--- a/metaphor/snowflake/extractor.py
+++ b/metaphor/snowflake/extractor.py
@@ -4,6 +4,7 @@ from typing import Collection, Dict, List, Mapping, Optional, Tuple
 
 from pydantic import parse_raw_as
 
+from metaphor.common import utils
 from metaphor.common.filter import DatabaseFilter, DatasetFilter
 from metaphor.common.query_history import chunk_query_logs
 from metaphor.common.utils import md5_digest, start_of_day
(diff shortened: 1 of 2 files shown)


mars-lan (commented):
Bump version?

metaphor/snowflake/extractor.py
```diff
@@ -213,10 +217,19 @@ def _fetch_columns(self, cursor: SnowflakeCursor, database: str) -> None:
...
 
-    def _fetch_table_info(self, tables: Dict[str, DatasetInfo], shared: bool) -> None:
+    def _fetch_table_information(
+        self, tables: Dict[str, DatasetInfo], shared: bool
+    ) -> None:
+        chunks = utils.chunks([item for item in tables.items()], TABLE_INFO_FETCH_SIZE)
+        for chunk in chunks:
+            self._fetch_table_info(chunk, shared)
+
+    def _fetch_table_info(
```
mars-lan, line 227: Consider rename this to `_fetch_table_info_internal` or make it an inner function?

alyiwang: updated

mars-lan: My bad. Given how long the function is, it's not so readable after making it an inner function. Maybe just rename it?

mars-lan (dismissed):

metaphor/snowflake/extractor.py
alyiwang: reverted.

mars-lan (approved):

PR #382, opened 2022-10-12
fix query log _id

alyiwang:
### 🤔 Why?

The `_id` currentlly looks like `DataPlatform.BIGQUERY:xxx` where the `DataPlatform` is useless and should be removed 

### 🤓 What?

- use Enum.name instead of `str` to stringify enum

### 🧪 Tested?

verified in local run

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.11.53"
+version = "0.11.55"
 license = "Apache-2.0"
 description = "A collection of Python-based 'connectors' that extract metadata from various sources to ingest into the Metaphor app."
 authors = ["Metaphor <dev@metaphor.io>"]
--- a/metaphor/bigquery/extractor.py
+++ b/metaphor/bigquery/extractor.py
@@ -302,7 +302,7 @@ class BigQueryExtractor(BaseExtractor):
         )
 
         return QueryLog(
-            id=f"{str(DataPlatform.BIGQUERY)}:{job_change.job_name}",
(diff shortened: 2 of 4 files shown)


mars-lan:
> Will fix the app mongoDB findRecentQueries in the next PR

Why does it need any fixing?

alyiwang:
> > Will fix the app mongoDB findRecentQueries in the next PR
> 
> Why does it need any fixing?

https://github.com/MetaphorData/app/blob/e918dd0a549203e59fd1aa7ec9665ebb472f4939/%40common/src/db/mongodb.ts#L820
Restore original _id

mars-lan:
> > > Will fix the app mongoDB findRecentQueries in the next PR
> > 
> > 
> > Why does it need any fixing?
> 
> https://github.com/MetaphorData/app/blob/e918dd0a549203e59fd1aa7ec9665ebb472f4939/%40common/src/db/mongodb.ts#L820 Restore original _id

Argh I see. The restoration is not a hard requirement since no one uses that but got it.

mars-lan (approved):

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.11.54"
+version = "0.11.55"
```
mars-lan, line 3: Not needed?

alyiwang: There are meaningful code changes, a version bump is needed?

mars-lan: This is only for the generated model, which had no change in this PR?

mars-lan: Oh never mind. I read it wrong.

PR #384, opened 2022-10-13
Merge Reshift query crawler into base crawler [sc-13085]

alyiwang:
### 🤔 Why?

To get access logs in the Redshift crawler for server side log processing. Will deprecate the query crawler next.

### 🤓 What?

- Fetch access log in Redshift crawler

### 🧪 Tested?

tested against Redshift instance, verified the result


mars-lan (approved):
Thanks for adding this quickly.

PR #387, opened 2022-10-14
Improve Redshift lineage

elic-eon:
### 🤔 Why?

`stl_query` didn't contain the full query. The sql parse may throw parsing error.

### 🤓 What?

- Join [SQL_QUERYTEXT](https://docs.aws.amazon.com/redshift/latest/dg/r_STL_QUERYTEXT.html) for full query text.
- Catch the exception and log it.

### 🧪 Tested?

Run crawler against dev and prod environment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.11.59"
+version = "0.11.60"
 license = "Apache-2.0"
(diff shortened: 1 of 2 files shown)


mars-lan (dismissed):

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.11.58"
+version = "0.11.59"
```
mars-lan, line 3: Need to bump again after merging main.

mars-lan (approved):

PR #394, opened 2022-10-24
Fill dataset structure from crawler

elic-eon:
### 🤔 Why?

Fill out the three level `DatasetStructure` to enable the next step of showing case sensitive name. Potentially we can support name with dots.

### 🤓 What?

- Fill out the field for `Bigquery`, `Snowflake`, `Redshift`, `PostgreSQL`, and `Glue`.
- Use `dataset_fullname` to fill out the logical id as possible

### 🧪 Tested?

- Unit tests.
- Run `Snowflake`, `Bigquery`, and `Redshift` crawler on local, ingest the MCEs, and verified the result on Metaphor.

PR #395, opened 2022-10-25
bigquery dataset name to lowercase

alyiwang:
### 🤔 Why?

BQ dataset logical name didn't use normalized name, causing customer data issue.

### 🤓 What?

- normalize BQ dataset name
- didn't change the name used in `structure` and kept the original case there

### 🧪 Tested?

unit tests

### DIFF
--- a/metaphor/bigquery/extractor.py
+++ b/metaphor/bigquery/extractor.py
@@ -12,6 +12,7 @@ from typing import (
 )
 
 from metaphor.bigquery.logEvent import JobChangeEvent
+from metaphor.common.entity_id import dataset_normalized_name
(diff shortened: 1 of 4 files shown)


mars-lan (commented):
This is breaking? It'll change the ID of existing datasets?

alyiwang:
> This is breaking? It'll change the ID of existing datasets?

Actually, just found out the BQ actually support case-sensitive table names https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#case_sensitivity 
Dropping this PR.

elic-eon:
Bigquery profile crawler use normalized name in #394 , should I revert it?

PR #397, opened 2022-10-26
Remove dependency on Snowflake Enterprise [sc-13398]

alyiwang:
### 🤔 Why?

Non-enterprise snowflake accounts won't have access to the `access_history` view to get access logs.

### 🤓 What?

- check if the `access_history` view is available, if not, only fetch the raw query text from `query_history`

### 🧪 Tested?

tested against snowflake instance. Tested with the `has_access_history` set to both true and false.

### DIFF
--- a/metaphor/snowflake/extractor.py
+++ b/metaphor/snowflake/extractor.py
@@ -45,6 +45,7 @@ from metaphor.snowflake.utils import (
     QueryWithParam,
     SnowflakeTableType,
     async_execute,
+    check_access_history,
     exclude_username_clause,
     fetch_query_history_count,
 )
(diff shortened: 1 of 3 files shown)


mars-lan (commented):
I don't think you need to check this specifically. Non-enterprise still has ACCESS_HISTORY but it'll just return empty result.

alyiwang:
> I don't think you need to check this specifically. Non-enterprise still has ACCESS_HISTORY but it'll just return empty result.

Exactly, I am checking if there is any available data in the access_history, and will base the following query on the flag.

metaphor/snowflake/extractor.py
```diff
@@ -384,29 +393,53 @@ def _fetch_query_logs(self) -> None:
...
+                        end_date,
+                        start_date,
+                        end_date,
+                        *self._query_log_excluded_usernames,
+                        x * self._query_log_fetch_size,
+                    ),
+                )
+                for x in range(batches)
+            }
+            if has_access_history
```
mars-lan, line 407: Instead of trying to span a ternary operator over 47 lines, let's move the query generation into 2 methods?

mars-lan:
> > I don't think you need to check this specifically. Non-enterprise still has ACCESS_HISTORY but it'll just return empty result.
> 
> Exactly, I am checking if there is any available data in the access_history, and will base the following query on the flag.

Argh nvm. I see what you're doing now.

metaphor/snowflake/extractor.py
mars-lan: Actually in the end state, we don't even need to query `access_history` even for the Enterprise edition, since the lineage will be processed regardless?

alyiwang: There are two reasons to still join access history when possible. 1 source of truth for dataset lineage, parser can't handle all cases. 2 automatically filter out queries that doesn't access any table. In our snowflake instance, it's actually majority of the cases.

mars-lan: SGTM. We'll need to be explicit about empty array vs null (which you might be doing already) so the parser knows how to distinguish entries that didn't access any tables vs lacking `access_history`.

alyiwang: Yes, we used empty `access_history` to indicate it's not available, then only fetch `query_history` (which contains statement without data accessing). Otherwise, two tables are joined, and the result contains only queries that has data access.

mars-lan: Resolved this offline.

alyiwang: updated, thanks

metaphor/snowflake/extractor.py
```diff
@@ -363,13 +364,39 @@ def _fetch_query_logs(self) -> None:
         start_date = start_of_day(self._query_log_lookback_days)
         end_date = start_of_day()
 
+        has_access_history = check_access_history(self._conn)
+        logger.info(f"has access history: {has_access_history}")
```
mars-lan, line 368:
```suggestion
        logger.info(f"Using Snowflake Enterprise edition: {has_access_history}")
```

mars-lan (commented):
Otherwise LGTM!

mars-lan (approved):

PR #401, opened 2022-10-30
Ignore malformed Looker dashboards

mars-lan:
### 🤔 Why?

It's possible to create malformed dashboards in Looker which causes the Looker SDK to throw errors when deserializing returned the API content.

### 🤓 What?

Catch and ignore SDK errors when encountering malformed dashboards. Also changed to the shared logger in `lookml_parser.py`.

### 🧪 Tested?

Manually tested against a production instance containing malformed dashboards.


alyiwang (approved):
LGTM

PR #402, opened 2022-10-31
Merge app non app version of pbi dashboards

elic-eon:
### 🤔 Why?

No reason to keep both app and non-app version of dashboards

### 🤓 What?

- Merge app information into non-app version
- Use `poetry` 1.2 to generate lock file. `poetry lock --no-update`

### 🧪 Tested?

Unit tests. Compared before after, we have 8 assets with app enabled in our PowerBI, verified that we reduce 8 asset after the change.

### DIFF
--- a/metaphor/power_bi/extractor.py
+++ b/metaphor/power_bi/extractor.py
@@ -341,13 +341,7 @@ class PowerBIExtractor(BaseExtractor):
 
     def _get_dashboard_id_from_url(self, url: str) -> Optional[str]:
         path = parse.urlparse(url).path
-        print(path)
-        pattern = re.compile(
-            r"apps/"
-            r"([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})/"
-            r"(reports|dashboards)/"
-            r"([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})"
-        )
+        pattern = re.compile(r"apps/([^/]+)/(reports|dashboards)/([^/]+)")
         match = pattern.search(path)
         if match and len(match.groups()) == 3:
             return match.group(3)
--- a/metaphor/power_bi/extractor.py
+++ b/metaphor/power_bi/extractor.py
@@ -290,6 +290,9 @@ class PowerBIExtractor(BaseExtractor):
             original_dashboard = self._dashboards.get(original_dashboard_id)
             if original_dashboard is None:
                 # Cannot not found corresponding non-app dashboard
+                logger.warning(
+                    f"Non app version dashboard not found, id: {dashboard_id}"
+                )
                 continue
 
             original_dashboard.dashboard_info.power_bi.app = (
--- a/metaphor/power_bi/extractor.py
+++ b/metaphor/power_bi/extractor.py
@@ -277,7 +277,7 @@ class PowerBIExtractor(BaseExtractor):
             )
             self._dashboards[wi_dashboard.id] = dashboard
 
-    def merge_app_version_into_original_version(self):
+    def dedupe_app_version_dashboards(self):
         for dashboard_id in list(self._dashboards.keys()):
             dashboard = self._dashboards.get(dashboard_id)
             if dashboard.dashboard_info.power_bi.app is None:
--- a/metaphor/power_bi/extractor.py
+++ b/metaphor/power_bi/extractor.py
@@ -280,15 +280,22 @@ class PowerBIExtractor(BaseExtractor):
     def merge_app_version_into_original_version(self):
         for dashboard_id in list(self._dashboards.keys()):
             dashboard = self._dashboards.get(dashboard_id)
-            if dashboard.dashboard_info.power_bi.app is not None:
-                original_id = self._get_original_id_from_url(
-                    dashboard.source_info.main_url
-                )
-                original_dashboard = self._dashboards.get(original_id)
-                original_dashboard.dashboard_info.power_bi.app = (
-                    dashboard.dashboard_info.power_bi.app
-                )
-                del self._dashboards[dashboard_id]
+            if dashboard.dashboard_info.power_bi.app is None:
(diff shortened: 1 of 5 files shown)


metaphor/power_bi/extractor.py
```diff
@@ -313,6 +327,9 @@ def _find_last_completed_refresh(
         iso_time = ".".join(refresh.endTime.split(".")[0:-1])
         return datetime.fromisoformat(iso_time).replace(tzinfo=timezone.utc)
 
+    def _get_original_id_from_url(self, url: str) -> str:
```
mars-lan, line 330:
```suggestion
    def _get_dashboard_id_from_url(self, url: str) -> str:
```

metaphor/power_bi/extractor.py
```diff
@@ -276,6 +277,19 @@ def map_wi_dashboards_to_dashboard(
             )
             self._dashboards[wi_dashboard.id] = dashboard
 
+    def merge_app_version_into_original_version(self):
+        for dashboard_id in list(self._dashboards.keys()):
+            dashboard = self._dashboards.get(dashboard_id)
+            if dashboard.dashboard_info.power_bi.app:
```
mars-lan, line 283:
```suggestion
            if dashboard.dashboard_info.power_bi.app is not None:
```

metaphor/power_bi/extractor.py
```diff
@@ -276,6 +277,19 @@ def map_wi_dashboards_to_dashboard(
...
+                original_id = self._get_original_id_from_url(
+                    dashboard.source_info.main_url
+                )
+                original_dashboard = self._dashboards.get(original_id)
```
mars-lan, line 287: Might want to defensively check that `original_dashboard` is not None.

elic-eon: Sounds good.

elic-eon: Will keep app version if non-app version is not exist (shouldn't be possible)

metaphor/power_bi/extractor.py
```diff
@@ -276,6 +277,26 @@ def map_wi_dashboards_to_dashboard(
             )
             self._dashboards[wi_dashboard.id] = dashboard
 
+    def merge_app_version_into_original_version(self):
```
mars-lan, line 280:
```suggestion
    def dedupe_app_version_dashboards(self):
```

mars-lan: Also add a comment for posterity.

metaphor/power_bi/extractor.py
```diff
@@ -276,6 +277,26 @@ def map_wi_dashboards_to_dashboard(
...
+            if dashboard.dashboard_info.power_bi.app is None:
+                # Non-app dashboard
+                continue
+
+            original_dashboard_id = self._get_dashboard_id_from_url(
+                dashboard.source_info.main_url
+            )
+            original_dashboard = self._dashboards.get(original_dashboard_id)
+            if original_dashboard is None:
+                # Cannot not found corresponding non-app dashboard
```
mars-lan, line 292: `logger.warn`?

metaphor/power_bi/extractor.py
```diff
@@ -313,6 +337,9 @@ def _find_last_completed_refresh(
         iso_time = ".".join(refresh.endTime.split(".")[0:-1])
         return datetime.fromisoformat(iso_time).replace(tzinfo=timezone.utc)
 
+    def _get_dashboard_id_from_url(self, url: str) -> str:
+        return url.rstrip("/").split("/")[-1]
```
mars-lan, line 341: Let's use proper regex for this? Also you may wish to use [urllib.parse](https://docs.python.org/3/library/urllib.parse.html#module-urllib.parse) to extract the path portion.

metaphor/power_bi/extractor.py
```diff
@@ -276,6 +277,29 @@ def map_wi_dashboards_to_dashboard(
...
+                continue
+
+            original_dashboard_id = self._get_dashboard_id_from_url(
+                dashboard.source_info.main_url
+            )
+            original_dashboard = self._dashboards.get(original_dashboard_id)
+            if original_dashboard is None:
+                # Cannot not found corresponding non-app dashboard
+                logger.warning(
+                    f"Non app version dashboard not found, id: {dashboard_id}"
```
mars-lan, line 294:
```suggestion
                    f"Non-app version dashboard not found, id: {dashboard_id}"
```

metaphor/power_bi/extractor.py
```diff
@@ -337,8 +339,19 @@ def _find_last_completed_refresh(
...
         return datetime.fromisoformat(iso_time).replace(tzinfo=timezone.utc)
 
-    def _get_dashboard_id_from_url(self, url: str) -> str:
-        return url.rstrip("/").split("/")[-1]
+    def _get_dashboard_id_from_url(self, url: str) -> Optional[str]:
+        path = parse.urlparse(url).path
+        print(path)
+        pattern = re.compile(
+            r"apps/"
+            r"([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})/"
```
mars-lan, line 347: This might be too specific to be robust over the long term. Let's use something like `[^/]+`?

elic-eon: Sounds good

elic-eon: Updated.

metaphor/power_bi/extractor.py
```diff
@@ -276,6 +279,29 @@ def map_wi_dashboards_to_dashboard(
...
+            )
+            original_dashboard = self._dashboards.get(original_dashboard_id)
+            if original_dashboard is None:
+                # Cannot not found corresponding non-app dashboard
+                logger.warning(
+                    f"Non-app version dashboard not found, id: {original_dashboard_id}"
+                )
+                continue
+
+            original_dashboard.dashboard_info.power_bi.app = (
```
mars-lan, line 300: Let's also use the app version URL for the original dashboard.

mars-lan (approved):

PR #403, opened 2022-10-31
fix customMetadata error

pk1230h:
### 🤔 Why?
The previous `CustomMetadata` format was wrong  and web app couldn't detect it. Therefore, the dataset cannot be found on web app.

### 🤓 What?
- fix CustomMetadataItem with `json.dumps()`

### 🧪 Tested?
before the fix, we could not see the data.

after fixed 
Screenshot 2022-10-31 at 12 40 15 PM


mars-lan (approved):

mars-lan (changes requested):
Bump up version

pk1230h:
bump to version `0.11.72`

mars-lan (approved):

PR #405, opened 2022-11-02
Downloaded log file name should encode date

elic-eon:
### 🤔 Why?

As title.

### 🤓 What?

Including timestamp in the output log filename.

### 🧪 Tested?

Verified on a local crawler run.


alyiwang (approved):
LGTM

PR #406, opened 2022-11-02
populate DBT metric URL [sc-13231]

alyiwang:
### 🤔 Why?

To be able to show direct link to the dbt metric docs page from metaphor UI

### 🤓 What?

- generate metric URL from base URL and metric id
- fix dbt cloud URL, it uses `job` instead of `run` now.

### 🧪 Tested?

Tested the dbt cloud crawler locally, and verified the URLs.


mars-lan (commented):
Is this a dbt cloud-only feature? Or is it a generic dbt doc URL?

alyiwang:
> Is this a dbt cloud-only feature? Or is it a generic dbt doc URL?

Should serve both. For dbt cloud, the base URL can be derived. For on-prem dbt docs server, the base URL needs to be provided in config

mars-lan (approved):

PR #408, opened 2022-11-06
Add runner wrapper and add instructions on custom integration

mars-lan:
### 🤔 Why?

Currently for third parties to author custom connectors, they'll need to deal with the complexity of event building and file writing. This PR aims to simplify that significantly by hiding all that behind a `run_connector` function.

Note that while it's possible to benefit the same by extending `BaseExtractor` directly, this would require external vendors to follow the config file requirements, which is often non-ideal.

### 🤓 What?

- Move events building and output writing logic to a `run_connector` function so it can be leveraged by both the third paries as well as `BaseExtractor`
- Repurposed the original model document into "how to add a custom connector"  

### 🧪 Tested?

Manually ran the existing connectors to make sure the refactoring continued to work as expected.


docs/custom.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Outputting Events
+
+To output the entities to a local path (for development) or an S3 bucket (for production) you can use the [`run_connector`](../metaphor/common/runner.py) wrapper function to execute your connector function:
+
+```py
+from metaphor.common.runner import run_connector
+
+
+run_connector(
```
alyiwang, line 99: Does it take a config file?

metaphor/common/runner.py
```diff
@@ -0,0 +1,129 @@
...
+    if file_sink_config is not None:
+        file_sink = FileSink(file_sink_config)
+        file_sink.sink(events)
+        file_sink.sink_metadata(run_metadata)
+        file_sink.sink_logs()
+
+    return events
+
+
+def metaphor_file_sink_config(
```
alyiwang, line 87: It's a little confusing that we have the `output` config and also its file_sink_config.

metaphor/common/base_extractor.py
```diff
@@ -35,56 +28,16 @@ def __init__(
...
-            f"Extractor ended with {run_status} at {end_time}, fetched {entity_count} entities, took {format((end_time - start_time).total_seconds(), '.1f')}s"
-        )
-
-        crawler_name = EventUtil.class_fqcn(self.__class__)
-        event_util = EventUtil(crawler_name)
-        events = [event_util.build_event(entity) for entity in entities]
 
-        run_metadata = CrawlerRunMetadata(
-            crawler_name=crawler_name,
+        return run_connector(
```
alyiwang, line 37: Still not 100% sure about this additional wrapper. I think a customer connector only needs to implement the `extract` function, and can use the `metaphor` command to test the connector. It's not much more work than this.

mars-lan: My original thought was to also extend `BaseExtractor`, but that would require 3rd party vendors to use a config file, which in many cases doesn't make sense—think batch running of Data Quality framework. This seems to offer the maximum flexibility while giving us control over the output format.

docs/custom.md
mars-lan: See https://github.com/MetaphorData/connectors/pull/408#discussion_r1014988449

metaphor/common/runner.py
mars-lan: Sorry but could you elaborate?

metaphor/common/base_extractor.py
```diff
@@ -35,56 +28,16 @@ def __init__(
...
-            end_time=end_time,
-            status=run_status,
-            entity_count=float(entity_count),
-            error_message=error_message,
-            stack_trace=stacktrace,
+            file_sink_config=self._output.file,
         )
-
-        if self._output.api is not None:
-            ApiSink(self._output.api).sink(events)
```
elic-eon, line 79: Didn't see `ApiSink` support in `run_connector`, are we going to drop this support for crawlers?

docs/custom.md
```diff
@@ -0,0 +1,151 @@
...
+      upstream = DatasetUpstream(source_datasets=[src1_id, src2_id]),
+  )
+
+  # Return a list of datasets
+  return [dataset]
+
+# Use the runner to run the connector and output events to the tenant's S3 bucket
+connector_name = "custom_lineage_connector"
+run_connector(
+  connector_func=custom_connector,
```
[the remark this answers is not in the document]
mars-lan: Thanks for catching!

docs/custom.md
```diff
@@ -0,0 +1,151 @@
...
+
+  # Create string-based entity IDs for src1 & src2
+  src1_id = str(to_dataset_entity_id('db.schema.src1', DataPlatform.BIGQUERY))
+  src2_id = str(to_dataset_entity_id('db.schema.src2', DataPlatform.BIGQUERY))
+
+  # Set the upstream aspect
+  dataset = Dataset(
+      logical_id = DatasetLogicalID(name='db.schema.dest', platform=DataPlatform.BIGQUERY),
+      upstream = DatasetUpstream(source_datasets=[src1_id, src2_id]),
+  )
```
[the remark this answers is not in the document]
mars-lan: Thanks. Been relying on black so much that I never realized that it's doing 4-space indentation.

metaphor/common/base_extractor.py
mars-lan: Good catch. Would like to deprecate it but shouldn't remove it completely. Let me add it back.

metaphor/common/base_extractor.py
alyiwang: Got it, thanks for explaining. Could you add this to the PR description, because `Abstract away the complexity in building events and writing output` can also be achieved by `BaseExtractor`?

metaphor/common/runner.py
alyiwang: I was thinking there are two ways to specify the output of a crawler, one in the config file, the other through this `metaphor_file_sink_config`, which can cause some confusion.

metaphor/common/base_extractor.py
mars-lan: Updated.

metaphor/common/runner.py
alyiwang: I guess `metaphor_file_sink_config` is meant to be used when calling `run_connector` method directly

mars-lan: That's correct. The `OutputConfig` class is only used in the config files for our own crawlers.

alyiwang (approved):
LGTM

PR #409, opened 2022-11-07
Add asset description to manual.governance connector

mars-lan:
### 🤔 Why?

To support manually assigning one or multiple descriptions to a table using the `manual.governance` connector.

### 🤓 What?

As titled.

### 🧪 Tested?

Manually ran the connector and verified the output.


alyiwang (commented):
Which aspect do we save this description to during ingestion? Do we create a new asset description?

alyiwang (approved):
lgtm

PR #411, opened 2022-11-10
Normalize BQ table names in lineage

elic-eon:
### 🤔 Why?

We lost the lineage because different treatment on BQ dataset name between BQ connector and dbt

### 🤓 What?

- Normalize the BQ name

### 🧪 Tested?

Add unit tests to cover

### DIFF
--- a/metaphor/bigquery/lineage/extractor.py
+++ b/metaphor/bigquery/lineage/extractor.py
@@ -120,7 +120,7 @@ class BigQueryLineageExtractor(BaseExtractor):
             dataset_ids.add(
                 str(
                     to_dataset_entity_id(
-                        dataset_normalized_name(table=dataset_name).replace("`", ""),
+                        dataset_normalized_name(table=dataset_name),
                         DataPlatform.BIGQUERY,
                         None,
                     )
--- a/metaphor/bigquery/lineage/extractor.py
+++ b/metaphor/bigquery/lineage/extractor.py
@@ -13,7 +13,7 @@ from metaphor.bigquery.lineage.config import BigQueryLineageRunConfig
 from metaphor.bigquery.logEvent import JobChangeEvent
 from metaphor.bigquery.utils import LogEntry, build_client, build_logging_client
 from metaphor.common.base_extractor import BaseExtractor
-from metaphor.common.entity_id import to_dataset_entity_id
+from metaphor.common.entity_id import dataset_normalized_name, to_dataset_entity_id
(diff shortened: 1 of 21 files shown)


mars-lan:
Why not normalize the lineage carwler instead?

metaphor/bigquery/lineage/extractor.py
```diff
@@ -118,7 +120,7 @@ def _parse_view_lineage(self, project_id, bq_table: bigquery.table.Table) -> Non
             dataset_ids.add(
                 str(
                     to_dataset_entity_id(
-                        dataset_name.replace("`", ""),
+                        dataset_name.replace("`", "").lower(),
```
mars-lan, line 123: User `dataset_normalized_name` directly?

elic-eon: You're right. Does the replace back quote logic need to move to `dataset_normalized_name`?

mars-lan: Yes, make senses to move the quote stripping into `dataset_normalized_name` as well.

elic-eon: Updated

metaphor/common/entity_id.py
```diff
@@ -78,7 +78,19 @@ def to_person_entity_id(email: str) -> EntityId:
 
 
 def dataset_normalized_name(
-    db: Optional[str] = None, schema: Optional[str] = None, table: Optional[str] = None
+    db: Optional[str] = None,
+    schema: Optional[str] = None,
+    table: Optional[str] = None,
+    lower_case: bool = True,
```
mars-lan, line 84: Would rather hard-code this to avoid any discrepancy when normalizing dataset names.

elic-eon: Make sense, updated.

metaphor/common/entity_id.py
```diff
@@ -78,7 +78,16 @@ def to_person_entity_id(email: str) -> EntityId:
 
 
 def dataset_normalized_name(
-    db: Optional[str] = None, schema: Optional[str] = None, table: Optional[str] = None
+    db: Optional[str] = None,
+    schema: Optional[str] = None,
+    table: Optional[str] = None,
+    strip_backquote: bool = True,
```
mars-lan, line 84: This as well

elic-eon: Removed.

mars-lan (approved):

PR #425, opened 2022-11-19
Stop filling in the event header in MCE

mars-lan:
### 🤔 Why?

Event headers are not used and are marked as optional before eventual removal.

### 🤓 What?

As titled.

### 🧪 Tested?

Unit tests


elic-eon (approved):
LGTM

PR #429, opened 2022-11-22
Replace custom checking code with Poetry's --skip-existing option when publishing to PyPI

mars-lan:
### 🤔 Why?

Poetry 1.2 added the new [--skip-existing](https://python-poetry.org/docs/cli/#:~:text=%2D%2Dskip-,%2D,-existing%3A%20Ignore%20errors) option to the `publish` command, which obviate the need of our custom checking code.

### 🤓 What?

As titled.

### 🧪 Tested?

https://github.com/MetaphorData/connectors/actions/runs/3525187389/jobs/5911594661

Screenshot 2022-11-22 at 8 24 50 AM

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -98,11 +98,9 @@ jobs:
 
       - name: Publish to PyPI
         run: >
-          poetry publish --build
+          poetry publish --build --skip-existing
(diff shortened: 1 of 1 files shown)


alyiwang (approved):
Nice find!

PR #431, opened 2022-11-24
Add support for environment variable substitution in config files

mars-lan:
### 🤔 Why?

This makes it easier to specify sensitive configs (e.g. password, secrets) using environment variables, especially in the context of https://github.com/MetaphorData/connectors-action, where the only possible way to pass GitHub secrets is via environment variables.

### 🤓 What?

As titled.

### 🧪 Tested?

Unit tests

### DIFF
--- a/metaphor/common/base_config.py
+++ b/metaphor/common/base_config.py
@@ -5,8 +5,9 @@ from pydantic import parse_obj_as
 from pydantic.dataclasses import dataclass
 from smart_open import open
 
-from .api_sink import ApiSinkConfig
-from .file_sink import FileSinkConfig
+from metaphor.common.api_sink import ApiSinkConfig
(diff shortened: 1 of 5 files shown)


metaphor/common/base_config.py
```diff
@@ -5,8 +5,9 @@
 from pydantic.dataclasses import dataclass
 from smart_open import open
 
-from .api_sink import ApiSinkConfig
-from .file_sink import FileSinkConfig
+from metaphor.common.api_sink import ApiSinkConfig
+from metaphor.common.file_sink import FileSinkConfig
```
elic-eon, line 9: 👍

elic-eon (approved):
LGTM!

elic-eon (approved):

PR #434, opened 2022-12-05
Fix the TS table field mappings

elic-eon:
### 🤔 Why?

The `physicalColumnGUID` is not available now, the lineage between "table" and external source table is broken

### 🤓 What?

- Use the connection to get the source entity id

### 🧪 Tested?

Update tests. Verified the MCEs before and after are identical.


mars-lan (commented):
Bump up version?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.11.93"
```
mars-lan, line 3: Bump up again.

mars-lan (commented):
Otherwise LGTM

pyproject.toml
elic-eon: Updated

mars-lan (approved):

PR #442, opened 2022-12-14
Support case sensitive Snowflake identifier [sc-10531]

alyiwang:
### 🤔 Why?

For case-sensitive snowflake table names, we need to put them in double quotes when formulating extra_info query. Otherwise, it's throwing invalid query exception.

### 🤓 What?

- use double quotes when formulating extra_info query

### 🧪 Tested?

tested against snowflake DB with a case-sensitive table name.


metaphor/snowflake/utils.py
```diff
@@ -37,6 +37,11 @@ class QueryWithParam:
     params: Optional[Tuple] = None
 
 
+def to_quoted_identifier(parts: List[Optional[str]]) -> str:
```
mars-lan, line 40: Test?

metaphor/common/entity_id.py
```diff
@@ -81,7 +81,7 @@ def normalize_full_dataset_name(name: str) -> str:
     """
     Normalizes a fully qualified dataset name
     """
-    return name.lower().replace("`", "")
+    return name.lower().replace("`", "").replace('"', "")
```
mars-lan, line 84: Consider adding a test.

mars-lan (commented):
Thanks for the quick fix.

metaphor/common/entity_id.py
alyiwang: added

metaphor/snowflake/utils.py
alyiwang: added

mars-lan (approved):

mars-lan (approved):

PR #446, opened 2022-12-29
Construct Tableau base URL from server_url & site_name from config

mars-lan:
### 🤔 Why?

The URL returned from the API may not be reachable, especially in the case of Tableau on-prem.

### 🤓 What?

As titled.

### 🧪 Tested?

Manually verified against a production instance.


alyiwang (approved):
LGTM

PR #448, opened 2023-01-05
initial MsSQL connector [sc-12761]

pk1230h:
### 🤔 Why?
initial the MsSQL connector and make MsSQL extractor as the library for Synapse

### 🤓 What?
- initial MsSQL connector
- share common SQL query code with Synapse and write into MsSQL extractor
-  remove duplicated methods in Synapse and use MsSQL methods instead
- add MsSQL readMe and adjust Synapse readMe

### 🧪 Tested?
Unit tests.
manual test  to deploy with ingestion API to dev environment


metaphor/synapse/README.md
```diff
@@ -38,7 +38,7 @@ Create a YAML config file based on the following template.
 ```yaml
 tenant_id: <tenant_id>  # The azure directory (tenant) id
 
-workspace_name: <workspace_name> # The Microsoft Synapse workspace name
```
pk1230h, line 41: change parameter because reuse code from mssql

metaphor/synapse/extractor.py
```diff
@@ -1,74 +1,73 @@
...
 
     def __init__(self, config: SynapseConfig):
-        super().__init__(config, "Synapse metadata crawler", Platform.SYNAPSE)
-        self._client = WorkspaceClient(
-            config.workspace_name, config.username, config.password
+        super().__init__(
+            config, "Synapse metadata crawler", Platform.SYNAPSE, DataPlatform.SYNAPSE
         )
-        self._tenant_id = config.tenant_id
+        self.config = config
```
elic-eon, line 34:
```suggestion
        self._config = config
```

elic-eon: Align with other class member?

tests/mssql/test_mssql_client.py
```diff
@@ -0,0 +1,176 @@
...
+        row.append(table.type)
+        row.append(table.create_time)
+        row.append(column.name)
+        row.append(column.max_length)
+        row.append(column.precision)
+        row.append(column.is_nullable)
+        row.append(column.type)
+        row.append(column.is_unique)
+        row.append(column.is_primary_key)
+        row.append(table.is_external)
```
elic-eon, line 83: Although this is testing code, why not use `row = [table.id, table.name, ... ]`?

tests/synapse/test_extractor.py
```diff
@@ -274,25 +260,26 @@ async def test_dedicated_sql_pool_extractor_with_query_log(test_root_dir):
...
+        mock_client.return_value = mock_client_instance
+        mock_query.return_value = [qyeryLogTable]
 
         config = synapse_config
         config.query_log = SynapseQueryLogConfig(lookback_days=1)
         extractor = SynapseExtractor(config)
 
         events = [EventUtil.trim_event(e) for e in await extractor.extract()]
+    # import json
+    # print(json.dumps(events, indent=2))
```
elic-eon, line 282: Drop these line?

metaphor/mssql/ReadMe.md
```diff
@@ -0,0 +1,50 @@
...
+## Setup
+We recommend creating a dedicated MsSQL user with limited permissions for the connector to use
+
+Use the following command to set up the read-only user for MsSQL database
+
+```sql
+-- nagivate to master database
+CREATE LOGIN <username> WITH PASSWORD = '<password>'
+CREATE USER <username> FROM LOGIN <username>
+-- switch to user  MsSQL database
```
elic-eon, line 14:
```suggestion
-- switch to user MsSQL database
```

metaphor/mssql/extractor.py
```diff
@@ -0,0 +1,159 @@
...
+            for database in client.get_databases():
+                tables = client.get_tables(database.name)
+                if len(tables) == 0:
+                    continue
+                datasets = self._map_tables_to_dataset(
+                    self.config.server_name, database, tables, client
+                )
+                entities.extend(datasets)
+        except Exception as error:
+            logger.exception(f"serverless sqlpool error: {error}")
```
elic-eon, line 66: Change the logging message?

metaphor/mssql/extractor.py
```diff
@@ -0,0 +1,159 @@
...
+
+        endpoint = f"{self.config.server_name}.database.windows.net"
+        client = MssqlClient(endpoint, self.config.username, self.config.password)
+
+        entities: List[ENTITY_TYPES] = []
+        try:
+            for database in client.get_databases():
+                tables = client.get_tables(database.name)
+                if len(tables) == 0:
+                    continue
```
elic-eon, line 60: Does this test necessary? `_map_tables_to_dataset` should return an empty result if we pass an empty list

metaphor/mssql/extractor.py
```diff
@@ -0,0 +1,159 @@
...
+        for fk in client.get_foreign_keys(database.name):
+            if fk.table_id in dataset_map:
+                foreign_key = ForeignKey(
+                    field_path=fk.column_name,
+                    parent=dataset_map[fk.referenced_table_id].logical_id,
+                    parent_field=fk.referenced_column,
+                )
+                dataset_map[fk.table_id].schema.sql_schema.foreign_key.append(
+                    foreign_key
+                )
```
elic-eon, line 141: Possible to move logic related to foreign keys out from `_map_tables_to_dataset`? Therefore, we don't have to pass the client object in.

metaphor/mssql/model.py
```diff
@@ -0,0 +1,63 @@
...
+    request_id: str
+    session_id: Optional[str]
+    sql_query: str
+    login_name: str
+    start_time: datetime
+    end_time: datetime
+    # in milliseconds
+    duration: int
+    # in MB
+    query_size: Optional[int]
```
elic-eon, line 60: Will `duration_in_ms` and `query_size_in_mb` be better here?

metaphor/mssql/model.py
```diff
@@ -0,0 +1,63 @@
+from datetime import datetime
+from typing import Dict, Optional
+
+from pydantic import BaseModel
+
+
+class MssqlConnectConfig(BaseModel):
```
elic-eon, line 7: Is `MssqlClientConfig` a better name here? We use this class in the `MssqlClient`, right?

metaphor/synapse/README.md
```diff
@@ -38,7 +38,7 @@ Create a YAML config file based on the following template.
...
+server_name: <workspace_name> # The Microsoft Synapse workspace name
```
elic-eon, line 41:
```suggestion
server_name: <workspace_name>  # The Microsoft Synapse workspace name
```

metaphor/synapse/README.md
```diff
@@ -38,7 +38,7 @@ Create a YAML config file based on the following template.
...
 
 username: <username> # The synapse workspace SQL username
```
elic-eon, line 43:
```suggestion
username: <username>  # The synapse workspace SQL username
```

metaphor/mssql/extractor.py
pk1230h: The reason to do this way is that I need to have completed dataset_map to add foreign_key.  
if not doing this way, the other way  take the dataset map object as return value and use another static method to add foreign. The method will look like:
```
def _get_foreign_key_for_dataset(
     dataset_map,
     database_name,
     client
)
```
is this the better approach?

metaphor/mssql/extractor.py
pk1230h: just fast check if empty don't need to go to the **_map_tables_to_dataset** and below logic, 
The foreign keys logic is not within tables logic and will trigger MsSQL fetch query call to check if foreign key existed. I  think it no need to make the call if we know table is not existed/

metaphor/mssql/extractor.py
elic-eon: Yes, we can return a map and pass into new function. I think two smaller functions would be better.

metaphor/mssql/extractor.py
elic-eon: My point is to reduce the number of indentions. If we can make sure all functions takes `tables` can handle empty list efficiently, would you think it better to drop these two lines?

metaphor/mssql/ReadMe.md
```diff
@@ -0,0 +1,50 @@
+# MsSQL Connector
+
+This connector extracts technical metadata from Azure Synapse workspaces using [System catalog views](https://learn.microsoft.com/en-us/sql/relational-databases/system-catalog-views/catalog-views-transact-sql?view=sql-server-ver16).
```
elic-eon, line 3:
```suggestion
This connector extracts technical metadata from a Microsoft SQL Server using [System catalog views](https://learn.microsoft.com/en-us/sql/relational-databases/system-catalog-views/catalog-views-transact-sql?view=sql-server-ver16).
```

metaphor/mssql/extractor.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def _map_tables_to_dataset(
+        self,
+        sql_server_name: str,
+        database: MssqlDatabase,
+        tables: List[MssqlTable],
+    ):
+        dataset_map: Dict[str, Dataset] = {}
+        if len(tables) == 0:
+            return dataset_map
```
elic-eon, line 76:
```suggestion
        if len(tables) == 0:
            return dataset_map

        dataset_map: Dict[str, Dataset] = {}
```

elic-eon (commented):
Otherwise, LGTM

PR #449, opened 2023-01-11
Use sqllineage to parse lineage from custom SQLs for Tableau data sources

mars-lan:
### 🤔 Why?

While Tableau claims to [compute lineage from custom SQL](https://help.tableau.com/current/pro/desktop/en-us/customsql.htm#supported-queries), in reality, the parser is rather unreliable and fails to process many queries. 

### 🤓 What?

- Retrieve all custom SQL queries using the Tableau Metadat API and use `sqllineage` to extract the source tables
- Refactor the extractor and unit tests to improve readability

### 🧪 Tested?

Manually tested against a production instance.


metaphor/tableau/query.py
```diff
@@ -15,6 +15,19 @@
 # the id of the same entity from the REST api, use luid instead
 workbooks_graphql_query = """
 query {
+  customSQLTables(filter:{id:"00b04ee0-1967-e302-563b-213e048429df"}) {
```
elic-eon, line 18: Is it a hard-code uuid?

metaphor/tableau/query.py
```diff
@@ -110,3 +123,54 @@ class WorkbookQueryResponse(BaseModel):
     vizportalUrlId: str
     upstreamDatasources: List[PublishedDatasource]
     embeddedDatasources: List[EmbeddedDatasource]
+
+
+# GraphQL that lists all custom SQL queries used in datasources.
+# To prevent the server from returning partial results when the query takes too long to run, we
+# 1. Run this as a separate query from the workbooks GraphQL
+# 2. Only return the first column as the datasource ID is the same for every column
```
elic-eon, line 131: Why do we only care about the first column?

mars-lan: As mentioned in the comment, all columsn will point to the same datasource ID so there's no need to fetch all of them.

mars-lan: No. `id` is the ID for the custom SQL table, which has nothing to do with datasource's ID.

metaphor/tableau/extractor.py
```diff
@@ -169,7 +190,70 @@ def _parse_dashboard(self, workbook: tableau.WorkbookItem) -> None:
...
+            if fullname.count(".") != 2:
+                logger.warn(f"Ignore non-fully qualified source table {fullname}")
+                continue
+
+            upstream_datasets.append(
+                str(to_dataset_entity_id(fullname, platform, account))
+            )
+
+        datasource_upstream_datasets = {}
+        for datasource_id in datasource_ids:
```
alyiwang, line 239: In the case of multiple datasources, is it correct to assign all the upstream datasets to each of the datasource?

metaphor/tableau/extractor.py
```diff
@@ -169,7 +190,70 @@ def _parse_dashboard(self, workbook: tableau.WorkbookItem) -> None:
...
+            return {}
+
+        try:
+            parser = LineageRunner(custom_sql_table.query)
+            source_tables = parser.source_tables
+        except Exception as e:
+            logger.error(f"Unable to parse custom query for {custom_sql_table.id}: {e}")
+            return {}
+
+        if len(source_tables) < 1:
```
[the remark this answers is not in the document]
mars-lan: Good point :)

metaphor/tableau/extractor.py
mars-lan: I don't think it'll happen in real life as each custom SQL belongs to a data source only. That said, if a custom SQL is ever shared across multiple data sources, it makes sense to assign the same upstream to all.

alyiwang (approved):
LGTM

PR #451, opened 2023-01-12
Save crawler run metadata and log file in a run-specific folder

elic-eon:
### 🤔 Why?

To capture each run metadata and log

### 🤓 What?

- Sink all output file in a run-specific folder

### 🧪 Tested?

Updated tests and verified locally.

### DIFF
--- a/metaphor/common/file_sink.py
+++ b/metaphor/common/file_sink.py
@@ -50,7 +50,7 @@ class FileSink(Sink):
         self.path = config.directory.rstrip("/")
         self.batch_size = config.batch_size
         self.write_logs = config.write_logs
-        self.run_dir = datetime.now(timezone.utc).strftime("%Y-%m-%d-%H%M%S")
+        self.run_dir = str(int(datetime.now().timestamp()))
 
         if config.directory.startswith("s3://"):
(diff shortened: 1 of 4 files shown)


metaphor/common/file_sink.py
```diff
@@ -50,6 +50,7 @@ def __init__(self, config: FileSinkConfig):
         self.path = config.directory.rstrip("/")
         self.batch_size = config.batch_size
         self.write_logs = config.write_logs
+        self.run_dir = datetime.now(timezone.utc).strftime("%Y-%m-%d %H-%M-%S")
```
mars-lan, line 53: Let's just use timestamp to avoid any potential issues with spaces in the S3 path?

elic-eon: Do you mean the unix timestamp, a number?
Changed to `%Y-%m-%d-%H%M%S` for now

mars-lan: Correct, e.g. `s3://crawler_x/1673544307/run.metadata`

elic-eon: Why not use a readable timestamp without space or other character?

mars-lan (approved):

PR #460, opened 2023-01-27
bump up library versions

alyiwang:
### 🤔 Why?

bring dependencies up-to-date

### 🤓 What?

- update library versions

> mypy upgrade to 0.991 raised ~100 of issues, and pydantic upgrade to 1.10 also needs some class validation config change, both are skipped in this upgrade and will be done in future PR

### 🧪 Tested?

tested several crawlers locally: snowflake, bigquery, looker, redshift


pyproject.toml
```diff
@@ -15,35 +15,34 @@ packages = [
...
 pymssql = { version = "^2.2.7", optional = true }
 pymysql = { version = "^1.0.2", optional = true }
 python = ">=3.7,<3.11"  # See https://github.com/googleapis/python-bigquery/issues/856
 python-dateutil = "^2.8.1"
 PyYAML = "^6.0"
-requests = "^2.27.1"
-smart-open = "^5.0.0"
-snowflake-connector-python = { version = "~=2.7.0", optional = true }
+requests = "^2.28.1"
+smart-open = "^6.3.0"
```
mars-lan, line 40: Any breaking changes?

alyiwang: Mostly minor version bumps, nothing major. Tested a few crawlers and they work fine. The impacting change would be mypy and pedantic, which required non-trivial code changes, and are skipped here.

mars-lan (approved):

PR #466, opened 2023-02-09
upgrade mypy and fix issues [sc-15688]

alyiwang:
### 🤔 Why?

To keep our code typing more robust

### 🤓 What?

- upgrade mypy to 0.991
- fix 100+ issues and bugs
- format some code

### 🧪 Tested?

all unit tests pass
run snowflake and bigquery crawlers locally and verified the results

### DIFF
--- a/metaphor/bigquery/config.py
+++ b/metaphor/bigquery/config.py
@@ -40,7 +40,6 @@ class BigQueryCredentials:
 
 @dataclass
 class BigQueryQueryLogConfig:
-
     # Number of days back of query logs to fetch, if 0, don't fetch query logs
     lookback_days: int = 1
 
@@ -56,14 +55,14 @@ class BigQueryQueryLogConfig:
 
 @dataclass
 class BigQueryRunConfig(BaseConfig):
+    # Project ID to use
+    project_id: str
+
     # Path to service account's JSON key file
     key_path: Optional[str] = None
 
     # The credentials from the BigQuery JSON key file
     credentials: Optional[BigQueryCredentials] = None
-
-    # Project ID to use. Use the service account's default project if not set
-    project_id: Optional[str] = None
 
     # Use a different project ID to run BigQuery jobs if set
     job_project_id: Optional[str] = None
--- a/metaphor/common/event_util.py
+++ b/metaphor/common/event_util.py
@@ -5,7 +5,7 @@ from typing import Union
 
 import fastjsonschema
 
-from metaphor import models
+from metaphor import models  # type: ignore
 from metaphor.models.metadata_change_event import (
     Dashboard,
     Dataset,
--- a/metaphor/power_bi/power_query_parser.py
+++ b/metaphor/power_bi/power_query_parser.py
@@ -181,7 +181,7 @@ class PowerQueryParser:
         return tables
 
     @staticmethod
-    def parse_native_query(
+    def _parse_native_query(
         power_query: str, snowflake_account: Optional[str]
     ) -> List[EntityId]:
         parameters = PowerQueryParser.extract_function_parameter(
@@ -203,7 +203,7 @@ class PowerQueryParser:
 
     @staticmethod
     def parse_source_datasets(
-        power_query: str, snowflake_account: Optional[str]
+        power_query: str, snowflake_account: Optional[str] = None
     ) -> List[EntityId]:
         def replacer(match: re.Match) -> str:
             controls = {
@@ -218,7 +218,7 @@ class PowerQueryParser:
         power_query = re.sub(r"#\((cr|lf|tab)(,(cr|lf|tab))*\)", replacer, power_query)
 
         if "Value.NativeQuery(" in power_query:
-            return PowerQueryParser.parse_native_query(power_query, snowflake_account)
+            return PowerQueryParser._parse_native_query(power_query, snowflake_account)
 
         try:
             return [PowerQueryParser.parse_power_query(power_query)]
--- a/metaphor/power_bi/power_query_parser.py
+++ b/metaphor/power_bi/power_query_parser.py
@@ -97,7 +97,6 @@ class PowerQueryParser:
             return text[start:end]
 
         while True:
-
             ch = text[index]
 
             if ch == '"':
@@ -123,7 +122,7 @@ class PowerQueryParser:
 
     @staticmethod
     def parse_platform(
-        exp: str, snowflake_account: str
+        exp: str, snowflake_account: Optional[str]
     ) -> Tuple[Optional[DataPlatform], Optional[str], Optional[str]]:
         """
         Parse platform information from native query expression.
@@ -182,7 +181,9 @@ class PowerQueryParser:
         return tables
 
     @staticmethod
-    def parse_native_query(power_query: str, snowflake_account: str) -> List[EntityId]:
+    def parse_native_query(
+        power_query: str, snowflake_account: Optional[str]
+    ) -> List[EntityId]:
         parameters = PowerQueryParser.extract_function_parameter(
             power_query, "NativeQuery"
         )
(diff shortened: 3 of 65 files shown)


metaphor/bigquery/config.py
```diff
@@ -56,15 +55,15 @@ class BigQueryQueryLogConfig:
...
+    project_id: str
+
     # Path to service account's JSON key file
     key_path: Optional[str] = None
 
     # The credentials from the BigQuery JSON key file
     credentials: Optional[BigQueryCredentials] = None
 
-    # Project ID to use. Use the service account's default project if not set
-    project_id: Optional[str] = None
```
mars-lan, line 66: Why changing this to required?

alyiwang: This is mentioned as a required config in https://github.com/MetaphorData/connectors/tree/main/metaphor/bigquery#required-configurations

mars-lan: SG

tests/power_bi/test_parse_power_query.py
```diff
@@ -5,14 +5,14 @@
 
 def test_parse_dataset_redshift():
     exp = 'let\n    Source = AmazonRedshift.Database("url:5439", "db" ),\n    public = Source{[Name="public"]}[Data],\n    table1 = public{[Name="table"]}[Data]\nin\n    table1'
-    assert PowerQueryParser.parse_source_datasets(exp) == [
+    assert PowerQueryParser.parse_source_datasets(exp, None) == [
```
mars-lan, line 8: Do we need to explicitly specify optional arguments as `None`?

tests/tableau/test_extractor.py
```diff
@@ -141,7 +141,7 @@ def test_parse_dataset_id():
             name="name",
             fullName="db.schema.table",
             database=Database(name="db", connectionType="redshift"),
-            schema_=None,
+            schema="foo",
```
mars-lan, line 144: Why changing these from `None` to `foo`?

tests/power_bi/test_parse_power_query.py
alyiwang: We can add default value `None` to the input param

metaphor/thought_spot/extractor.py
```diff
@@ -453,6 +458,7 @@ def _populate_source_virtual_views(
                 )
                 for viz, *_ in charts
                 for column in viz.vizContent.columns
+                if column.referencedTableHeaders
```
mars-lan, line 461: The complex comprehension is hard to enough to read in the first place, adding this `if` making it even worse. Also shouldn't it comes after the last for?

metaphor/tableau/extractor.py
```diff
@@ -360,7 +364,12 @@ def _parse_upstream_datasets(
         )
 
     def _parse_dataset_id(self, table: DatabaseTable) -> Optional[EntityId]:
-        if None in (table.name, table.schema, table.fullName, table.database):
+        if (
```
mars-lan, line 367: What was wrong with the previous code, which was more readable?

metaphor/snowflake/extractor.py
```diff
@@ -130,7 +129,7 @@ async def extract(self) -> Collection[ENTITY_TYPES]:
             if self._query_log_lookback_days > 0:
                 self._fetch_query_logs()
 
-        datasets = self._datasets.values()
+        datasets = list(self._datasets.values())
```
mars-lan, line 132: I thought `dict.values()` is already a `list`?

tests/power_bi/test_parse_power_query.py
mars-lan: Yeah might be more pythonic that way.

tests/tableau/test_extractor.py
alyiwang: This is to test the case that with `fullname`, the `schema` won't be used.

metaphor/tableau/extractor.py
alyiwang: for `if None in (table.name, table.schema, table.fullName, table.database)`, mypy won't be able to tell that all of the terms are non-null. Has to write it the other way. 😞

metaphor/snowflake/extractor.py
alyiwang: it's a view

metaphor/thought_spot/extractor.py
alyiwang: I personally find it pretty straightforward to read, the structure is the same as writing nested for loops. The `if` has to come after `column` but before `reference`, otherwise, mypy throws `None value is not iterable` for `column.referencedTableHeaders`

alyiwang: Just have to visually imagine there are indents.

mars-lan: Got it.

metaphor/tableau/extractor.py
mars-lan: Okay. In that case, using `is None` is probably safer than testing for falsey?

metaphor/snowflake/extractor.py
```diff
@@ -244,7 +243,7 @@ def _fetch_columns(self, cursor: SnowflakeCursor, database: str) -> None:
     def _fetch_table_info(
         self, tables: Dict[str, DatasetInfo], is_shared_database: bool
     ) -> None:
-        dict_cursor = self._conn.cursor(DictCursor)
+        dict_cursor: DictCursor = self._conn.cursor(DictCursor)  # type: ignore
```
mars-lan, line 246: Why ignore?

metaphor/power_bi/power_query_parser.py
```diff
@@ -217,7 +218,7 @@ def replacer(match: re.Match) -> str:
         power_query = re.sub(r"#\((cr|lf|tab)(,(cr|lf|tab))*\)", replacer, power_query)
 
         if "Value.NativeQuery(" in power_query:
-            return PowerQueryParser.parse_native_query(power_query, snowflake_account)
+            return PowerQueryParser._parse_native_query(power_query, snowflake_account)
 
         try:
             return [PowerQueryParser.parse_power_query(power_query)]
```
mars-lan, line 224: Also mark `parse_power_query` as private for consistency?

metaphor/common/event_util.py
```diff
@@ -5,7 +5,7 @@
 
 import fastjsonschema
 
-from metaphor import models
+from metaphor import models  # type: ignore
```
mars-lan, line 8: Why ignore?

metaphor/snowflake/extractor.py
alyiwang: `connection.cursor()` method has dynamic typing based on the given type param. But the method's returned type is always labeled `SnowflakeCursor`, and mypy throws `SnowflakeCursor` not assignable to `DictCursor`.

metaphor/common/event_util.py
alyiwang: Mypy: `model` is not a module under metaphor.
We use this import to load the schema json file for validation.

metaphor/power_bi/power_query_parser.py
alyiwang: updated

metaphor/tableau/extractor.py
alyiwang: falsy also test for an empty string which we don't want either.

elic-eon:
https://mypy-lang.blogspot.com/2023/02/mypy-10-released.html Mypy 1.0 was released.

alyiwang:
> https://mypy-lang.blogspot.com/2023/02/mypy-10-released.html Mypy 1.0 was released.

I tried that 😄 . But mypy 1.0 requires python 3.8+, and we are at 3.7+ now.

mars-lan (approved):
Thanks for fixing these!

PR #469, opened 2023-02-24
Add support for dbt manifest v8

mars-lan:
### 🤔 Why?

dbt introduced a new manifest version since 1.4: https://docs.getdbt.com/reference/artifacts/manifest-json. 
It combined many "parsed" & "complied" classes into a single generic class but not much material changes beyond that.

### 🤓 What?

As titled.

### 🧪 Tested?

Verified against a manifest produced using dbt 1.4.


alyiwang (approved):
Thanks for updating this!

PR #470, opened 2023-02-28
Fix Power BI cross-workspace dataset references

mars-lan:
### 🤔 Why?

Depending the workspace processing order, the crawler may crash when processing Power BI dashboards/reports that reference a dataset from another workspace with the following error:

```
023-02-27 13:29:39:ERROR:metaphor:'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'
Traceback (most recent call last):
  File "/Users/marslan/Metaphor/connectors/metaphor/power_bi/extractor.py", line 92, in extract
    self.map_wi_reports_to_dashboard(workspace, app_map)
  File "/Users/marslan/Metaphor/connectors/metaphor/power_bi/extractor.py", line 191, in map_wi_reports_to_dashboard
    self._virtual_views[wi_report.datasetId].logical_id,
KeyError: 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'
```

### 🤓 What?

- Process datasets across all workspaces up front to avoid invalid cross-workspace references in datasets & reports. 
- Refactored various exception handling avoid log cluttering and breaks the crawler when unhandled exceptions occur.
- Use `from None` to avoid printing nested exception context when rethrowing exceptions (see [PEP-409](https://peps.python.org/pep-0409/)).

### 🧪 Tested?

### DIFF
--- a/metaphor/power_bi/extractor.py
+++ b/metaphor/power_bi/extractor.py
@@ -150,12 +150,8 @@ class PowerBIExtractor(BaseExtractor):
 
             last_refreshed = None
             if ds.isRefreshable:
-                try:
-                    # Log errors instead of throwing exceptions as this API call requires extra permissions
-                    refreshes = self._client.get_refreshes(workspace.id, wds.id)
-                    last_refreshed = self._find_last_completed_refresh(refreshes)
-                except Exception as e:
-                    logger.exception(e)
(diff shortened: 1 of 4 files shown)


elic-eon (approved):
LGTM

PR #474, opened 2023-03-31
BigQuery truncate long SQL in audit log [sc-16983]

alyiwang:
### 🤔 Why?

BigQuery audit metadata log has a size limit 100KB
https://cloud.google.com/bigquery/docs/reference/auditlogs#limitation
Longer SQL will be truncated, thus we need to call `jobs.get` to get the full job spec.

### 🤓 What?

- Check `queryTruncated` flag and call `jobs` API to get full query SQL

### 🧪 Tested?

tested locally against BigQuery instance

### DIFF
--- a/metaphor/bigquery/config.py
+++ b/metaphor/bigquery/config.py
@@ -52,6 +52,9 @@ class BigQueryQueryLogConfig:
     # The number of query logs to fetch from BigQuery in one batch
     fetch_size: int = DEFAULT_QUERY_LOG_FETCH_SIZE
 
+    # Fetch the full query SQL from job API if it's truncated in the audit metadata log
+    fetch_job_query_if_truncated: bool = True
+
 
 @dataclass
 class BigQueryRunConfig(BaseConfig):
--- a/metaphor/bigquery/extractor.py
+++ b/metaphor/bigquery/extractor.py
@@ -1,3 +1,4 @@
+import re
 from concurrent.futures import ThreadPoolExecutor
 from typing import (
     Any,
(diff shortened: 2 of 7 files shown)


metaphor/bigquery/extractor.py
```diff
@@ -281,7 +284,23 @@ def _fetch_query_logs(self, logging_client: logging_v2.Client) -> List[QueryLog]
...
 
-    def _parse_job_change_entry(self, entry: LogEntry) -> Optional[QueryLog]:
+    @staticmethod
+    def _fetch_job_query(client: bigquery.Client, job_name: str) -> Optional[str]:
+        logger.info(f"Query {job_name}")
+        match = re.match(r"^projects/([^/]+)/jobs/([^/]+)$", job_name)
+        if match:
+            project = match.group(1)
+            job_id = match.group(2)
+            job = client.get_job(job_id, project)
```
mars-lan, line 294: Would this cause any performance issues? Hopefully truncation is rare and the API returns quickly?

alyiwang: It will certainly be slower than before, since it's an additional function call. If the customer has many large queries, then it might take a long time to finish. I only tried a couple, and the call returned very fast.

mars-lan: Any additional permissoins needed? Consider adding a config to control this behavior?

alyiwang: We already asked for the `job user` permission, so no more is needed. Added a config to disable the job API call if set to false.

metaphor/bigquery/config.py
```diff
@@ -52,6 +52,9 @@ class BigQueryQueryLogConfig:
     # The number of query logs to fetch from BigQuery in one batch
     fetch_size: int = DEFAULT_QUERY_LOG_FETCH_SIZE
 
+    # Fetch the full query SQL from job API if it's truncated in the audit metadata log
+    fetch_job_query_if_truncated: bool = True
```
mars-lan, line 56: Not clear that this config was actually used?

alyiwang: Good catch! I got distracted half-way through adding the config.

mars-lan (approved):

PR #486, opened 2023-05-01
ambiguous column name error for Snowflake [sc-17430]

alyiwang:
### 🤔 Why?

It seems that the ACCESS_HISTORY view also has a column "USER_NAME" now, so when joining it with QUERY_HISTORY view and adding the exclude username clause, it causes ambiguous column error

### 🤓 What?

- add table qualifier to the "USER_NAME" column, the joining query already used `q` as the alias, so we reused that

### 🧪 Tested?

added `excluded_usernames` config and crawler our testing Snowflake instance, previous code will throw error, but the updated code works fine.


mars-lan (approved):
Thanks for the quick fix.

PR #491, opened 2023-05-03
Store bytes instead of MBs for dataset size [sc-17562]

alyiwang:
### 🤔 Why?

to keep more precise stats of data size

### 🤓 What?

- store data size bytes in addition to existing MB

### 🧪 Tested?

tested using snowflake and bigquery crawler and verified the results


mars-lan (approved):
Thanks for the refactoring. We may lose some precisions for large numbers when parsing the JSON string back to JavaScript, but it should be fine for display purposes.

PR #498, opened 2023-05-18
Add support for dbt 1.5

mars-lan:
### 🤔 Why?

[dbt 1.5](https://docs.getdbt.com/docs/dbt-versions/core) is now GA'ed and introduced a new version of manifest ([v9](https://docs.getdbt.com/reference/artifacts/manifest-json)).
### 🤓 What?

While the [changes](https://docs.getdbt.com/guides/migration/versions/upgrading-to-v1.5#for-consumers-of-dbt-artifacts-metadata) do not impact the metadata we care about, it does require some minor updates to `manifest_parser.py`.

### 🧪 Tested?

Tested against a production instance.


alyiwang (approved):
Thanks for updating this

PR #500, opened 2023-05-21
Add support to specify Snowflake private key data directly in the config

mars-lan:
### 🤔 Why?

Sometimes it's more convenient to specify the content of the private key directly in the config file, instead of pointing it to an external file.

### 🤓 What?

As titled.

### 🧪 Tested?

Tested end-to-end against a production instance.


alyiwang (approved):
LGTM

PR #504, opened 2023-05-24
Normalize PrivateLink Snowflake account names for dbt & FiveTran

mars-lan:
### 🤔 Why?

Some systems (e.g. dbt & Fivetran) support [Snowflake PrivateLink](https://docs.snowflake.com/en/user-guide/admin-security-privatelink) and use the PrivateLink version of the account name (i.e. `<normal_account_name>.privatelink`). This can cause issues when matching datasets across systems.

### 🤓 What?

Added `normalize_snowflake_account` function that is currently limited to stripping the `.privatelink` suffix from the account name. Will address any other variations listed in https://docs.snowflake.com/en/user-guide/admin-account-identifier#label-account-name in future PRs.

### 🧪 Tested?

Tested against a production instance and manually verified the MCEs.

### DIFF
--- a/metaphor/common/snowflake.py
+++ b/metaphor/common/snowflake.py
@@ -1,0 +1,14 @@
+PRIVATE_LINK_SUFFIX = ".privatelink"
+
+
+def normalize_snowflake_account(account: str) -> str:
+    """
+    Normalize different variations of Snowflake account.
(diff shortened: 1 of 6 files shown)


alyiwang (approved):
LGTM

PR #506, opened 2023-05-25
[Tableau] Output placeholder upstream datasets in MCE

mars-lan:
### 🤔 Why?

Currently the connector reference upstream datasets using entity IDs hoping that the datasets have already been created by the corresponding dataset connector (e.g. Snowflake, BigQuery, etc.). However, this can become confusing and very difficult to debug for the users. It'd be better to create placeholder datasets so at least they show up in the lineage. 

### 🤓 What?

As titled.

### 🧪 Tested?

Tested against a production instance and verified the MCEs manually.


elic-eon (approved):
LGTM

PR #509, opened 2023-05-26
Use size-based chunking when spliting MCE files

mars-lan:
### 🤔 Why?

The current size-based chunking can lead to large files (100's of MB) for large individual items. These large files are slow to process and can also lead to OOM issues.

### 🤓 What?

- Add a `batch_size_bytes` config to replace `batch_size`
- Add a generic `chunk_by_size` util method
- Update `FileSink` to chunk output by the size JSON-serialized string

### 🧪 Tested?

Tested against a production instance:

Screenshot 2023-05-26 at 12 08 04 PM

### DIFF
--- a/metaphor/common/file_sink.py
+++ b/metaphor/common/file_sink.py
@@ -1,6 +1,5 @@
 import json
 import logging
-import math
 import tempfile
 from dataclasses import field
 from datetime import datetime, timezone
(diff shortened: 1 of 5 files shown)


alyiwang (approved):
LGTM
thanks for the fix!

PR #512, opened 2023-05-28
Remove unncessary setting of entity_type field

mars-lan:
### 🤔 Why?

The field is a constant and shouldn't be set by the connectors.

### 🤓 What?

As titled.

### 🧪 Tested?

Verified end-to-end that the generated MCEs can be ingested.


elic-eon (approved):
Thanks for the refactoring

PR #514, opened 2023-05-31
Fivetran crawler should create "placeholder nodes" for systems that are natively supported.

elic-eon:
### 🤔 Why?

To complete the lineage for known platform when we don't have the corresponding crawler, for example, SQL server.

### 🤓 What?

- Add dummy dataset for source.

### 🧪 Tested?

Update tests


mars-lan (approved):
Thanks for the quick fix. Please also verify e2e against production.

PR #515, opened 2023-05-31
[Snowflake] Make sure upstream datasets contain only unique IDs

mars-lan:
### 🤔 Why?

Encountered the following error when ingesting Snowflake Lineage: 

```
MCE parsing error Error: Class MetadataChangeEvent2 validation error:  "sourceDatasets" is not a valid unique array of entity IDs
```

Turns out that Snowflake's [OBJECT_DEPENDENCY view](https://docs.snowflake.com/en/sql-reference/account-usage/object_dependencies) can contain rows with repeated content, leading to the same IDs added to `source_datasets` field multiple times.

### 🤓 What?

Make sure the entity ID is unique before appending it to `source_datasets` in `_parse_object_dependencies`.

### 🧪 Tested?

Verified against a production instance

### DIFF
--- a/metaphor/snowflake/lineage/extractor.py
+++ b/metaphor/snowflake/lineage/extractor.py
@@ -264,9 +264,11 @@ class SnowflakeLineageExtractor(BaseExtractor):
             )
 
             if target_normalized_name in self._datasets:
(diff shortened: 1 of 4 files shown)


alyiwang (approved):
LGTM

PR #516, opened 2023-06-01
Fivetran validation error

elic-eon:
### 🤔 Why?

First, turned out the "name_in_source" was optional. Second, the validation part was too strict if the model didn't match the api response, the whole crawler stop.

### 🤓 What?

- Make the field optional, verified the output json dump file, name_in_source in `SchemaMetadata` could be null
- add `try...except` around `parse_obj_as` to add more context to log.
- Add a random string to debug file to avoid duplicated name, repeated call with different query or body.

### 🧪 Tested?

Test against production environment.

### DIFF
--- a/metaphor/common/api_request.py
+++ b/metaphor/common/api_request.py
@@ -1,10 +1,11 @@
 import json
+import secrets
 import tempfile
 from typing import Any, Callable, Dict, Type, TypeVar
 from urllib.parse import urlparse
 
 import requests
-from pydantic import parse_obj_as
+from pydantic import ValidationError, parse_obj_as
 
 from metaphor.common.logger import debug_files, get_logger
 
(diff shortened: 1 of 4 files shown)


metaphor/common/api_request.py
```diff
@@ -31,7 +32,9 @@ def get_request(
     result = requests.get(url, headers=headers, timeout=timeout, **kwargs)
     if result.status_code == 200:
         # Add JSON response to log.zip
-        file_name = f"{urlparse(url).path[1:].replace('/', u'__')}"
+        file_name = (
+            f"{urlparse(url).path[1:].replace('/', u'__')}_{secrets.token_hex(1)}"
```
mars-lan, line 36: Does this really create a different filename each time? Wouldn't `secrets` be the same for each request?

elic-eon: Yes, it's a 2 char length random string.

mars-lan: Nvm. I thought it was the secret used to authenticate the request. This works.

mars-lan (approved):

PR #518, opened 2023-06-06
Support column level Snowflake tags [sc-18488]

alyiwang:
### 🤔 Why?

Support column level Snowflake tags  

### 🤓 What?

- extract tags for columns from snowflake

> Note that the [TAG_REFERENCES View](https://docs.snowflake.com/en/sql-reference/account-usage/tag_references) takes up to 2hr to populate

### 🧪 Tested?

tested locally against snowflake instance, and verified MCE


mars-lan (approved):
Thanks for adding this quickly!

PR #519, opened 2023-06-09
Support extracting metadata from multiple BQ projects in a single run

mars-lan:
### 🤔 Why?

It's common to have BigQuery instances across large numbers of GCP projects. Setting up a connector for each project quickly becomes cumbersome and error-prone.

### 🤓 What?

- Add `project_ids` to support specifying multiple projects in a single config
- Update BigQuery base, lineage, and profile connectors to support multiple projects 

### 🧪 Tested?

Ran against a production instance and verified that the before & after MCEs are identical.


metaphor/bigquery/config.py
```diff
@@ -58,8 +58,11 @@ class BigQueryQueryLogConfig:
 
 @dataclass
 class BigQueryRunConfig(BaseConfig):
-    # Project ID to use
-    project_id: str
+    # Deprecated: Use project_ids instead
+    project_id: Optional[str] = None
```
alyiwang, line 62: We will keep existing customer configs as is, right?

alyiwang (approved):
LGTM

metaphor/bigquery/config.py
mars-lan: Yes. We'll keep supporting `project_id` until all clients have migrated.

PR #520, opened 2023-06-10
Update to Python 3.8

mars-lan:
### 🤔 Why?

Python 3.7 will enter in 2 weeks: https://devguide.python.org/versions/. Updating to Python 3.8 as part of [v0.12]( https://github.com/MetaphorData/connectors/issues/310).

### 🤓 What?

- Update required Python version to 3.8
- Use `@patch` decorators in tests as it's finally supported for async functions in Python 3.8
- Fix various typing issues


alyiwang (approved):
Great refactoring! Much cleaner now

PR #525, opened 2023-06-10
Drop project_id from BigQueryRunConfig

mars-lan:
### 🤔 Why?

The field has been deprecated and should be dropped as part of [v0.12](https://github.com/MetaphorData/connectors/issues/310.)

### 🤓 What?

As titled.

### 🧪 Tested?

Ran against a production instance to make sure the refactoring works.


alyiwang (approved):
Did we migrate all the current customer configs to use project_ids?

mars-lan:
> Did we migrate all the current customer configs to use project_ids?

Yup

PR #526, opened 2023-06-10
Rename metaphor.manual to metaphor.custom

mars-lan:
### 🤔 Why?

The name `manual` can be misleading and doesn't convey the purpose well. Renaming it to `custom` as part of [v0.12](https://github.com/MetaphorData/connectors/issues/310).

### 🤓 What?

As titled. Also updated various class names, README, and test data paths for consistency.

> Note: GitHub sometimes has difficulty diff'ing small files that have been moved and subsequently edited.

### 🧪 Tested?

Tested the renamed connectors locally.


alyiwang (approved):
Thanks for updating the README too!

PR #528, opened 2023-06-11
upgrade pydantic to 1.10

alyiwang:
### 🤔 Why?

pydantic 1.9 is outdated (1.10 is out in 2022.8 and v2.0 alpha has been released), and because there are some breaking changes, it prevents us from upgrading some other dependencies, such as apache-airflow.

### 🤓 What?

- upgrade pydantic to latest 1.10 version
- since the 1.10 pydantic dataclass default behavior is to ignore extra fields, we need to add the config to force it to throw an error as before. 

### 🧪 Tested?

all unit tests passed
local run snowflake, bigquery, power_bi, thought_spot crawlers

### DIFF
--- a/metaphor/common/dataclass.py
+++ b/metaphor/common/dataclass.py
@@ -1,0 +1,10 @@
+from pydantic import Extra
+
+
+class DataclassConfig:
+    """
+    Config for pydantic dataclass and BaseModel
+    """
+
+    validate_all = True
+    extra = Extra.forbid
--- a/metaphor/common/dataclass.py
+++ b/metaphor/common/dataclass.py
@@ -1,9 +1,9 @@
 from pydantic import Extra
 
 
-class DataclassConfig:
+class ConnectorConfig:
     """
-    Config for pydantic dataclass and BaseModel
+    Pydantic dataclass Config for metaphor connector configurations
     """
 
     validate_all = True
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.12.0"
+version = "0.12.1"
 license = "Apache-2.0"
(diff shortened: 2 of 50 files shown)


metaphor/common/dataclass.py
```diff
@@ -0,0 +1,10 @@
+from pydantic import Extra
+
+
+class DataclassConfig:
```
mars-lan, line 4: Consider a more explicit name for the class, e.g. `ForbidExtraFields`.

mars-lan (commented):
Thanks for refactoring!

metaphor/common/dataclass.py
alyiwang: This was meant for all the configs for dataclass and BaseModel, there are dozens of options, in the future if we decided to add other options, such as ser-de,  we don't have to rename the config class again.

mars-lan: For config classes it makes sense to forbid extra fields. For API responses, it's actually better to use ignore or allowed, e.g. https://github.com/MetaphorData/connectors/blob/9a4f547efdc4b61b2c4d99a5a86f60bf013ee9bd/metaphor/power_bi/power_bi_client.py#L22

Another way is to use a name that conveys the purpose, e.g. `ConfigClass`.

alyiwang: Got it. Updated the name to `ConnectorConfig` and used only on crawler config definitions.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.12.0"
+version = "0.12.1"
```
mars-lan, line 3: Bump again?

alyiwang: updated

mars-lan (approved):

PR #532, opened 2023-06-14
Lower case the Snowflake account name when normalizing

mars-lan:
### 🤔 Why?

While not explicitly documented in https://docs.snowflake.com/en/user-guide/admin-account-identifier, the Snowflake account name appears to be case insensitive. This makes sense since it's used as part of the URL (e.g. `https://acme-marketing-test-account.snowflakecomputing.com`). We need to normalize it as such across systems.

### 🤓 What?

As titled.

### 🧪 Tested?

Unit tests.


elic-eon (approved):
LGTM

PR #535, opened 2023-06-20
Table not found error in Bigquery lineage [sc-18780]

alyiwang:
### 🤔 Why?

For some reason, the `get_table` may throw `Table Not Found` error even though the tables are from `list_tables` API. 

### 🤓 What?

- put the `get_table` API call inside the try-catch so it won't crash the crawler

### 🧪 Tested?

tested against metaphor instance, didn't encounter such issue.


mars-lan (approved):
When there are a lot of tables to process, it's possible that the specific table has already been dropped by the time it's processed.

PR #542, opened 2023-06-28
Extract dataset URL from Monte Carlo

mars-lan:
### 🤔 Why?

To fill in the `url` field, which can be used to access the dataset page on Monte Carlo side.

### 🤓 What?

- Extract MC dataset URL
- Normalize column names
- Store the raw GraphQL response JSON in `log.zip` to help debug issues.

### 🧪 Tested?

Manually tested against a production instance.


elic-eon (approved):
LGTM

PR #543, opened 2023-07-03
Display Tableau nested hierarchy in metaphor [sc-18850]

alyiwang:
### 🤔 Why?

The current tableau workbook extractor only used the immediate project name but did not include the parent project name. This will result in a flattened hierarchy. 

### 🤓 What?

- fetch all projects and build each project full name
- use project id to link to the project full name of each workbook

### 🧪 Tested?

tested against tableau server

### DIFF
--- a/metaphor/tableau/extractor.py
+++ b/metaphor/tableau/extractor.py
@@ -73,8 +73,9 @@ class TableauExtractor(BaseExtractor):
         self._bigquery_project_name_to_id_map = config.bigquery_project_name_to_id_map
         self._disable_preview_image = config.disable_preview_image
 
-        self._datasets: Dict[EntityId, Dataset] = {}
         self._views: Dict[str, tableau.ViewItem] = {}
+        self._projects: Dict[str, str] = {}
+        self._datasets: Dict[EntityId, Dataset] = {}
         self._virtual_views: Dict[str, VirtualView] = {}
         self._dashboards: Dict[str, Dashboard] = {}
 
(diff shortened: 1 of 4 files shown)


metaphor/tableau/extractor.py
```diff
@@ -182,19 +191,33 @@ def _extract_datasources(self, server: tableau.Server) -> None:
...
+            self._projects.get(workbook.project_id, None) or workbook.project_name
+        )
 
         views: List[tableau.ViewItem] = workbook.views
         charts = [self._parse_chart(self._views[view.id]) for view in views if view.id]
         total_views = sum([view.total_views for view in views])
 
         dashboard_info = DashboardInfo(
-            title=f"{workbook.project_name}.{workbook.name}",
+            title=f"{project_name}.{workbook.name}",
```
mars-lan, line 220: Seems like a bit of a hack? We should probably keep track of the hierarchy in a separate array field, but this is okay for now.

alyiwang: Agree, we should probably add a `structure` field for dashboard and virtual view later, like we did with dataset.

alyiwang: https://app.shortcut.com/metaphor-data/story/19174

mars-lan (approved):

metaphor/tableau/extractor.py
```diff
@@ -182,19 +191,34 @@ def _extract_datasources(self, server: tableau.Server) -> None:
                     f"failed to parse workbook {item['vizportalUrlId']}, error {error}"
                 )
 
+    def _parse_project_names(self, projects: List[tableau.ProjectItem]) -> None:
+        for project in projects:
+            logger.debug(json.dumps(project.__dict__, default=str))
```
mars-lan, line 196: Probably not needed since we add the raw JSON output to log.zip?

alyiwang: good catch, removed

mars-lan (approved):

PR #545, opened 2023-07-03
fix tableau asset name when there is no project

alyiwang:
### 🤔 Why?

There are cases that the tableau dashboard or datasource doesn't have project or under personal space. We don't want the name to like `None.xxx`

### 🤓 What?

- add `<project>` to name only when the project exists.

### 🧪 Tested?

tested against Tableau server and verified MCE.


metaphor/tableau/extractor.py
```diff
@@ -147,6 +148,7 @@ def _extract_dashboards(self, server: tableau.Server) -> None:
         )
         for workbook in workbooks:
             server.workbooks.populate_views(workbook, usage=True)
+            logger.info(json.dumps(workbook.__dict__, default=str))
```
mars-lan, line 151: Drop?

metaphor/tableau/extractor.py
```diff
@@ -214,7 +216,7 @@ def _parse_dashboard(self, workbook: tableau.WorkbookItem) -> None:
         total_views = sum([view.total_views for view in views])
 
         dashboard_info = DashboardInfo(
-            title=f"{project_name}.{workbook.name}",
+            title=f"{project_name}.{workbook.name}" if project_name else workbook.name,
```
mars-lan, line 219: Moved the logic to an util function?

alyiwang: updated

metaphor/tableau/extractor.py
```diff
@@ -201,22 +199,32 @@ def _parse_project_names(self, projects: List[tableau.ProjectItem]) -> None:
                 parent_name = self._projects[project.parent_id]
                 self._projects[project.id] = f"{parent_name}.{project.name}"
 
+    def _build_asset_name(
```
mars-lan, line 202: Add a test as well? Thanks.

alyiwang: added

mars-lan (approved):

PR #547, opened 2023-07-06
Update Fivetran crawler to output Pipeline [sc-19032]

alyiwang:
### 🤔 Why?

We made pipeline a first-class citizen, now we should make use of it in the crawler

### 🤓 What?

- Update Fivetran crawler to output Pipeline
- Link dataset upstream to pipeline entity
- Remove `entityUpstream` aspect since it's a exact duplicate of `datasetUpstream`, and the resolver should be able to fallback to `datasetUpstream` automatically. 

### 🧪 Tested?

tested against fivetran server and verified MCE


tests/fivetran/test_extractor.py
```diff
@@ -95,4 +96,6 @@ async def test_extractor(mock_get: MagicMock, test_root_dir: str):
     extractor = FivetranExtractor(dummy_config())
     events = [EventUtil.trim_event(e) for e in await extractor.extract()]
 
+    print(json.dumps(events))
```
mars-lan, line 99: Drop?

metaphor/fivetran/extractor.py
```diff
@@ -263,16 +274,28 @@ def _init_dataset(
                 platform=destination_platform,
                 account=self.get_snowflake_account_from_config(destination.config),
             ),
-            upstream=DatasetUpstream(source_datasets=[], field_mappings=[]),
+            entity_upstream=EntityUpstream(
```
mars-lan, line 277: Let's stick to `DatasetUpstream` for now? We can do a complete migration later?

alyiwang: updated

mars-lan (approved):

PR #548, opened 2023-07-07
PowerBI crawler parsing error on not existing workspace

elic-eon:
### 🤔 Why?

We will get a workspace scan result without `name` and `type` for not existing workspace.

### 🤓 What?

- Make these two field optional in the model

### 🧪 Tested?

Verified against production env.

### DIFF
--- a/metaphor/power_bi/power_bi_client.py
+++ b/metaphor/power_bi/power_bi_client.py
@@ -352,6 +352,8 @@ class PowerBIClient:
             transform_response=transform_scan_result,
         )
 
+        # Filter workspace without name or type
+        # The scan result will return workspace without name or type if the workspace is not exist
         return [
             workspace for workspace in workspaces if workspace.name and workspace.type
         ]
--- a/metaphor/power_bi/power_bi_client.py
+++ b/metaphor/power_bi/power_bi_client.py
@@ -126,8 +126,8 @@ class WorkspaceInfoReport(BaseModel):
 
 class WorkspaceInfo(BaseModel):
     id: str
-    name: str
-    type: str
+    name: Optional[str]
+    type: Optional[str]
     state: str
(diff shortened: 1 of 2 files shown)


metaphor/power_bi/power_bi_client.py
```diff
@@ -346,12 +346,16 @@ def transform_scan_result(response: requests.Response) -> dict:
...
         # https://docs.microsoft.com/en-us/rest/api/power-bi/admin/workspace-info-get-scan-result
         url = f"{self.API_ENDPOINT}/admin/workspaces/scanResult/{scan_id}"
-        return self._call_get(
+        workspaces = self._call_get(
             url,
             List[WorkspaceInfo],
             transform_response=transform_scan_result,
         )
 
+        return [
```
mars-lan, line 355: Add a comment on why filtering.

metaphor/power_bi/power_bi_client.py
```diff
@@ -352,6 +352,8 @@ def transform_scan_result(response: requests.Response) -> dict:
             transform_response=transform_scan_result,
         )
 
+        # Filter workspace without name or type
+        # The scan result will return workspace without name or type if the workspace is not exist
```
mars-lan, line 356:
```suggestion
        # The scan result will return workspaces without name or type if the workspace does not exist
```

metaphor/power_bi/power_bi_client.py
```diff
@@ -352,6 +352,8 @@ def transform_scan_result(response: requests.Response) -> dict:
             transform_response=transform_scan_result,
         )
 
+        # Filter workspace without name or type
```
mars-lan, line 355: Feel free to drop this line as it doesn't add value beyond the next line.

mars-lan (approved):

PR #554, opened 2023-07-16
Add region_name to S3StorageConfig

mars-lan:
### 🤔 Why?

It was missing before, which prevents users from specifying all S3 credentials via config files.

### 🤓 What?

As titled. Also updated README & added tests.

### 🧪 Tested?


alyiwang (approved):
LGTM

PR #555, opened 2023-07-16
NaN value in MCE [sc-19559]

alyiwang:
### 🤔 Why?

The data profiler may receive values that's python `NaN`, and that's put into JSON like `"maxValue": NaN` which can't be decoded.

### 🤓 What?

- convert `NaN` value to None in the profiler

### 🧪 Tested?

unit tests


mars-lan (approved):
LGTM!

Issue #523, opened 2023-06-10
Breaking changes in 0.13


mars-lan:
- https://github.com/MetaphorData/connectors/pull/561
- https://github.com/MetaphorData/connectors/pull/562
- https://github.com/MetaphorData/connectors/pull/565
- https://github.com/MetaphorData/connectors/pull/619
- https://github.com/MetaphorData/connectors/pull/620

PR #561, opened 2023-07-30
Drop support for dbt 0.x

mars-lan:
### 🤔 Why?

dbt Labs has long [stopped supporting](https://docs.getdbt.com/docs/dbt-versions/core) dbt 0.x. While 1.0 & 1.1 are deprecated, they are still available from dbt Cloud and thus should be supported for now.

### 🤓 What?

### 🧪 Tested?


alyiwang (approved):
Thanks for the clean up!

PR #562, opened 2023-07-30
Drop airflow plugin

mars-lan:
### 🤔 Why?

The Airflow plugin is no longer needed now that we've [adopted OpenLineage](https://docs.metaphor.io/docs/airflow).

### 🤓 What?

### 🧪 Tested?

CI should verify the changes.


pyproject.toml
```diff
@@ -39,6 +39,7 @@ python = ">=3.8.1,<4.0"
 python-dateutil = "^2.8.1"
 PyYAML = "^6.0"
 requests = "^2.28.1"
+setuptools = "^68.0.0"
```
alyiwang, line 42: Why need this?

alyiwang (commented):
Finally!!
Great works on the airflow openlineage!

pyproject.toml
mars-lan: It was needed by MC and maybe other connectors as well. It was brought in transitively by apache-airflow.

alyiwang (approved):

PR #565, opened 2023-07-31
Drop support for API sink

mars-lan:
### 🤔 Why?

We have long deprecated API-based ingestion. It's time to drop the support completely.

### 🤓 What?

### 🧪 Tested?

Verified by running a few connectors against production instances.


alyiwang (approved):
LGTM

PR #572, opened 2023-08-15
Update dbt v10 manifest

mars-lan:
### 🤔 Why?

Since https://github.com/dbt-labs/dbt-core/issues/8328 has been fixed, it's time to pull in the latest (and fixed) schema.

### 🤓 What?

As titled. Continue to bypass dbt 1.6 metrics parsing for now. 

### 🧪 Tested?


alyiwang (approved):
LGTM

PR #578, opened 2023-08-24
Fix custom dashboard example

mars-lan:
### 🤔 Why?

It wasn't running correctly out of the box.

### 🤓 What?

As titled. Also showcase how to set the `structure` field and S3 auth credentials.

### 🧪 Tested?

Verified the example end-to-end.


elic-eon (approved):
LGTM

PR #580, opened 2023-08-28
Power BI dataset column level lineage [sc-12334]

alyiwang:
### 🤔 Why?

CLL can give more detailed information on how the PowerBI dataset is generated from the source dataset. As a first step, we generate 1 to 1 field mapping if the power query directly imports from the source dataset.

### 🤓 What?

- For direct import dataset, generate 1 to 1 field mapping in entity upstream

### 🧪 Tested?

tested against power BI server, and verified the MCE


elic-eon (approved):
LGTM

PR #581, opened 2023-08-29
Azure data factory crawler

elic-eon:
### 🤔 Why?

To extract table lineage information for Azure Data Factory pipeline. Note that this PR only cover `SnowflakeTable` and `dataflow` table lineage. 

### 🤓 What?

- Base structure for ADF connector
- Get table lineage from `dataflow`
- Dump metadata for developing other integration, such as copy data from salesforce to a dataset.

### 🧪 Tested?

Unit tests and test against an ADF instance.


mars-lan (approved):
Great work

PR #582, opened 2023-08-30
Get data platform type directly from Monte Carlo GraphQL

mars-lan:
### 🤔 Why?

The current implementation assumes that there's only one data platform and asks the user to explicitly specify the value in the config.

### 🤓 What?

Use the [getTables](https://apidocs.getmontecarlo.com/#query-getTables) GraphQL query to get the platform information for each table and set the corresponding data platform when building the MCEs.

> Note: Based on MC's [API doc](https://apidocs.getmontecarlo.com/), it doesn't seem possible to retrieve the Snowflake account information. As such, we still need the user to manually specify it in the config.

### 🧪 Tested?

Verified the before & after against a production instance and saw no difference.


alyiwang (approved):
LGTM

PR #584, opened 2023-09-06
Add structure field in thoughtspot crawler [sc-20609]

alyiwang:
### 🤔 Why?

To support standardized asset hierarchy

### 🤓 What?

- add `structure` field in thoughtspot dashboard and dataset, using the `type` as hierarchy

### 🧪 Tested?

tested against server, and verified the MCE


poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.5.1 and should not be changed by hand.
```
elic-eon, line 1: Upgrade local `poetry` to the latest?

elic-eon (approved):
LGTM

poetry.lock
alyiwang: Thanks for catching. Updated in the next PR for power BI

PR #591, opened 2023-09-11
Support fetching metadata from multiple dbt cloud jobs

mars-lan:
### 🤔 Why?

Setting up separate connectors for multiple dbt jobs quickly becomes a maintenance headache. Given how quickly it is to fetch metadata for a single dbt job, supporting multiple jobs in a single connector seems like an obvious extension.

### 🤓 What?

- Add `job_ids` config and deprecate `job_id`
- Update connector to support multiple job IDs
- Move `DbtAdminAPIClient` into a separate file and add tests

### 🧪 Tested?


alyiwang (approved):
LGTM

PR #601, opened 2023-09-20
Get Power BI dataflow metadata

elic-eon:
### 🤔 Why?

To extract Power BI Dataflow metadata, and lineage created by the dataflow

### 🤓 What?

- Model PBI Dataflow as Metaphor pipeline
- Parse table lineage from dataflow definition

### 🧪 Tested?

Ingested MCE generated to dev.

Unit tests


mars-lan (approved):
Great work

PR #604, opened 2023-09-21
Safely parse power query for dataflow

elic-eon:
### 🤔 Why?

If the source of power query is not a database, the crawler will throw error.

### 🤓 What?

- Catch the parse power query error

### 🧪 Tested?

Verified the crawler can run all workspace under the tenant.


metaphor/power_bi/power_bi_client.py
```diff
@@ -154,7 +154,7 @@ class WorkspaceInfoDataflow(BaseModel):
     configuredBy: Optional[str] = None
     modifiedBy: Optional[str] = None
     modifiedDateTime: Optional[str] = None
-    refreshSchedule: Optional[PowerBiRefreshSchedule]
+    refreshSchedule: Optional[PowerBiRefreshSchedule] = None
```
mars-lan, line 157: I believe you don't need `Optional` if there's a default value already?

elic-eon: But I need the default value to be `None`, the type has to be `Optional[T]` in this case.

mars-lan: Argh you're right. I got confused.

mars-lan (approved):

PR #608, opened 2023-09-28
Fix Power BI dataflow model errors

elic-eon:
### 🤔 Why?

The name and description could be null or undefined.

### 🤓 What?

- Make field optional with default `None`

### 🧪 Tested?

Unit test added.

Verified against production environment.

### DIFF
--- a/metaphor/power_bi/power_bi_client.py
+++ b/metaphor/power_bi/power_bi_client.py
@@ -161,8 +161,8 @@ class PowerBiRefreshSchedule(BaseModel):
 
 class WorkspaceInfoDataflow(BaseModel):
     objectId: str
-    name: str
-    description: str
+    name: Optional[str] = None
+    description: Optional[str] = None
     configuredBy: Optional[str] = None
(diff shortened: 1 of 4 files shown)


metaphor/power_bi/power_bi_client.py
```diff
@@ -439,12 +438,6 @@ def wait_for_scan_result(scan_id: str, max_timeout_in_secs: int = 30) -> bool:
             return False
 
         def transform_scan_result(response: requests.Response) -> dict:
-            # Write output to file to help debug issues
```
mars-lan, line 442: Why dropping this?

elic-eon: To make it testable, and the same content also be dumped in the `get_request`

mars-lan: Got it.

mars-lan (dismissed):

mars-lan (approved):

PR #611, opened 2023-10-02
improve MCE validator performance

alyiwang:
### 🤔 Why?

The way we use `jsonschema` validator is the default usage which loads the SCHEMA file everytime. This is a huge waste of time if we need to validate a large amount of records. A better way is to instantiate the validator and use it repeatedly. 
Some performance analysis https://www.peterbe.com/plog/jsonschema-validate-10x-faster-in-python

### 🤓 What?

-  instantiate jsonschema validator

### 🧪 Tested?

Tested locally, json validation of ~300 records took 1.17s before vs 0.06s after the change

### DIFF
--- a/metaphor/common/event_util.py
+++ b/metaphor/common/event_util.py
@@ -3,7 +3,8 @@ import logging
 from importlib import resources
 from typing import Union
 
-from jsonschema import ValidationError, validate
+from jsonschema import ValidationError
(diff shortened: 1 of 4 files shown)


mars-lan (approved):
Thanks for the quick fix!

mars-lan:
Forgot to bump version?

PR #615, opened 2023-10-04
Fix dbt connector by removing undocumented fields

mars-lan:
### 🤔 Why?

dbt 1.6.5 introduced new fields that are not covered by the [existing manifest JSON schema](https://schemas.getdbt.com/dbt/manifest/v10/index.html). See https://github.com/dbt-labs/dbt-core/issues/8763 for more details.

### 🤓 What?

Strip off the extra fields for now until https://github.com/dbt-labs/dbt-core/issues/8763 is fixed.

> Note: Attempted to solve the issue by using the `--allow-extra-fields`(https://github.com/koxudaxi/datamodel-code-generator#:~:text=%2D%2Dallow%2Dextra%2Dfields) flag but for some unknown reasons, the generated Pydantic model can't parse dbt tests properly. It might be a Pydantic 1.x bug. 

### 🧪 Tested?

Verified against a production dbt cloud instance.


metaphor/common/utils.py
```diff
@@ -148,3 +148,11 @@ def removesuffix(text: str, suffix: str):
...
     else:
         return text
+
+
+def safe_del(dict: Dict, key: str) -> Dict:
+    """Safely remove a key from a dict even if the key doesn't exist"""
+    if key in dict:
+        del dict[key]
+
+    return dict
```
alyiwang, line 158: Is this the same as `dict.pop(key, None)` ?

mars-lan: lol good point.

alyiwang (approved):
LGTM

PR #617, opened 2023-10-06
Power bi crawler should use new MCE model

elic-eon:
### 🤔 Why?

A breaking change of `metaphor-models` was introduced. Need to use new MCE model

### 🤓 What?

- Extract PBI workspace as hierarchy
- Use latest model

### 🧪 Tested?

Unit tests are updated.

Ingest MCE and verified end-to-end


pyproject.toml
mars-lan, line 1: Also bump up version?

alyiwang (approved):
LGTM

PR #618, opened 2023-10-09
Fix dbt parsing errors due to invalid values in manifest

mars-lan:

### DIFF
--- a/metaphor/common/utils.py
+++ b/metaphor/common/utils.py
@@ -148,3 +148,8 @@ def removesuffix(text: str, suffix: str):
         return text[: -len(suffix)]
     else:
         return text
+
+
+def filter_none(lst: List) -> List:
+    """Filter out all None values from the list"""
+    return list(filter(lambda e: e is not None, lst))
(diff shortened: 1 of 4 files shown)


metaphor/common/utils.py
```diff
@@ -148,3 +148,8 @@ def removesuffix(text: str, suffix: str):
         return text[: -len(suffix)]
     else:
         return text
+
+
+def filter_none(lst: List) -> List:
+    """Filter out all None values from the list"""
+    return list(filter(lambda e: e is not None, lst))
```
alyiwang, line 155: nit: a slightly more pythonic way is list comprehension
`[e for e in lst if e is not None]`.
This is fine too.

alyiwang (dismissed):
LGTM
Thanks for fixing!

metaphor/common/utils.py
mars-lan: Good point.

alyiwang (approved):

PR #627, opened 2023-10-13
Fix dbt 1.6.6 manifest schema validation errors

mars-lan:
### 🤔 Why?

dbt 1.6.6 introduced changes that are not compatible with the latest manifest JSON schema: https://github.com/dbt-labs/dbt-core/issues/8851

### 🤓 What?

Strip off the extra fields in metrics as a temporary work around. Also add tests for all current workarounds.

### 🧪 Tested?


alyiwang (approved):
Thanks for adding tests

PR #634, opened 2023-10-17
Fix Unity Catalog parsing erros due to permission issues

mars-lan:
### 🤔 Why?

The Unity Catalog API returns a different (and undocumented) data model when the user has [no permission](https://docs.databricks.com/en/data-governance/unity-catalog/data-lineage.html#lineage-permissions) to select the upstream tables. This led to errors like these:

```
2023-10-16 20:57:32:ERROR:metaphor:3 validation errors for ParsingModel[TableLineage]
__root__ -> upstreams -> 0 -> tableInfo -> name
  field required (type=value_error.missing)
__root__ -> upstreams -> 0 -> tableInfo -> catalog_name
  field required (type=value_error.missing)
__root__ -> upstreams -> 0 -> tableInfo -> schema_name
  field required (type=value_error.missing)
```

### 🤓 What?

- Update the models to handle the cases when permission is missing
- Update the models to handle file-based upstream/downstream
- Include all API responses as JSON files in the output zip file

### 🧪 Tested?


elic-eon (approved):
LGTM, but CI fails, might need add more tests?

PR #636, opened 2023-10-18
Extract refresh schedule

elic-eon:
### 🤔 Why?

Get refresh schedule information for PowerBI Dataset

### 🤓 What?

- Extract modeled dataset using [get-refresh-schedule](https://learn.microsoft.com/en-us/rest/api/power-bi/datasets/get-refresh-schedule)
- Extract direct query dataset using [get-direct-query-refresh-schedule](https://learn.microsoft.com/en-us/rest/api/power-bi/datasets/get-direct-query-refresh-schedule-in-group)

### 🧪 Tested?

Test the crawler against testing and production environment.


mars-lan:
Test failed?

metaphor/power_bi/power_bi_client.py
```diff
@@ -367,11 +368,61 @@ def get_refreshes(self, group_id: str, dataset_id: str) -> List[PowerBIRefresh]:
         except Exception as e:
             # Fail gracefully for any other errors
             logger.error(
-                f"Failed to get pages from dataset {dataset_id} in workspace {group_id}: {e}"
+                f"Failed to get refreshes from dataset {dataset_id} in workspace {group_id}: {e}"
             )
 
         return []
 
+    def get_refresh_schedule(
```
mars-lan, line 376: Is this subjected to the 200 calls/hr API limit? If so it'll likely exceed when there are more than 200 datasets?

mars-lan (approved):

PR #638, opened 2023-10-19
Get Power BI dataflow last refresh

elic-eon:
### 🤔 Why?

Get last refresh like we did for PBI dataset

### 🤓 What?

Get last success refresh time from the API.

### 🧪 Tested?

Verified the output MCE

### DIFF
--- a/metaphor/power_bi/extractor.py
+++ b/metaphor/power_bi/extractor.py
@@ -61,6 +61,7 @@ from metaphor.models.metadata_change_event import (
 )
 from metaphor.power_bi.config import PowerBIRunConfig
 from metaphor.power_bi.power_bi_client import (
+    DataflowTransaction,
     PowerBIApp,
     PowerBIClient,
     PowerBIPage,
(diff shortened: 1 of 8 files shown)


metaphor/power_bi/extractor.py
```diff
@@ -737,6 +745,24 @@ def _find_last_completed_refresh(
 
         return safe_parse_ISO8601(refresh.endTime)
 
+    @staticmethod
+    def _find_refresh_time_from_transaction(
```
mars-lan, line 749: We seem to have accumulated quite a few of these static util methods. Consider moving them to a separate file to improve readability & testability?

mars-lan: Let's do that in a follow-up PR.

mars-lan (commented):
Bump up versnion?

alyiwang (approved):
LGTM

metaphor/power_bi/extractor.py
elic-eon: SG

alyiwang (approved):

mars-lan (approved):

PR #640, opened 2023-10-19
Include table-level lineage when column-level lineage is missing

mars-lan:
### 🤔 Why?

Turns out that sometimes Unity Catalog will return table lineage (via `/lineage-tracking/table-lineage` endpoint) but not column lineage (via `/lineage-tracking/column-lineage`). The current implementation relies on the latter as a source for both types of lineage and will therefore return no lineage in this case.

### 🤓 What?

Use table lineage to build source datasets and allow column lineage to be "decorated" if exists.

### 🧪 Tested?


alyiwang (approved):
LGTM

PR #642, opened 2023-10-21
Move UC-specific metadata from CustomMetadata to dedicated fields

mars-lan:
### 🤓 What?

As titled. Also fill out the `source_info` field to point the dataset to the corresponding page on Databircks.

### 🧪 Tested?


elic-eon (approved):
LGTM!

PR #643, opened 2023-10-24
Upgrade Pydantic to v2

usefulalgorithm:
### 🤔 Why?

[SC19184](https://app.shortcut.com/metaphor-data/story/19184)

### 🤓 What?

- Ran `bump_pydantic` migration tool
- `root_validator` -> `model_validator`
  - `model_validator` is run *after* the object is instantiated - I tried validating the thing before it's a proper object, but if the object is created with kwargs (e.g. `Foo(bar=42, baz="quax")`) then in the validator function it will be a `pydantic_core._pydantic_core.ArgsKwargs`, which does not behave like a `dict` and doesn't have the `get` method. I figured `self.__dict__` should be sufficient, since we're only checking that only one in the supported optional fields is not a `None`.
- `parse_raw_as` -> `TypeAdapter(cls).validate_python(...)`

### 🧪 Tested?

- All unit tests are passing on my machine.
- Tested with snowflake by running `metaphor snowflake SOME_CONFIG.yml`, where the config file was downloaded from metaphor-config-stage bucket, with the output dir pointing to a directory in my local environment. The output directory was successfully populated with data.


tests/fivetran/data/v1__connectors__connector_1.json
```diff
@@ -18,7 +18,7 @@
       "setup_state": "connected",
       "sync_state": "scheduled",
       "update_state": "on_schedule",
-      "is_historical_sync": false,
+      "is_historical_sync": "False",
```
mars-lan, line 21: This looks wrong? Maybe the migration script made a mistake?

usefulalgorithm: no it's actually correct, since in `metaphor/fivetran/models.py` the class is defined as
```python
class ConnectorStatus(BaseModel):
    setup_state: str
    update_state: str
    sync_state: str
    is_historical_sync: str # <-------- This
```
Don't know why it worked in V1 though, maybe some implicit conversion happened.

mars-lan: Oh interesting. Was this intentional, @elic-eon?

usefulalgorithm: Actually maybe this is like the `coerce_numbers_to_str` stuff. I'll double check.

usefulalgorithm: I changed it back.

pyproject.toml
```diff
@@ -33,8 +33,8 @@ metaphor-models = "0.27.9"
 msal = { version = "^1.20.0", optional = true }
 oscrypto = { git = "https://github.com/wbond/oscrypto.git", rev = "1547f53" } # Until oscrypto 1.3.1 is release: https://github.com/wbond/oscrypto/issues/78
 pycarlo = { version = "^0.8.1", optional = true }
-pydantic = { version = "^1.10.0", extras = ["email"]}
-pymssql = { version = "2.2.7", optional = true } # pymssql 2.2.8 is currently broken
+pydantic = { version = "^2.4.2", extras = ["email"]}
+pymssql = { version = "^2.2.9", optional = true }
```
usefulalgorithm, line 37: Couldn't install 2.2.7 on my machine, and since 2.2.9 is out I decided to just update to that version. Is there any other reason to use 2.2.7 aside from 2.2.8 being broken?

elic-eon: Know the reason you can't install 2.2.7? I remember there being a dependency issue. Need to manually install some library from brew

mars-lan: See https://github.com/MetaphorData/connectors/blob/main/docs/apple_silicon.md

mars-lan: Additional discussion: https://github.com/pymssql/pymssql/issues/763

usefulalgorithm: It just fails to compile, see the whole output [here](https://pastebin.com/h2j8nfAZ). I have exported the compilation flags but it still does not work. 5 minutes of googling did not get me the solution to this.

Is it ok to bump it to 2.2.9?

elic-eon: I think it's fine since it's a patch version change. wdyt @mars-lan

mars-lan: Please verify against production and update https://github.com/MetaphorData/connectors/blob/main/docs/apple_silicon.md if the instruction is no longer needed on Mac.

usefulalgorithm: I ran `metaphor synapse config.yaml` and it looks ok.

metaphor/bigquery/config.py
```diff
@@ -83,7 +83,7 @@ class BigQueryRunConfig(BaseConfig):
...
     query_log: BigQueryQueryLogConfig = BigQueryQueryLogConfig()
 
-    @root_validator
-    def have_key_path_or_credentials(cls, values):
-        must_set_exactly_one(values, ["key_path", "credentials"])
-        return values
+    @model_validator(mode="after")
+    def have_key_path_or_credentials(self) -> "BigQueryRunConfig":
+        must_set_exactly_one(self.__dict__, ["key_path", "credentials"])
+        return self
```
elic-eon, line 89: Are these changes modified by the migration tool?

metaphor/dbt/generated/dbt_manifest_v10.py
```diff
@@ -1392,7 +1330,7 @@ class Config:
         'operation',
         'seed',
         'rpc',
-        'sql_operation',
+        'sql operation',
```
elic-eon, line 1333: Shouldn't we change this?

metaphor/snowflake/lineage/extractor.py
```diff
@@ -158,7 +158,8 @@ def _parse_access_log(
         source_datasets = []
 
         # Extract source tables/views
-        source_objects = parse_raw_as(List[AccessedObject], objects_accessed)
+        ta = TypeAdapter(List[AccessedObject])
+        source_objects = ta.validate_json(objects_accessed)
```
elic-eon, line 162: Does `validate_json` parse bytes or str to an object?

metaphor/unity_catalog/models.py
```diff
@@ -1,7 +1,7 @@
 from enum import Enum
 from typing import List, Optional, Union
 
-from pydantic import BaseModel, parse_obj_as, validator
+from pydantic import BaseModel, field_validator, parse_obj_as
```
elic-eon, line 4: Still need `parse_obj_as`?

mars-lan:
Please also sign your commits: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

mars-lan:
Please make sure to sign your commits: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#ssh-commit-signature-verification

metaphor/dbt/generated/dbt_manifest_v10.py
usefulalgorithm: orz I missed this one, thanks for catching this

metaphor/snowflake/lineage/extractor.py
usefulalgorithm: `str`.

metaphor/bigquery/config.py
usefulalgorithm: No I did those manually. The old `root_validator` is deprecated, and it needs additional arguments to work. I decided to use the new `model_validator`, and this is how it's supposed to be used.

metaphor/unity_catalog/models.py
usefulalgorithm: it's deprecated but not yet removed. I should remove it though.

metaphor/bigquery/config.py
usefulalgorithm: I don't think so, there are two ways to stick with the original syntax but I don't think they're any good
1. `model_validator(mode="before")` -> can keep the method as a classmethod, but doesn't really work. See PR description: 
  > I tried validating the thing before it's a proper object, but if the object is created with kwargs (e.g. Foo(bar=42, baz="quax")) then in the validator function it will be a pydantic_core._pydantic_core.ArgsKwargs, which does not behave like a dict and doesn't have the get method. I figured self.__dict__ should be sufficient, since we're only checking that only one in the supported optional fields is not a None.
2. `root_validator` -> looks like it's deprecated, from pydantic source code:
```python
raise PydanticUserError(
    'If you use `@root_validator` with pre=False (the default) you MUST specify `skip_on_failure=True`.'
    ' Note that `@root_validator` is deprecated and should be replaced with `@model_validator`.',
    code='root-validator-pre-skip',
)
```
    We didn't specify `skip_on_failure=True` before, I decided since it's deprecated might as well drop it altogether.

usefulalgorithm: tl;dr: yes, but probably best not to

metaphor/thought_spot/config.py
```diff
@@ -17,8 +17,8 @@ class ThoughtSpotRunConfig(BaseConfig):
     secret_key: Optional[str] = None
     password: Optional[str] = None
 
-    @validator("password")
-    def check_password_or_secret_key(cls, password, values):
-        if "secret_key" not in values and not password:
+    @model_validator(mode="after")
+    def check_password_or_secret_key(self) -> "ThoughtSpotRunConfig":
+        if self.password is None and self.secret_key is None:
```
mars-lan, line 22: Replace with `must_set_exactly_one`?

metaphor/snowflake/lineage/extractor.py
```diff
@@ -158,7 +158,8 @@ def _parse_access_log(
         source_datasets = []
 
         # Extract source tables/views
-        source_objects = parse_raw_as(List[AccessedObject], objects_accessed)
+        ta = TypeAdapter(List[AccessedObject])
```
mars-lan, line 161: ditto

metaphor/snowflake/lineage/extractor.py
```diff
@@ -177,7 +178,8 @@ def _parse_access_log(
             return
 
         # Assign source tables as upstream of each destination tables
-        target_objects = parse_raw_as(List[AccessedObject], objects_modified)
+        ta = TypeAdapter(List[AccessedObject])
```
mars-lan, line 181: ditto

metaphor/snowflake/extractor.py
```diff
@@ -599,7 +599,8 @@ def build_table_url(account: str, full_name: str) -> str:
         )
 
     def _parse_accessed_objects(self, raw_objects: str) -> List[QueriedDataset]:
-        objects = parse_raw_as(List[AccessedObject], raw_objects)
+        ta = TypeAdapter(List[AccessedObject])
```
mars-lan, line 602: ditto

metaphor/fivetran/models.py
```diff
@@ -59,10 +58,11 @@ class DestinationPayload(BaseModel):
     region: str
     time_zone_offset: str
     setup_status: str
-    config: Any
+    config: Any = None
 
 
 class ConnectorStatus(BaseModel):
+    model_config = ConfigDict(coerce_numbers_to_str=True)
```
mars-lan, line 65: Since this is likely to be a default for all models, maybe create our own version of `BaseModel` to extend from?

metaphor/bigquery/config.py
usefulalgorithm: yes

metaphor/fivetran/models.py
```diff
@@ -1,30 +1,29 @@
...
-from pydantic.generics import GenericModel
+from pydantic import BaseModel, ConfigDict, Field
 
 DataT = TypeVar("DataT")
 
 
-class GenericData(GenericModel, Generic[DataT]):
-    next_cursor: Optional[str]
+class GenericData(BaseModel, Generic[DataT]):
+    next_cursor: Optional[str] = None
```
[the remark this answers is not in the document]
usefulalgorithm: this is automatically done by the migration tool. I figured since we always create pydantic objects with keyword arguments this probably doesn't matter?

tests/redshift/test_utils.py
```diff
@@ -4,14 +4,14 @@
 
 def test_exclude_system_databases():
     filter = DatasetFilter(
-        includes={"db1": []},
-        excludes={"db2": {"schema2": []}},
+        includes={"db1": {}},
```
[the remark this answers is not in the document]
usefulalgorithm: maybe...

metaphor/thought_spot/config.py
usefulalgorithm: I think the original logic is to allow setting both fields?

mars-lan: Argh got it. Maybe add a `must_set_all` util method? :P

metaphor/fivetran/models.py
usefulalgorithm: Good call

metaphor/common/git.py
```diff
@@ -28,10 +28,10 @@ class GitRepoConfig:
     # relative path to the project, default to the root of the repo
     project_path: str = ""
 
-    @root_validator
-    def have_token_or_password(cls, values):
-        must_set_exactly_one(values, ["access_token", "password"])
-        return values
+    @model_validator(mode="after")
+    def have_token_or_password(self) -> "GitRepoConfig":
```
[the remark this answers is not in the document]
usefulalgorithm: We can do `from __future__ import annotation` and just use the class symbol, other than that I don't really know any other way

metaphor/thought_spot/config.py
usefulalgorithm: Yeah I can do that, don't know if there's any other place with this validation logic though

metaphor/common/git.py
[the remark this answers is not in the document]
mars-lan: Is it available in 3.8? https://peps.python.org/pep-0563/

metaphor/thought_spot/config.py
mars-lan: Never mind. Let's do that in a future PR.

metaphor/fivetran/models.py
usefulalgorithm: I didn't modify the base models with ok types (e.g. `id` is an `int), nor did I change the dbt ones - I'm guessing they're generated with datamodel-codegen?

metaphor/thought_spot/config.py
usefulalgorithm: `must_set_at_least_one`

metaphor/common/models.py
```diff
@@ -44,3 +45,12 @@ def to_dataset_statistics(
         record_count=safe_float(rows),
         last_updated=last_updated,
     )
+
+
+class V1CompatBaseModel(BaseModel):
+    """
+    A Pydantic model that preserves the behavior in Pydantic v1, i.e.
+    numbers and `bool`s are parsed to `str`s by default.
```
mars-lan, line 53: Bools are also coerced into strings? 😱

metaphor/common/git.py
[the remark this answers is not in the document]
usefulalgorithm: https://docs.python.org/3/library/__future__.html#id1

It becomes available without additional setup in python 3.11

metaphor/snowflake/extractor.py
```diff
@@ -599,7 +599,7 @@ def build_table_url(account: str, full_name: str) -> str:
         )
 
     def _parse_accessed_objects(self, raw_objects: str) -> List[QueriedDataset]:
-        objects = parse_raw_as(List[AccessedObject], raw_objects)
+        objects = TypeAdapter(List[AccessedObject]).validate_json(raw_objects)
```
mars-lan, line 602: On second thought, this `TypeAdapter(<type>).validate_json` is so prevalent, maybe we should add an util method?

metaphor/thought_spot/config.py
```diff
@@ -17,8 +18,7 @@ class ThoughtSpotRunConfig(BaseConfig):
...
     password: Optional[str] = None
 
-    @validator("password")
-    def check_password_or_secret_key(cls, password, values):
-        if "secret_key" not in values and not password:
-            raise ValueError("Either password or secret_key is required")
-        return password
+    @model_validator(mode="after")
+    def check_password_or_secret_key(self) -> "ThoughtSpotRunConfig":
+        must_set_at_least_one(self.__dict__, ["secret_key", "password"])
```
mars-lan, line 23: Actually the semantic should have been `must_set_exactly_one`, since setting both or none is not acceptable?

metaphor/common/models.py
usefulalgorithm: yeah, in Python 3 bools are actually integers: https://docs.python.org/3/reference/datamodel.html#index-10

metaphor/thought_spot/config.py
usefulalgorithm: I think it should be `must_set_at_least_one`? At least according to the original error message it looks that way
> "Either password or secret_key is required"

I don't know which one takes precedence though

mars-lan: Yeah it's not super clear from the code: https://github.com/MetaphorData/connectors/blob/53807630c18575e392cd0f340b774d6a42545c0b/metaphor/thought_spot/utils.py#L98

But I believe `secret_key` takes precedence over `password`. @elic-eon can confirm.

metaphor/common/models.py
mars-lan: 🤯 Does that mean `false` in YAML will get converted into `"0"` then back to `False`

metaphor/thought_spot/config.py
usefulalgorithm: anyway, I think this should be `must_set_at_least_one`? Looks like thoughtspot api takes both of them just fine:
https://github.com/thoughtspot/thoughtspot_rest_api_v1_python/blob/main/src/thoughtspot_rest_api_v1/tsrestapiv2.py#L124

metaphor/common/models.py
usefulalgorithm: I think it's more like whenever python interpreter see a `False`, it thinks `0`.

metaphor/snowflake/extractor.py
usefulalgorithm: I think this is good enough for now ~ I used `TypeAdapter` to both validate_json and validate_python, don't think adding an util to wrap over `TypeAdapter` would make it easier to understand

mars-lan: SGTM

metaphor/thought_spot/config.py
elic-eon: https://developers.thoughtspot.com/docs/api-authv2#trusted-auth-v2 Can't tell from this doc. The original logic is the same as `must_set_at_least_one`

usefulalgorithm:
todo:
1. gen_model.sh in dbt directory should be run with pydantic v2 models
2. add tests for [metaphor/unity_catalog/utils.py](https://app.codecov.io/gh/MetaphorData/connectors/pull/643/blob/metaphor/unity_catalog/utils.py) so that patch line coverage >= 80%

metaphor/common/models.py
```diff
@@ -44,3 +45,12 @@ def to_dataset_statistics(
         record_count=safe_float(rows),
         last_updated=last_updated,
     )
+
+
+class V1CompatBaseModel(BaseModel):
```
mars-lan, line 50: Please add a TODO & ticket to drop this eventually.

metaphor/thought_spot/config.py
mars-lan: SGTM

metaphor/common/models.py
usefulalgorithm: ok

usefulalgorithm: https://app.shortcut.com/metaphor-data/story/21948/drop-v1compatbasemodel-from-crawlers

pyproject.toml
elic-eon, line 1: Please bump connector version

elic-eon (commented):
Otherwise, LGTM. Great refactoring

elic-eon:
Resolve merge conflicts?

elic-eon (approved):

mars-lan:
Great work!

PR #644, opened 2023-10-25
Update snowflake system tags with hierarchy

usefulalgorithm:
### 🤔 Why?

[SC21047](https://app.shortcut.com/metaphor-data/story/21047/)

### 🤓 What?

In `_fetch_tags`, when a database-level or schema-level tag is found, it will be included in the MCE as a part of a `Hierarchy`, which has a `HierarchyLogicalID` that starts with `DataPlatform.SNOWFLAKE.value`.

I will modify this PR once #643 is merged. 

Need to create story for frontend work I think.

### 🧪 Tested?

Wrote unit test for this. Tested against local dev environment.

Also tested against modified snowflake env, added the following system tags:
截圖 2023-10-26 晚上11 33 51
After running the crawler, we will have the following hierarchies in the output json:
[res.json](https://github.com/MetaphorData/connectors/files/13179681/res.json)

### DIFF
--- a/metaphor/snowflake/auth.py
+++ b/metaphor/snowflake/auth.py
@@ -34,10 +34,10 @@ class SnowflakeKeyPairAuthConfig:
     # provide decryption passphrase if private key is encrypted
     passphrase: Optional[str] = None
 
-    @model_validator(mode="after")
-    def have_key_file_or_key_content(self):
-        must_set_exactly_one(self.__dict__, ["key_file", "key_data"])
-        return self
+    @model_validator(mode="before")
+    def have_key_file_or_key_content(cls, values):
+        must_set_exactly_one(values, ["key_file", "key_data"])
+        return values
 
 
 @dataclass(config=ConnectorConfig)
@@ -63,10 +63,10 @@ class SnowflakeAuthConfig(BaseConfig):
     # the query tags for each snowflake query the connector issues
     query_tag: Optional[str] = METAPHOR_DATA
 
-    @model_validator(mode="after")
-    def have_password_or_private_key(self):
-        must_set_exactly_one(self.__dict__, ["password", "private_key"])
-        return self
+    @model_validator(mode="before")
+    def have_password_or_private_key(cls, values):
+        must_set_exactly_one(values, ["password", "private_key"])
+        return values
 
 
 def connect(config: SnowflakeAuthConfig) -> snowflake.connector.SnowflakeConnection:
--- a/metaphor/snowflake/auth.py
+++ b/metaphor/snowflake/auth.py
@@ -34,10 +34,10 @@ class SnowflakeKeyPairAuthConfig:
     # provide decryption passphrase if private key is encrypted
     passphrase: Optional[str] = None
 
-    @model_validator(mode="before")
-    def have_key_file_or_key_content(cls, values):
-        must_set_exactly_one(values, ["key_file", "key_data"])
-        return values
+    @model_validator(mode="after")
+    def have_key_file_or_key_content(self):
+        must_set_exactly_one(self.__dict__, ["key_file", "key_data"])
+        return self
 
 
 @dataclass(config=ConnectorConfig)
@@ -63,10 +63,10 @@ class SnowflakeAuthConfig(BaseConfig):
     # the query tags for each snowflake query the connector issues
     query_tag: Optional[str] = METAPHOR_DATA
 
-    @model_validator(mode="before")
-    def have_password_or_private_key(cls, values):
-        must_set_exactly_one(values, ["password", "private_key"])
-        return values
+    @model_validator(mode="after")
+    def have_password_or_private_key(self):
+        must_set_exactly_one(self.__dict__, ["password", "private_key"])
+        return self
 
 
 def connect(config: SnowflakeAuthConfig) -> snowflake.connector.SnowflakeConnection:
--- a/metaphor/snowflake/extractor.py
+++ b/metaphor/snowflake/extractor.py
@@ -140,7 +140,7 @@ class SnowflakeExtractor(BaseExtractor):
         entities: List[ENTITY_TYPES] = []
         entities.extend(datasets)
         entities.extend(chunk_query_logs(self._logs))
-        entities.extend(list(self._hierarchies.values()))
+        entities.extend(self._hierarchies.values())
         return entities
 
     @staticmethod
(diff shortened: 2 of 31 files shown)


metaphor/snowflake/auth.py
```diff
@@ -34,7 +34,7 @@ class SnowflakeKeyPairAuthConfig:
     # provide decryption passphrase if private key is encrypted
     passphrase: Optional[str] = None
 
-    @root_validator
+    @root_validator(pre=False, skip_on_failure=True)
```
usefulalgorithm, line 37: no

metaphor/snowflake/auth.py
```diff
@@ -63,7 +63,7 @@ class SnowflakeAuthConfig(BaseConfig):
     # the query tags for each snowflake query the connector issues
     query_tag: Optional[str] = METAPHOR_DATA
 
-    @root_validator
+    @root_validator(pre=False, skip_on_failure=True)
```
usefulalgorithm, line 66: no

metaphor/snowflake/extractor.py
```diff
@@ -599,7 +616,7 @@ def build_table_url(account: str, full_name: str) -> str:
         )
 
     def _parse_accessed_objects(self, raw_objects: str) -> List[QueriedDataset]:
-        objects = parse_raw_as(List[AccessedObject], raw_objects)
+        objects = parse_raw_as(List[AccessedObject], raw_objects)  # type: ignore
```
usefulalgorithm, line 619: no

metaphor/snowflake/auth.py
perepelitca: ? 🙂

usefulalgorithm: ah sorry, I put this here just to remind myself to get rid of this once pydantic v2 PR is merged - in v2 `root_validator` is deprecated, instead it's going to be `model_validator`

mars-lan: Another approach is to base your PR directly on https://github.com/MetaphorData/connectors/tree/tsung-julii/sc-19184/update-to-pydantic-2-x

elic-eon:
There are another PR changes, can you separate them?

usefulalgorithm:
@elic-eon  @mars-lan  @perepelitca  this is ready for review~

metaphor/snowflake/extractor.py
```diff
@@ -135,6 +140,7 @@ async def extract(self) -> Collection[ENTITY_TYPES]:
         entities: List[ENTITY_TYPES] = []
         entities.extend(datasets)
         entities.extend(chunk_query_logs(self._logs))
+        entities.extend(list(self._hierarchies.values()))
```
elic-eon, line 143: Update tests to cover new entities

metaphor/snowflake/extractor.py
```diff
@@ -359,45 +365,92 @@ def _fetch_primary_keys(self, cursor: SnowflakeCursor) -> None:
...
             SELECT TAG_NAME, TAG_VALUE, DOMAIN, OBJECT_DATABASE, OBJECT_SCHEMA, OBJECT_NAME, COLUMN_NAME
             FROM snowflake.account_usage.tag_references
-            WHERE DOMAIN in ('TABLE', 'COLUMN')
+            WHERE DOMAIN in ('TABLE', 'COLUMN', 'DATABASE', 'SCHEMA')
             ORDER BY OBJECT_DATABASE, OBJECT_SCHEMA, OBJECT_NAME
             """
         )
 
-        for key, value, object_type, database, schema, table_name, column in cursor:
+        for key, value, object_type, database, schema, object_name, column in cursor:
```
elic-eon, line 420: What `object_name` would be for schema/database level tag?

metaphor/snowflake/extractor.py
```diff
@@ -359,45 +365,92 @@ def _fetch_primary_keys(self, cursor: SnowflakeCursor) -> None:
...
+        if not object_name:
+            logger.error("Cannot extract tag without object name")
+            return
+
+        # Set hierarchy key and logical_id
+        path = [DataPlatform.SNOWFLAKE.value]
+        if object_type == "DATABASE":
+            path.extend([object_name])
+            hierarchy_key = dataset_normalized_name(object_name)
+        elif object_name is not None:
```
elic-eon, line 391: `object_name` should be truthy here

metaphor/snowflake/extractor.py
```diff
@@ -359,45 +365,92 @@ def _fetch_primary_keys(self, cursor: SnowflakeCursor) -> None:
                 sql_schema.primary_key = []
             sql_schema.primary_key.append(column)
 
+    def _add_system_tag(
+        self,
+        database: Optional[str],
+        object_name: Optional[str],
+        object_type: str,
```
elic-eon, line 372: Maybe we can narrow down `object_type` to `database|schema`?

metaphor/snowflake/extractor.py
```diff
@@ -359,45 +365,92 @@ def _fetch_primary_keys(self, cursor: SnowflakeCursor) -> None:
...
+            )
+
+        # `system_tags` is never None here
+        self._hierarchies[hierarchy_key].system_tags.append(
+            SystemTag(
+                key=tag_key,
+                system_tag_source=SystemTagSource.SNOWFLAKE,
+                value=tag_value,
+            )
+        )
```
elic-eon, line 408: Maybe you can use [dict.setdefault](https://docs.python.org/3/library/stdtypes.html?highlight=setdefault#dict.setdefault) to set the default value and get the reference to append the system tag.

metaphor/snowflake/extractor.py
usefulalgorithm: I don't think the existing tests ever run `extractor.extract`, so none of these entities are actually covered. I'm currently having trouble on this...

metaphor/snowflake/extractor.py
usefulalgorithm: for schema level tags it's schema name, and vice versa

metaphor/snowflake/extractor.py
elic-eon: NVM, the test you added should be good enough

poetry.lock
elic-eon, line 1: Normally, I use `poetry lock --no-update` and `poetry install -E all` to update the package. It won't update all dependencies every time.

usefulalgorithm: got it

metaphor/snowflake/extractor.py
elic-eon, line 143: I think `extend` can take `iterable`?

elic-eon (approved):
LGTM!

PR #647, opened 2023-10-26
Add config to filter out large queries from Snowflake query logs

mars-lan:
### 🤔 Why?

Sometimes there can be queries that insert a huge amount of raw data that should be ignored.

### 🤓 What?

As titled.

### 🧪 Tested?


elic-eon (approved):
LGTM!

PR #650, opened 2023-10-29
Extract column description from Unity Catalog tables

mars-lan:
### 🤔 Why?

The connector currently only extracts table-level descriptions but not column-level ones.

### 🤓 What?

As titled.

### 🧪 Tested?


usefulalgorithm (approved):
LGTM!

PR #651, opened 2023-10-30
Get query logs from unity catalog

usefulalgorithm:
### 🤔 Why?

- To get query logs and relevant information from databricks system.
- Use newer databricks SDK.
(SC links: [SC21951](https://app.shortcut.com/metaphor-data/story/21951), [SC21623](https://app.shortcut.com/metaphor-data/story/21623))

### 🤓 What?

- updated to use new sdk.
- get query logs with `WorkspaceClient.queriy_history.list`, and return those logs along with the datasets. 

### 🧪 Tested?

- Modified and ran unit tests.
- Tested with staging configurations. Was able to extract query logs with the new SDK. See description for sample query log entry.

### Description

A sample query log entry:
```json
        {
          "_id": "UNITY_CATALOG:01ee7724-1377-18f6-9781-b0aa2bb3fd13",
          "bytesRead": 0.0,
          "bytesWritten": 0.0,
          "duration": 33686.0,
          "platform": "UNITY_CATALOG",
          "queryId": "01ee7724-1377-18f6-9781-b0aa2bb3fd13",
          "rowsRead": 0.0,
          "rowsWritten": 3.0,
          "sql": "-- this is a system generated query from catalog explorer\ndescribe history `main`.`default`.`testing` LIMIT 100;",
          "sqlHash": "fd1d327232c9df5c354a6c7cd51276a7",
          "startTime": "2023-10-30T12:58:52.420000+00:00",
          "userId": "dev@metaphor.io"
        }
```


usefulalgorithm:
# TODO

- [ ] add config values in the crawler configs:
```python
@dataclass(config=ConnectorConfig)
class SnowflakeQueryLogConfig:
    # Number of days back of query logs to fetch, if 0, don't fetch query logs
    lookback_days: int = 1

    # Query log filter to exclude certain usernames
    excluded_usernames: Set[str] = field(default_factory=lambda: set())

    # The number of query logs to fetch from Snowflake in one batch
    fetch_size: int = DEFAULT_QUERY_LOG_FETCH_SIZE

    # Queries larger than this size will not be processed
    max_query_size: int = DEFAULT_MAX_QUERY_SIZE
```
- [ ] Test on a real environment that new `list_table/column_lineage` works

pyproject.toml
```diff
@@ -48,6 +48,7 @@ sql-metadata = { version = "^2.8.0", optional = true }
 sqllineage = { version = "~=1.3.8", optional = true }
 tableauserverclient = { version = "^0.25", optional = true }
 thoughtspot_rest_api_v1 = { version = "1.5.3", optional = true }
+databricks-sql-connector = "^2.9.3"
```
usefulalgorithm, line 51: remove this

metaphor/unity_catalog/config.py
```diff
@@ -1,16 +1,30 @@
...
 from metaphor.common.filter import DatasetFilter
 
 
+@dataclass(config=ConnectorConfig)
+class UnityCatalogQueryLogConfig:
+    # Filter for querying logs. No filter if not specified.
+    query_start_time_range: Optional[TimeRange] = None
+
+    # Query log filter to match certain usernames
+    user_ids: List[int] = field(default_factory=lambda: [])
```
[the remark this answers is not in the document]
mars-lan: Yeah, generally it's easier to say which users you don't want, rather than the other way round.

metaphor/unity_catalog/extractor.py
```diff
@@ -266,7 +280,55 @@ def _populate_lineage(self, dataset: Dataset):
...
+            ):  # SDK throws if it's already enabled
+                raise e
+        except Exception as e:
+            raise e
+
+        logs: List[QueryLog] = []
+        for query_info in self._api.query_history.list(
+            filter_by=self._query_filter,
+            include_metrics=True,
+            max_results=None,  # No pagination!
```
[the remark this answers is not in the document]
mars-lan: > one can provide a page token to browse the next page
This will likely lead to timeout or other performance issues. Let's do proper pagination.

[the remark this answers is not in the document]
usefulalgorithm: upon more staring at the sdk source code it looks like `page_token` acts more like an internal thing, the caller wouldn't need to worry about it:
```python
        query = {}
# snipped
        if page_token is not None: query['page_token'] = page_token
# snipped

        while True:
            json = self._api.do('GET', '/api/2.0/sql/history/queries', query=query, headers=headers) # Gets the thing
            if 'res' not in json or not json['res']:
                return
            for v in json['res']:
                yield QueryInfo.from_dict(v)
            if 'next_page_token' not in json or not json['next_page_token']: # If the returned queries can't fit in a single page, `'next_page_token'` would appear in the json body
                return
            query['page_token'] = json['next_page_token'] # This handles pagination
```
So I think we just set `max_results` to specify how many queries can a single page store and it should be sufficient.

metaphor/unity_catalog/utils.py
```diff
@@ -11,8 +17,8 @@ def list_table_lineage(client: ApiClient, table_name: str) -> TableLineage:
     resp = None
 
     try:
-        resp = client.perform_query(
-            "GET", "/lineage-tracking/table-lineage", data=_data, version="2.0"
+        resp = client.do(
+            "GET", "/api/2.0/lineage-tracking/table-lineage", data=json.dumps(_data)
```
usefulalgorithm, line 21: we have to dump the payload to a json string here with the new api client method

mars-lan (commented):
Please also update https://github.com/MetaphorData/connectors/blob/main/metaphor/unity_catalog/README.md with the new configs and the requirement of Databricks Premium or above.

metaphor/unity_catalog/extractor.py
```diff
@@ -266,7 +279,55 @@ def _populate_lineage(self, dataset: Dataset):
...
+        except DatabricksError as e:
+            if (
+                e.error_code != "SCHEMA_ALREADY_EXISTS"
+            ):  # SDK throws if it's already enabled
+                raise e
+        except Exception as e:
+            raise e
+
+        logs: List[QueryLog] = []
+        for query_info in self._api.query_history.list(
```
mars-lan, line 296: What happens if the user has the standard plan? Does the API simply return an empty list or throw an exception?

mars-lan (commented):
Please also include a sample entry of the query log in the description.

metaphor/unity_catalog/models.py
mars-lan, line 1: Let's try to replace the rest of custom models with the corresponding data classes in the new SDK in a follow-up PR.

metaphor/unity_catalog/utils.py
```diff
@@ -42,3 +48,26 @@ def list_column_lineage(
             return ColumnLineage()
 
         raise e
+
+
+def parse_query_log_filter_by(
```
mars-lan, line 53:
```suggestion
def build_query_log_filter_by(
```

metaphor/unity_catalog/utils.py
```diff
@@ -42,3 +48,26 @@ def list_column_lineage(
...
+    query_filter = QueryFilter(
+        query_start_time_range=TimeRange(
+            end_time_ms=int(end_time.timestamp() * 1000),
+            start_time_ms=int(start_time.timestamp() * 1000),
+        )
+    )
+    if config.excluded_usernames:
+        query_filter.user_ids = [
+            user.id
+            for user in client.users.list()
```
mars-lan, line 69: Not sure if this is scalable when the list of users is large. It's fine to do post-filtering, i.e. dumping all the query logs and dropping the one with the unwanted users.

metaphor/unity_catalog/models.py
```diff
@@ -41,26 +27,21 @@ def __str__(self):
...
-    sql_path: Optional[str] = None
-    table_type: TableType
-    updated_at: int
-    updated_by: str
-    view_definition: Optional[str] = None
-
-
-def parse_table_from_object(obj: object):
-    return TypeAdapter(Table).validate_python(obj)
+def parse_schema_field_from_column_info(column: ColumnInfo) -> SchemaField:
```
mars-lan, line 30:
```suggestion
def extract_schema_field_from_column_info(column: ColumnInfo) -> SchemaField:
```

metaphor/unity_catalog/utils.py
usefulalgorithm: I guess if there are more queries than users, then filtering users before getting all queries should be more efficient?

metaphor/unity_catalog/models.py
usefulalgorithm: sg!

usefulalgorithm: https://app.shortcut.com/metaphor-data/story/22005

metaphor/unity_catalog/utils.py
mars-lan: Which is almost always the case. But the issue here is whether the API can accept (and perform adequately) a large list of users as part of the filter.

metaphor/unity_catalog/extractor.py
usefulalgorithm: The docs doesn't say this is a premium feature. I think it should be fine?
https://docs.databricks.com/en/sql/admin/query-history.html

mars-lan (commented):
LGTM at a high level. Will let @elic-eon give the final ship it.

metaphor/unity_catalog/utils.py
usefulalgorithm: I think it should be able to do that, and this should be fine for now. IMO let's just stick with this and if this proves to be a bad idea in the future we can come back and fix this?

elic-eon (approved):

PR #654, opened 2023-11-01
Get email address for the power bi view activity

elic-eon:
### 🤔 Why?

We need to show the actors even if they are not in Metaphor system. If we only captured `personId` we would not be able to get unknown person information.

### 🤓 What?

Fill email in actorInfo

### 🧪 Tested?

Unit test

### DIFF
--- a/metaphor/power_bi/extractor.py
+++ b/metaphor/power_bi/extractor.py
@@ -61,6 +61,7 @@ from metaphor.models.metadata_change_event import (
     PowerBIWorkspaceUser,
     SourceInfo,
     UserActivity,
+    UserActivityActorInfo,
     UserActivitySource,
     UserActivityType,
     VirtualView,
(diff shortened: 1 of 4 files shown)


metaphor/power_bi/extractor.py
```diff
@@ -690,6 +693,7 @@ def extract_activities(self) -> List[UserActivity]:
                     actor=user_id,
                     measure=1.0,
                     timestamp=activity.CreationTime,
+                    actor_info=UserActivityActorInfo(email=activity.UserId),
```
usefulalgorithm, line 696:
```suggestion
                    actor_info=UserActivityActorInfo(email=actor_email),
```

usefulalgorithm (changes requested):
just a nit, otherwise lgtm

usefulalgorithm (approved):
LGTM

tz.txt
```diff
@@ -0,0 +1,19 @@
+localTimeZoneId='Dateline Standard Time'
```
mars-lan, line 1: Where is this used?

elic-eon: Dropped

mars-lan (approved):

PR #655, opened 2023-11-01
Fix missing SQL code in dbt models

mars-lan:
### 🤔 Why?

Some dbt models are missing the raw/complied SQL code in the generated MCE.

### 🤓 What?

 

Use `hasattr` to test the existence of version-specific fields instead of relying on logic based on `isinstance`. The latter is more error-prone and more difficult to reason, which led to the original bug.

### 🧪 Tested?


alyiwang (approved):
LGTM

PR #656, opened 2023-11-02
Support column descriptions for `custom.governance` extractor

usefulalgorithm:
### 🤔 Why?

For tags we already have support for both table-level and column-level, but for descriptions only table-level is currently supported. This PR aims to implement column-level descriptions.

Tracked with [SC22046](https://app.shortcut.com/metaphor-data/story/22046).

### 🤓 What?

- Implemented the parsing logic.
- Updated custom.governance readme with example.
- Added unit test.
-

### 🧪 Tested?

Ran unit tests locally.

### TODO

- [x] Bump models package version

### DIFF
--- a/metaphor/custom/governance/extractor.py
+++ b/metaphor/custom/governance/extractor.py
@@ -5,13 +5,10 @@ from metaphor.common.entity_id import to_person_entity_id
 from metaphor.common.logger import get_logger
 from metaphor.custom.governance.config import CustomGovernanceConfig
 from metaphor.models.metadata_change_event import (
-    ColumnTagAssignment,
     Dataset,
-    DescriptionAssignment,
     MetadataChangeEvent,
     Ownership,
     OwnershipAssignment,
-    TagAssignment,
 )
 
 logger = get_logger()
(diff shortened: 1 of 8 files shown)


metaphor/custom/governance/extractor.py
```diff
@@ -67,16 +66,19 @@ async def extract(self) -> List[MetadataChangeEvent]:
...
-
                 dataset.description_assignment = DescriptionAssignment(
-                    asset_descriptions=asset_descriptions
+                    asset_descriptions=[
+                        d.to_asset_description() for d in governance.descriptions
+                    ]
                 )
 
+            if len(governance.column_descriptions) > 0:
+                if dataset.description_assignment is None:
```
mars-lan, line 76: This relies on the order of execution, which is less readable. Consider pre-creating `DescriptionAssignment` if `len(governance.descriptions) + len(column_descriptions) > 0`.

usefulalgorithm: I did this because the tag assignments part was implemented like this:
```python
            if len(governance.column_tags) > 0:
                if dataset.tag_assignment is None:
                    dataset.tag_assignment = TagAssignment()

                dataset.tag_assignment.column_tag_assignments = [
                    ColumnTagAssignment(
                        column_name=column_tag.column, tag_names=column_tag.tags
                    )
                    for column_tag in governance.column_tags
                ]
```

usefulalgorithm: to me it's easier to understand if those are actually methods under the `DatasetGovernance` class, so I moved those there.

mars-lan: Makes sense. Might as well move ownership assignment too?

pyproject.toml
```diff
@@ -29,11 +29,11 @@ google-cloud-logging = { version = "^3.5.0", optional = true }
 jsonschema = "^4.18.6"
 lkml = { version = "^1.3.1", optional = true }
 looker-sdk = { version = "^23.6.0", optional = true }
-metaphor-models = "0.27.11"
+metaphor-models = "0.27.12"
 msal = { version = "^1.20.0", optional = true }
 pycarlo = { version = "^0.8.1", optional = true }
 pydantic = { version = "^2.4.2", extras = ["email"]}
-pymssql = { version = "^2.2.9", optional = true }
+pymssql = { version = "2.2.9", optional = true }
```
mars-lan, line 36: Why fixing the version?

usefulalgorithm: if I update it would fail to build

usefulalgorithm: nvm, I'm changing this back

mars-lan (approved):
Great work

PR #657, opened 2023-11-03
Fix PyPI publication by updating snowflake-connector-python

mars-lan:
### 🤔 Why?

We were unable to publish to PyPI due to the workaround suggested by https://github.com/wbond/oscrypto/issues/78#issuecomment-1754169413. 

### 🤓 What?

Now that the latest version of https://github.com/snowflakedb/snowflake-connector-python has moved away from `oscrypt`, we can simply update and remove the workaround.

### 🧪 Tested?


alyiwang (approved):
Finally!
Thanks for the fix

PR #658, opened 2023-11-03
Extract pbi sensitivity label information

elic-eon:
### 🤔 Why?

Get sensitivity label information from Graph

### 🤓 What?

- Factor out some static function
- Factor out models from client
- Get sensitivity label information
- Update `pydantic` model

### 🧪 Tested?

Unit tests


pyproject.toml
```diff
@@ -31,6 +31,8 @@ lkml = { version = "^1.3.1", optional = true }
 looker-sdk = { version = "^23.6.0", optional = true }
 metaphor-models = "0.27.12"
 msal = { version = "^1.20.0", optional = true }
+msgraph-beta-sdk = { version = "1.0.0", optional = true }
```
mars-lan, line 34: beta?

metaphor/power_bi/README.md
```diff
@@ -10,9 +10,18 @@ We recommend creating a dedicated Azure AD Application and a dedicated security
 
 > Note: Make sure to NOT add any Power BI Service permissions to the app. Doing so will lead to authentication errors when calling the APIs. See [this notice](https://docs.microsoft.com/en-us/power-bi/enterprise/read-only-apis-service-principal-authentication#:~:text=Make%20sure%20there%20are%20no%20Power%20BI%20admin%2Dconsent%2Drequired%20permissions%20set%20on%20this%20application.%20For%20more%20information%2C%20see%20Managing%20consent%20to%20applications%20and%20evaluating%20consent%20requests.) for more details.
 
-2. Follow [this doc](https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/active-directory-groups-create-azure-portal#create-a-basic-group-and-add-members) to create a security group and add the app's service principal as a member.
-
-3. Log into [Power BI Admin Portal](https://app.powerbi.com/admin-portal/tenantSettings) as a Power BI admin, enable the following settings under **Admin API settings** for the security group created in the previous step:
+2. To get the sensitivity labe information. Follow the steps to ddd `InformationProtectionPolicy.Read.All` to the dedicated Azure AD application.
```
mars-lan, line 13: Can we make sure that this is an optional step, i.e. the crawler will still function even if it's missing the permission?

pyproject.toml
elic-eon: Right. https://learn.microsoft.com/en-us/graph/api/security-informationprotection-list-sensitivitylabels?view=graph-rest-beta&tabs=http The API is in beta.

metaphor/power_bi/README.md
elic-eon: Sure. It's optional.

pyproject.toml
mars-lan: Please add a comment to that effect for posterity.

metaphor/power_bi/models.py
```diff
@@ -0,0 +1,235 @@
...
+from metaphor.common.models import BaseModel
+
+
+class PowerBIApp(BaseModel):
+    id: str
+    name: str
+    workspaceId: str
+
+
+class PowerBIDataSource(BaseModel):
```
usefulalgorithm, line 14: these should probably still be `V1CompatBaseModel`, unless the configs have proper str ids

usefulalgorithm:
```yaml
id: 12345 
```
Pydantic V2 would parse this as an `int`, if we do not specify `coerce_num_to_str` in the config it will straight up break.

---

If the config looks like
```yaml
id: "12345"
```
then it is fine.

metaphor/power_bi/power_bi_utils.py
```diff
@@ -0,0 +1,186 @@
...
+    refreshes: List[PowerBIRefresh],
+) -> Optional[datetime]:
+    try:
+        # Assume refreshes are already sorted in reversed chronological order
+        # Empty endTime means still in progress
+        refresh = next(
+            r for r in refreshes if r.status == "Completed" and r.endTime != ""
+        )
+    except StopIteration:
+        return None
```
usefulalgorithm, line 147:
```suggestion
    # Assume refreshes are already sorted in reversed chronological order
    # Empty endTime means still in progress
    refresh = next(
        (r for r in refreshes if r.status == "Completed" and r.endTime != ""),
        None
    )
    if refresh is None:
        return None
```

metaphor/power_bi/models.py
elic-eon: The id here is UUID. should be safe?

metaphor/power_bi/power_bi_utils.py
elic-eon: If refreshes only contain running task, we need to handle `StopIteration`

metaphor/power_bi/models.py
usefulalgorithm: if it's uuid then it's fine.

usefulalgorithm (commented):
LGTM, but I don't think I know enough abourt PowerBI to be able to approve this.

metaphor/power_bi/README.md
```diff
@@ -58,6 +58,15 @@ When extracting lineages, the connector parses the snowflake account directly fr
 snowflake_account: <snowflake_account>
 ```
 
+To get the sensitivity labe information. Follow the steps to ddd `InformationProtectionPolicy.Read.All` to the dedicated Azure AD application.
```
mars-lan, line 61: Sorry what I meant is to still keep this under "Setup", but add an `(Optional) ` prefix.

mars-lan (commented):
Otherwise LGTM

metaphor/power_bi/README.md
elic-eon: Reverted.

mars-lan (approved):

PR #659, opened 2023-11-03
Use v2 `ConfigDict` instead of custom class for pydantic dataclasses

usefulalgorithm:
### 🤔 Why?

Because I am wary of seeing
```
UserWarning: Valid config keys have changed in V2:
* 'validate_all' has been renamed to 'validate_default'
```

### 🤓 What?

Use properly typed `ConfigDict` instead of custom class as dataclass configs.

### 🧪 Tested?

Ran unit tests.

### DIFF
--- a/metaphor/common/dataclass.py
+++ b/metaphor/common/dataclass.py
@@ -1,11 +1,5 @@
-from pydantic import Extra
+from pydantic import ConfigDict
 
-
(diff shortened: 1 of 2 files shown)


mars-lan (commented):
Bump up version?

mars-lan (approved):

PR #661, opened 2023-11-05
Drop oscrypt to fix PyPi publication

mars-lan:
### 🤔 Why?

It was removed in https://github.com/MetaphorData/connectors/pull/657 but was brought back in https://github.com/MetaphorData/connectors/pull/658 by accident.

### 🤓 What?

As titled.

### 🧪 Tested?

N/A


poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.6.1 and should not be changed by hand.
```
alyiwang, line 1: May consider upgrading poetry version

mars-lan: Updated

alyiwang (approved):

PR #660, opened 2023-11-05
Handle email-based Snowflake user ID in query logs

mars-lan:
### 🤔 Why?

It's possible to have email address as a Snowflake user ID. Query issued by these users should have the `email` address field set instead of `user_id`. 

### 🤓 What?

As titled.

### 🧪 Tested?

Verified against a production instance. Example:

```
...
                {
                    "_id": "SNOWFLAKE:01b01a65-0603-c12b-0029-c00304f93dae",
                    "account": "metaphor-dev",
                    "bytesRead": 0.0,
                    "bytesWritten": 0.0,
                    "cost": 0.0,
                    "duration": 0.326,
                    "email": "mars@metaphor.io",
                    "platform": "SNOWFLAKE",
                    "queryId": "01b01a65-0603-c12b-0029-c00304f93dae",
                    "rowsWritten": 0.0,
                    "sources":
                    [],
                    "sql": "GET '@~/worksheet_data/metadata' 'file:///'",
                    "sqlHash": "70496b3935c36ec0fc4027d9a26d1dd5",
                    "startTime": "2023-11-04T03:49:52.298000-08:00",
                    "targets":
                    []
                },
...
```

PR #662, opened 2023-11-06
Set email field in UC query log if it contains a valid email address

mars-lan:
### 🤔 Why?

Simliar to https://github.com/MetaphorData/connectors/pull/660, but for Unity Catalog.

### 🤓 What?

As titled. Also factor out common code to an until method.

### 🧪 Tested?


usefulalgorithm (approved):
讚

PR #663, opened 2023-11-06
Update metaphor models

elic-eon:
### 🤔 Why?

Update `metaphor-models` to latest; otherwise, the ingestion would fail

### 🤓 What?

- Update model version
- Drop `pipelineEntityId` field

### 🧪 Tested?

Unit tests


mars-lan:
> Do we need to bump minor version for this change?

Ideally, since it's a backward incompatible change.

mars-lan:
> > Do we need to bump minor version for this change?
> 
> Ideally, since it's a backward incompatible change.

Actually never mind. This is the connectors repo. No need to bump since there's no user-facing changes.

PR #664, opened 2023-11-06
Parse `run_results.json` into `DatasetDataQuality`

usefulalgorithm:
### 🤔 Why?

`dbt run` and `dbt test` generates an artifact named `run_results.json` containing the results from the run command. We can use the outcome from `dbt test`'s `run_results.json` to provide dataset quality indicators.

### 🤓 What?

- `ManifestParser` -> `ArtifactParser`
- Generated dataclass definitions for `run_results.json`
- Implemented logic for parsing dbt run results to `DatasetDataQuality`

### 🧪 Tested?

- Unit test
- Ran against artifacts generated by locally running dbt core
- Ran against dbt cloud artifacts
- Tested end-to-end: 
截圖 2023-11-08 晚上11 24 54


metaphor/dbt/artifact_parser.py
```diff
@@ -378,6 +390,96 @@ def _parse_test(self, test: TEST_NODE_TYPE) -> None:
...
+                        DataMonitorStatus.UNKNOWN,
+                        DataMonitorStatus.PASSED,
+                    ]:
+                        if status in statuses:
+                            return status
+                    assert False  # Impossible!
+
+                exisiting_monitor.status = get_next_status(
+                    [exisiting_monitor.status, status]
+                )
```
usefulalgorithm, line 481: pretty sure this is too hard to follow... refactor this into something easier to read

metaphor/dbt/artifact_parser.py
```diff
@@ -378,6 +390,96 @@ def _parse_test(self, test: TEST_NODE_TYPE) -> None:
...
+        def is_target_dataset(dataset: Dataset) -> bool:
+            name = dataset_normalized_name(
+                model.database, model.schema_, model.alias or model.name
+            ).lower()
+            # We found the dataset if the name matches or it's dbt model's materialized table
+            return dataset.logical_id and (
+                dataset.logical_id.name == name
+                or str(to_dataset_entity_id(name, self._platform, self._account))
+                == dbt_model.materialization.target_dataset
+            )
```
usefulalgorithm, line 435: is this logic correct?

metaphor/dbt/util.py
```diff
@@ -226,3 +233,76 @@ def build_source_code_url(
...
+    def is_target_dataset(dataset: Dataset) -> bool:
+        name = dataset_normalized_name(database, schema, table).lower()
+        # We want to find the dataset whose name matches or is dbt model's materialized table
+        return dataset.logical_id and (
+            dataset.logical_id.name == name
+            or str(to_dataset_entity_id(name, platform, account))
+            == dbt_model.materialization.target_dataset
+        )
+
+    return next((dataset for dataset in datasets if is_target_dataset(dataset)), None)
```
usefulalgorithm, line 281: I want to find the dataset (the materialized table) that is referenced by this test run. Not sure if the filtering logic is correct here.

metaphor/dbt/config.py
```diff
@@ -41,6 +41,8 @@ class MetaTag:
 class DbtRunConfig(BaseConfig):
     manifest: str
 
+    run_results: str
```
mars-lan, line 44: Might want to make this optional and process the file only when it's available.

metaphor/dbt/artifact_parser.py
```diff
@@ -188,8 +193,19 @@
...
+    "fail": DataMonitorStatus.ERROR,
+    "runtime error": DataMonitorStatus.ERROR,
+    "pass": DataMonitorStatus.PASSED,
+    "success": DataMonitorStatus.PASSED,
+}
+"""Maps a `RunResultOutput.status` to `DataMonitorStatus`."""
+
 
-class ManifestParser:
+class ArtifactParser:
```
mars-lan, line 208: Instead of doing this, let's create a separate `RunResultParser` for separation of responsibility.

metaphor/dbt/gen_models.sh
```diff
@@ -7,27 +7,31 @@ cd ${SCRIPT_DIR}
 # Generate various data models for dbt manifest & catalog using official JSON schemas
 
 if [ $# -ne 2 ]; then
-  echo "Usage: $0 <manifest or catalog> <version: v1, v2...>"
+  echo "Usage: $0 <manifest, catalog or run_results> <version: v1, v2...>"
```
mars-lan, line 10: We can kill catalog now since it's no longer used by the versions of dbt we support.

metaphor/dbt/generated/dbt_run_results_v4.py
```diff
@@ -0,0 +1,114 @@
+# mypy: ignore-errors
```
mars-lan, line 1: Need to figure out if we need to support v1-v3, based on the dbt version we support (1.0+).

mars-lan (commented):
Some high-level comments.

metaphor/dbt/artifact_parser.py
usefulalgorithm: That was my initial idea, but then `RunResultOutput` doesn't know about which dbt model it's related to, instead it only holds an unique_id. This means `RunResultParser` needs to have access to whatever was parsed by `ManifestParser`, and there really isn't much to separate.

metaphor/dbt/config.py
usefulalgorithm: it's always gonna be available, if not then there wouldn't be anything to parse?

metaphor/dbt/util.py
usefulalgorithm: @mars-lan can you take a look at this and gimme some tips if it's incorrect? Thanks

metaphor/dbt/generated/dbt_run_results_v4.py
usefulalgorithm: I'm guessing we don't need to support v1-v3:
- the oldest doc version is v1.2, which says run_result.json has v4 schema
- the [parsing part in dbt repo](https://github.com/dbt-labs/dbt-core/tree/v1.1.0/core/dbt/parser) doesn't have anything for run_results.json
- for `manifest.json`, v4 corresponds to dbt core v1.0.

I think it make sense to assume `run_results.json` schema version is v4 for dbt core v1.0 and v1.1, or in these two versions there's no `run_results.json`.

metaphor/dbt/config.py
usefulalgorithm: The following commands would generate `run_results.json`:
- build
- compile
- docs generate
- run
- seed
- snapshot
- test
Is it possible to run dbt pipeline without doing any of the above? If so then I'll change this to optional.

mars-lan: What about `dbt parse`? https://docs.getdbt.com/reference/commands/parse

metaphor/dbt/generated/dbt_run_results_v4.py
mars-lan: Could you verify with v1 & v1.1 using docker images: https://github.com/MetaphorData/dbt#run-older-versions-of-dbt

metaphor/dbt/config.py
usefulalgorithm: ah yes, then I'll change this to optional

metaphor/dbt/generated/dbt_run_results_v4.py
usefulalgorithm: v1.0:
```
{
  "metadata": {
    "dbt_schema_version": "https://schemas.getdbt.com/dbt/run-results/v4.json",
    "dbt_version": "1.0.9",
    "generated_at": "2023-11-07T15:04:33.416658Z",
    "invocation_id": "6e74957c-a718-4d59-9ce6-4c0c032427a9",
    "env": {}
  },
```

v1.1:
```
{
  "metadata": {
    "dbt_schema_version": "https://schemas.getdbt.com/dbt/run-results/v4.json",
    "dbt_version": "1.1.5",
    "generated_at": "2023-11-07T15:05:46.417274Z",
    "invocation_id": "e0496c5e-b5d5-4834-bbfd-0997280899db",
    "env": {}
  },
```

mars-lan: Great. Please update the table in that repo for posterity.

usefulalgorithm: https://github.com/MetaphorData/dbt/pull/89

metaphor/dbt/artifact_parser.py
usefulalgorithm: More specifically, [here](https://github.com/MetaphorData/connectors/pull/664/files#diff-a1df7e7750265ceca906dcd5962a36e71a550fa20dea40d2729a4ccde150bc4dR428) we actually need the dbt model to be able to map test results back to its target dataset. If we parse `run_results.json` in a separate class, then `ManifestParser` would need to either store the dbt models or return them to make them visible to `RunResultsParser`. Instead I think it makes more sense to parse both artifacts in the same class, so as to indicate that what we're parsing is dependent on both artifacts.

mars-lan: SGTM

metaphor/dbt/util.py
alyiwang: Just to clarify a couple of things:
- Is the `datasets` all the source datasets parsed from manifest before?
- Does the dbt test work on the source dataset or the generated (materialized) dataset? I'm thinking the latter.
- If test is on the materialized target table, then maybe just check the dbt model target dataset?

usefulalgorithm: > Is the datasets all the source datasets parsed from manifest before?
Not really, it can contain both source datasets and model datasets (I don't know what this actually mean, in the debugger I can see `datasets` having two kinds of keys:
- "source.XXX.YYY...": this should mean its a source dataset
- "model.ZZZ.WWW...": this dataset uses the model as the key. Is this the materialized dataset?
 
> Does the dbt test work on the source dataset or the generated (materialized) dataset?
It works on the dbt model. From the manifest:
```
  "test.metaphor_subscriptions.unique_subscriptions_sales_SALES_ID.f83222e32c": {
    ...
    "depends_on": {
      "macros": [
        "macro.dbt.test_unique",
        "macro.dbt.get_where_subquery",
        "macro.dbt_snowflake.set_query_tag",
        "macro.dbt.materialization_test_default",
        "macro.dbt_snowflake.unset_query_tag"
      ],
      "nodes": [
        "model.metaphor_subscriptions.subscriptions_sales"
      ]
    }
```
So I guess it works on the materialized table? Which can be absent in the manifest artifact?

> If test is on the materialized target table, then maybe just check the dbt model target dataset?
I looked into this a little bit more and fixed the code, can you take a look at the renewed logic?


Thanks a bunch!

metaphor/dbt/artifact_parser.py
```diff
@@ -378,6 +404,39 @@ def _parse_test(self, test: TEST_NODE_TYPE) -> None:
...
+        status = dbt_run_result_output_data_monitor_status_map[run_result.status]
+
+        # Get the materialized table's entity id...
+        target_dataset_entity_id = self._get_model_entity_id(model)
+        # ... and use it to find the dataset we're looking for. Note that
+        # at this point all datasets have been parsed.
+        dataset = find_target_dataset(
+            self._datasets.values(),
+            target_dataset_entity_id,
+        )
```
usefulalgorithm, line 430: @mars-lan  @alyiwang  I modified the condition to get the right dataset, and added some comments to describe what I'm thinking. Please take a look and see if this implementation is correct. Thanks!

mars-lan (commented):
Please also update https://github.com/MetaphorData/connectors/blob/main/README.md#connectors

tests/dbt/data/trial_v4/run_results.json
```diff
@@ -0,0 +1,11 @@
+{
+  "metadata": {
+    "dbt_schema_version": "https://schemas.getdbt.com/dbt/run-results/v4.json",
+    "dbt_version": "1.0.0",
+    "generated_at": "2023-11-06T08:14:33.314124Z",
+    "invocation_id": "4608fb03-ed99-4fcf-9408-cb71b627cc1b",
+    "env": {}
+  },
+  "results": [],
```
mars-lan, line 9: Recommend including the run results for old versions as well to ensure full coverage.

metaphor/dbt/gen_models.sh
```diff
@@ -7,27 +7,29 @@ cd ${SCRIPT_DIR}
...
-OUTPUT=generated/dbt_${SCHEMA}_${VERSION}.py
-
+PATH_ELEMENT=$SCHEMA
 CLASS_NAME=""
 if [[ "${SCHEMA}" == "manifest" ]]; then
   CLASS_NAME="DbtManifest"
-elif [[ "${SCHEMA}" == "catalog" ]]; then
-  CLASS_NAME="DbtCatalog"
+elif [[ "${SCHEMA}" == "run_results" ]]; then
+  PATH_ELEMENT="${SCHEMA//_/-}"
```
mars-lan, line 22: Rather doing this, why not change the argument to `run-results` for consistency?

tests/dbt/test_extractor.py
```diff
@@ -71,16 +71,27 @@ async def test_jaffle_v10(test_root_dir):
     )
 
 
+@pytest.mark.asyncio
+async def test_jaffle_valid_tests_v10(test_root_dir):
```
mars-lan, line 75: Instead of doing this, I'll just modify jaffle dbt project to output run_results.json.

tests/dbt/data/trial_v4/run_results.json
usefulalgorithm: I think that's non trivial, can we do that afterwards?

usefulalgorithm: I agree it's better to have those run results for older versions

usefulalgorithm:
@mars-lan 

> Please also update https://github.com/MetaphorData/connectors/blob/main/README.md#connectors

What should I update in this file?

mars-lan:
> @mars-lan
> 
> > Please also update https://github.com/MetaphorData/connectors/blob/main/README.md#connectors
> 
> What should I update in this file?

Never mind, we already have "tests" for dbt in the table.

tests/dbt/data/trial_v4/run_results.json
mars-lan: Please create a ticket to track it.

metaphor/dbt/gen_models.sh
usefulalgorithm: Is it inconsistent now? Reason why I had it like this is because `run-results.py` is not a valid python file name, but the url for the schema is `run-results` instead of `run_results`. I can modify the logic if you think it's better that way

mars-lan: Ideally the name of `$1` should match the URL, i.e. 
```
./gen_models.sh run-resuls v4
```

The actual filename can be `run_results.py`

metaphor/dbt/util.py
```diff
@@ -226,3 +233,72 @@ def build_source_code_url(
...
+        ),
+        None,
+    )
+    if exisiting_monitor is None:
+        monitor = DataMonitor(
+            targets=[DataMonitorTarget(column=column_name)], status=status
+        )
+        dataset.data_quality.monitors.append(monitor)
+    else:
+        exisiting_monitor.status = find_most_severe_status(
```
mars-lan, line 302: Not exactly sure what you're trying to achieve here. Please add some comments.

metaphor/dbt/artifact_parser.py
```diff
@@ -378,6 +404,39 @@ def _parse_test(self, test: TEST_NODE_TYPE) -> None:
...
+            logger.warn(f"Cannot find run result for test: {test.unique_id}")
+            return
+
+        status = dbt_run_result_output_data_monitor_status_map[run_result.status]
+
+        # Get the materialized table's entity id...
+        target_dataset_entity_id = self._get_model_entity_id(model)
+        # ... and use it to find the dataset we're looking for. Note that
+        # at this point all datasets have been parsed.
+        dataset = find_target_dataset(
```
mars-lan, line 427: I assume `self._dataset` is already keyed by entity ID?

usefulalgorithm: no, it is keyed by a plain string, not entity id.

metaphor/dbt/util.py
usefulalgorithm: Say the there are two tests for column `foo`, one passed but the other one ended up spitting warnings. Since each `DataMonitor` only covers one column, it makes sense to try to locate the existing monitor, and update its status with the more severe status out of the test results on the column.

tests/dbt/data/trial_v4/run_results.json
usefulalgorithm: https://app.shortcut.com/metaphor-data/story/22269/create-meaningful-tests-for-run-results-json-with-older-dbt-versions

metaphor/dbt/artifact_parser.py
mars-lan: Is it a string representation of entity ID?

tests/dbt/test_extractor.py
usefulalgorithm: it's actually not jaffle, but metaphor_subscription XD

metaphor/dbt/util.py
usefulalgorithm: ok.

metaphor/dbt/artifact_parser.py
usefulalgorithm: no, it's `EntityId.logical_id.name`

usefulalgorithm: or something similar, they always start with `source.` or `model.`

mars-lan: If it's `model.*` you can look it up using the same dbt model ID directly?

usefulalgorithm:
@mars-lan  can you take another look? thanks

mars-lan (approved):

mars-lan:
Great work!

mars-lan:
@usefulalgorithm  Looks like you forgot to bump the version. Please create a follow-up PR just for that, thanks.

PR #673, opened 2023-11-09
Capture ADF pipeline information

elic-eon:
### 🤔 Why?

Need this information to improve the ADF crawler to support sources like SFTP, REST resource.

### 🤓 What?

- Add json_dump for pipelines

### 🧪 Tested?

unit test


usefulalgorithm (approved):
猛

PR #674, opened 2023-11-09
Explicitly forbid setting `accout` in config when platform isn't `snowflake`

usefulalgorithm:
### 🤔 Why?

If the data platform isn't snowflake but user specified an account, the resulting logical id would be incorrect. We want to forbid people from accidentally including account in the configs when their data platform isn't snowflake.

### 🤓 What?

- Updated readme.
- Added new field validator for governance configs.
- Added unit test.
### 🧪 Tested?

Unit test passed.

### DIFF
--- a/metaphor/common/models.py
+++ b/metaphor/common/models.py
@@ -1,7 +1,7 @@
 from datetime import datetime
-from typing import Optional, Set, Union
+from typing import Optional, Union
 
-from pydantic import BaseModel, ConfigDict
+from pydantic import BaseModel, ConfigDict, model_validator
 from pydantic.dataclasses import dataclass
 
 from metaphor.common.dataclass import ConnectorConfig
@@ -34,20 +34,22 @@ class DeserializableDatasetLogicalID:
             self.account,
         )
 
-    def _platform_forbids_account_config(
-        self, allow_account: Set[DataPlatform]
-    ) -> bool:
+    @model_validator(mode="after")
+    def _platform_forbids_account_config(self) -> "DeserializableDatasetLogicalID":
         """
-        Returns `False` if the platform does not allow account to be set.
-
-        Parameters
-        ----------
-        allow_account : Set[DataPlatform]
-            The set of platforms that allow account to be set.
+        Throws if the platform does not allow account to be set.
         """
-        return (DataPlatform[self.platform] not in allow_account) and (
+        allow_account_be_set = {
+            DataPlatform.SNOWFLAKE,
+            DataPlatform.MSSQL,
+        }
+        if DataPlatform[self.platform] not in allow_account_be_set and (
             self.account is not None
-        )
+        ):
+            raise ValueError(
+                f"Field `account` only permitted for platforms: {[x.value for x in allow_account_be_set]}"
+            )
+        return self
 
 
 def to_dataset_statistics(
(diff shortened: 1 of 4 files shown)


metaphor/custom/governance/config.py
```diff
@@ -83,6 +85,13 @@ class DatasetGovernance:
         default_factory=lambda: []
     )
 
+    @field_validator("id")
+    @classmethod
+    def _only_snowflake_can_have_account(cls, id_: DeserializableDatasetLogicalID):
+        if DataPlatform[id_.platform] != DataPlatform.SNOWFLAKE and id_.account:
```
mars-lan, line 91: I believe some other platform, e.g. MS SQL (@elic-eon to confirm), also allows setting of `account`. This logic is also common across a few connectors so best move it to `metaphor.common`.

metaphor/custom/governance/README.md
```diff
@@ -50,7 +50,7 @@ output:
     directory: <output_directory>
 ```
 
-> Note: You only need to specify `account` if the platform is `SNOWFLAKE`.
+> Note: You only need to specify `account` if the platform is `SNOWFLAKE`. Do **NOT** specify `account` if the platform isn't `SNOWFLAKE`!
```
mars-lan, line 53: I think the first sentence is clear already. No need fo the extra emphasis, especially now a validtor is added.

usefulalgorithm: I added it because I was using BigQuery dataset and I didn't realize I'm not supposed to have account specified haha.

mars-lan: Understood. After adding the validator it should fail loud and clear.

metaphor/custom/governance/config.py
elic-eon: @mars-lan You're right. The MS SQL crawler is using `account` field in the logical ID.

metaphor/common/models.py
```diff
@@ -34,6 +34,23 @@ def to_entity_id(self):
...
         )
 
+    @model_validator(mode="after")
+    def _platform_forbids_account_config(self) -> "DeserializableDatasetLogicalID":
+        """
+        Throws if the platform does not allow account to be set.
+        """
+        allow_account_be_set = {
+            DataPlatform.SNOWFLAKE,
+            DataPlatform.MSSQL,
```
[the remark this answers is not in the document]
usefulalgorithm: Actually that probably won't be enough, specifying account for platforms that don't need it would not cause any error, instead the crawler would just generate a dataset entity id that does not exist in the actual database.

I think an exhaustive list of platforms is still the better way to go.

metaphor/common/models.py
```diff
@@ -34,6 +34,23 @@ def to_entity_id(self):
...
+        }
+        if DataPlatform[self.platform] not in allow_account_be_set and (
```
usefulalgorithm, line 46:
```python
if (DataPlatform[self.platform] not in allow_account_be_set) == (sefl.account is not None):
```

metaphor/common/models.py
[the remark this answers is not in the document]
usefulalgorithm: todo : move these to app/@models

PR #675, opened 2023-11-10
Fix parse dbt test error: assert -> return if test.column_name not found

usefulalgorithm:
### 🤔 Why?

In `run_results.json` it is possible for the test results to not have column names. Allow this in the connector code.

### 🤓 What?

as titled.

### 🧪 Tested?

N/A.


mars-lan (approved):

mars-lan:
Forgot to bump the version 😉

PR #681, opened 2023-11-13
Remove unused UnityCatalog dataclass

usefulalgorithm:
### 🤓 What?

Removed `DataSourceFormat`, which is no longer used once we switched to `databricks-sdk-py`.

### 🧪 Tested?

No need.


elic-eon (approved):
LGTM!

PR #682, opened 2023-11-13
Add CI check for updated version on actual code change

usefulalgorithm:
### 🤔 Why?

If there's actual changes in the repo we want to make sure the version number has been bumped.

### 🤓 What?

- Added CI action and script.
- Added checklist section in PR template.

### 🧪 Tested?

Tested locally.

### DIFF
--- a/.github/scripts/check_version.sh
+++ b/.github/scripts/check_version.sh
@@ -5,14 +5,7 @@ if git diff --name-only origin/main | grep ".*\.py$" > /dev/null 2>&1; then
     MAIN_TAG=$(git describe origin/main --tags --abbrev=0)
     CURRENT_VERSION=$(poetry version --short 2>/dev/null)
 
-    # Check if current version is larger than latest main version.
-    # Note: We want this program to exit 1 when the check fails. In Python `True` == `1`, so we want this
-    # comparison to evaluate to `0` if we want the check to pass.
-    LINES=(
-        "from packaging import version" 
-        "import sys" 
-        "sys.exit(version.parse(\"$MAIN_TAG\") >= version.parse(\"$CURRENT_VERSION\"))"
-    )
-    # Concat the lines and pass them to python
-    printf '%s;' "${LINES[@]}" | python
+    GIT_ROOT=$(git rev-parse --show-toplevel)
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.13.43"
+version = "0.13.44"
```
mars-lan, line 3: No need, since there's no real code change here 😉

.github/scripts/check_version.sh
```diff
@@ -0,0 +1,12 @@
...
+
+if git diff --name-only origin/main | grep ".*\.py$" > /dev/null 2>&1; then
+    # Found modified stuff, make sure version is bumped!
+    MAIN_TAG=$(git describe origin/main --tags --abbrev=0)
+    CURRENT_VERSION=$(poetry version --short 2>/dev/null)
+
+    # Check if current version is larger than latest main version.
+    # Note: We want this program to exit 1 when the check fails. In Python `True` == `1`, so we want this
+    # comparison to evaluate to `0` if we want the check to pass.
+    echo "from packaging import version; import sys; sys.exit(version.parse(\"$MAIN_TAG\") >= version.parse(\"$CURRENT_VERSION\"))" | python
```
mars-lan, line 11: Consider moving the inline python to a file to improve readability?

mars-lan: In fact the whole shell script could be replace by a single Python script?

usefulalgorithm: Hmm... I think this is actually good enough:

- single Python script -> not a fan of `GItPython`, nor of `subprocess.run("git diff --name-only", stdout=PIPE)`.
- move that line into its own file -> this line just compares 2 version numbers. I don't think it's that bad...

mars-lan: We'll need to do the same for https://github.com/MetaphorData/app/blob/main/%40models/pyproject.toml so would prefer a single script. At the minimum break the inline Python to a script for readability?

mars-lan (approved):

PR #685, opened 2023-11-15
Include all possible table types in the enum

usefulalgorithm:
### 🤔 Why?

We didn't include all possible snowflake table types in the enum, which could cause the connector to fail.

### 🤓 What?

As titled.

### 🧪 Tested?

N/A.

### ☑️ Checks


elic-eon (approved):
LGTM!

PR #687, opened 2023-11-16
`Hierarchy.logiclId.path` is case sensitive

usefulalgorithm:
### 🤔 Why?

Because it should be case sensitive, but isn't.

### 🤓 What?

Fixes that.

### 🧪 Tested?

Unit test.

### ☑️ Checks

### DIFF
--- a/metaphor/snowflake/extractor.py
+++ b/metaphor/snowflake/extractor.py
@@ -399,13 +399,17 @@ class SnowflakeExtractor(BaseExtractor):
         # Set hierarchy key and logical_id
         path = [DataPlatform.SNOWFLAKE.value]
         if object_type == "DATABASE":
-            path.extend([object_name])
+            path.extend(
+                [object_name.lower()]
+            )  # path should be normalized, i.e. should be lowercase
             hierarchy_key = dataset_normalized_name(object_name)
         else:
             if not database:
                 logger.error("Cannot extract schema level tag without database name")
                 return
-            path.extend([database, object_name])
+            path.extend(
+                [database.lower(), object_name.lower()]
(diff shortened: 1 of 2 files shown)


metaphor/snowflake/extractor.py
```diff
@@ -399,13 +399,17 @@ def _add_system_tag(
         # Set hierarchy key and logical_id
         path = [DataPlatform.SNOWFLAKE.value]
         if object_type == "DATABASE":
-            path.extend([object_name])
+            path.extend(
+                [object_name.lower()]
+            )  # path should be normalized, i.e. should be lowercase
             hierarchy_key = dataset_normalized_name(object_name)
```
mars-lan, line 405: Path is supposed to contain the original casing, whereas `dataset.name` is noramlized to lower case.

mars-lan (changes requested):

metaphor/snowflake/extractor.py
mars-lan: Actually never mind. I got confused by the `structure` field. This is the correct fix.

mars-lan: On third thought. @alyiwang can you confirm if `logicalId.path` should be normalized?

usefulalgorithm: The problem is that `object_name` is in all caps

elic-eon: `structure` should be the only one that contains case-sensitive name

alyiwang: Yes, `path` is the ID of `hierarchy`, so should be normalized.

mars-lan (approved):

PR #693, opened 2023-11-20
Get dbt table level test result

elic-eon:
### 🤔 Why?

We didn't set `MonitorTarget.dataset` so the table level DQ status will be removed because all fields are none.

### 🤓 What?

- Use `attatched_node` if possible because discovered that `test.depends_on[0]` is not always the real dataset.
- Set `MonitorTatget.dataset` to resolve the problem

### 🧪 Tested?

Update unit tests

### ☑️ Checks

### DIFF
--- a/metaphor/dbt/artifact_parser.py
+++ b/metaphor/dbt/artifact_parser.py
@@ -378,6 +378,8 @@ class ArtifactParser:
             return
 
         model_unique_id = test.depends_on.nodes[0]
+        if hasattr(test, "attached_node") and test.attached_node:
+            model_unique_id = test.attached_node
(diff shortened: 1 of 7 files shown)


metaphor/dbt/artifact_parser.py
```diff
@@ -434,6 +436,7 @@ def _parse_test_run_result(
         )
 
         status = dbt_run_result_output_data_monitor_status_map[run_result.status]
+        print(test.name, test.column_name)
```
mars-lan, line 439: Drop?

elic-eon: Dropped

mars-lan (approved):

PR #694, opened 2023-11-21
fix: [sc-22517] Support "External Table" source for Snowflake streams

usefulalgorithm:
Story details: https://app.shortcut.com/metaphor-data/story/22517

Reference: https://docs.snowflake.com/en/sql-reference/sql/show-streams#output


metaphor/snowflake/utils.py
```diff
@@ -51,6 +51,8 @@ class SnowflakeTableType(Enum):
 str_to_source_type: Dict[str, SnowflakeStreamSourceType] = {
     "TABLE": SnowflakeStreamSourceType.TABLE,
     "VIEW": SnowflakeStreamSourceType.VIEW,
+    "EXTERNAL_TABLE": SnowflakeStreamSourceType.TABLE,
```
mars-lan, line 54: Might be worthwhile creating explicit enums for these?

usefulalgorithm: I think let's do that after we actually do something with this value?

mars-lan (approved):

PR #696, opened 2023-11-21
Parse databricks table's last update time with unity catalog profile extractor

usefulalgorithm:
### 🤔 Why?

Now that we parse things from running commands thru SQL warehouse, we can extract the latest update for a specific dataset and parse its update time.

### 🤓 What?

Parsed last update time from SQL warehouse for databricks tables.

### 🧪 Tested?

Unit test.

### ☑️ Checks


metaphor/unity_catalog/profile/extractor.py
```diff
@@ -109,6 +109,12 @@ def _get_dataset_statistics(self, table_info: TableInfo) -> DatasetStatistics:
                     ]
                     dataset_statistics.data_size_bytes = bytes_count
                     dataset_statistics.record_count = row_count
+                cursor.execute(f"DESCRIBE HISTORY {escaped_name} LIMIT 1")
```
mars-lan, line 112: The history is sorted by time in descending order?

metaphor/unity_catalog/profile/extractor.py
```diff
@@ -109,6 +109,12 @@ def _get_dataset_statistics(self, table_info: TableInfo) -> DatasetStatistics:
...
+                latest_history = cursor.fetchone()
+                if latest_history:
```
mars-lan, line 114: Could there be any entries that are not related to table updates? In other words, can we assume the timestamp of the lastest entry always equal to the last refresh date?

metaphor/unity_catalog/profile/extractor.py
usefulalgorithm: according to the doc, yes

usefulalgorithm: > The operations are returned in reverse chronological order. By default table history is retained for 30 days.

metaphor/unity_catalog/profile/extractor.py
usefulalgorithm: There's the `operation` column that describes what that history row is, but there isn't an exhaustive list of possible values. Some values I've seen so far:
- 'WRITE'
- 'CREATE OR REPLACE TABLE'
- 'COPY INTO'
- 'CREATE TABLE'
- 'CREATE TABLE AS SELECT'
- 'SET TBLPROPERTIES'
- 'ADD CONSTRAINT'
I'm not sure if the last two count as modifications, but I'm pretty sure all the other ones are.

mars-lan: Yeah, the last two don't really count. We may need to add an exclusion list and post-filter the results, assuming `DESCRIBE HISTORY` doesn't take a `WHERE` clause.

mars-lan: Yeah, but I assume most of them are "real updates", which means we just need to exclude the ones that are not to our best knowledge. We can keep improving the exclusion list as we go.

usefulalgorithm: SG

metaphor/unity_catalog/profile/extractor.py
```diff
@@ -109,6 +109,15 @@ def _get_dataset_statistics(self, table_info: TableInfo) -> DatasetStatistics:
                     ]
                     dataset_statistics.data_size_bytes = bytes_count
                     dataset_statistics.record_count = row_count
+                cursor.execute(f"DESCRIBE HISTORY {escaped_name}")
+                excluded_operations = {
```
mars-lan, line 113: Let's make these CONST and move them  to the top so it's easier to change them later.

mars-lan (commented):
Otherwise LGTM

mars-lan (approved):

PR #698, opened 2023-11-21
Use snowsight compatible url for snowflake dataset mainUrl

usefulalgorithm:
### 🤔 Why?

The old url no longer points to the dataset on snowflake.

### 🤓 What?

Modified the url to point to the snowflake dataset on snowsight.

Note that `account` in the configuration should be the combination of org name and account name:
```
{ORG_NAME}-{ACCOUNT_NAME}
```
But since this format is not enforced in the configuration, yet required for the new URL to work, I made `build_table_url` possible to return a None so that things do not break. I don't know if an account identifier without a `dash` in the middle would work with the old url.

### 🧪 Tested?

Fixed unit test and tested the link.

### ☑️ Checks


metaphor/snowflake/auth.py
```diff
@@ -43,6 +43,10 @@ def have_key_file_or_key_content(self):
 @dataclass(config=ConnectorConfig)
 class SnowflakeAuthConfig(BaseConfig):
     account: str
+    """
+    This is the account identifier, which is defined as `{ORG_NAME}-{ACCOUNT_NAME}`.
```
mars-lan, line 47: Is this a comment for line 45 or 50?

metaphor/snowflake/extractor.py
```diff
@@ -778,11 +778,18 @@ def _init_dataset(
         return dataset
 
     @staticmethod
-    def build_table_url(account: str, full_name: str) -> str:
+    def build_table_url(account: str, full_name: str) -> Optional[str]:
+        tokens = account.split("-")
+        if len(tokens) != 2:
```
mars-lan, line 783: Can `account_name` contains `-`? Also Snowflake account identifier can have cloud region suffix that contains `-` as well: https://docs.snowflake.com/en/user-guide/admin-account-identifier#non-vps-account-locator-formats-by-cloud-platform-and-region

metaphor/snowflake/auth.py
usefulalgorithm: 45.

mars-lan: Is it more pythonic to put the comments after the declaration statement? I thought it's before?

metaphor/snowflake/extractor.py
usefulalgorithm: > Can account_name contains -? 

Doesn't say on the document, but I found a PR that says it's possible to register one account with dash inside.

If it is possible to have dashes in org name then it's basically impossible to know where org name and account name are separated, since both could have dashes.

> Also Snowflake account identifier can have cloud region suffix that contains -

That's the legacy format. I think the correct fix for this ticket probably is to check whether the account is a legacy one, i.e. it's created before organization feature is introduced. If so, we should stick with the old URL; otherwise use the new one.

mars-lan: According to https://docs.snowflake.com/en/user-guide/admin-account-identifier#organization-name:
> The name must start with a letter and can only contain letters (lowercase and uppercase) and numbers. The name cannot contain underscores or other delimiters.

So it's safe to assume that `org_name` can't contain `-`. 

However, based on https://docs.snowflake.com/en/user-guide/admin-account-identifier#account-name, it seems that `account_name` can contain hyphens.

mars-lan: For legacy formats, I think you can split via `.` first and then just process the first segment?

usefulalgorithm: https://docs.snowflake.com/en/sql-reference/sql/create-account#required-parameters

actually I don't think account name allows dashes

metaphor/snowflake/auth.py
```diff
@@ -43,10 +43,8 @@ def have_key_file_or_key_content(self):
 @dataclass(config=ConnectorConfig)
 class SnowflakeAuthConfig(BaseConfig):
     account: str
-    """
-    This is the account identifier, which is defined as `{ORG_NAME}-{ACCOUNT_NAME}`.
-    See https://docs.snowflake.com/en/user-guide/admin-account-identifier for more info.
-    """
+    """This is the account identifier. See https://docs.snowflake.com/en/user-guide/admin-account-identifier for more info."""
```
mars-lan, line 46: For consistency, let's either turn this into a block comment (above the statement), or all others fields in this class into Docstrings?

metaphor/snowflake/extractor.py
```diff
@@ -780,6 +780,22 @@ def _init_dataset(
     @staticmethod
     def build_table_url(account: str, full_name: str) -> str:
         db, schema, table = full_name.upper().split(".")
+        if "-" in account:
+            # New account identifier with org name and account name.
```
mars-lan, line 784: Add link to the doc for posterity?

mars-lan (commented):
Otherwise LGTM

metaphor/snowflake/extractor.py
```diff
@@ -780,6 +780,22 @@ def _init_dataset(
     @staticmethod
     def build_table_url(account: str, full_name: str) -> str:
         db, schema, table = full_name.upper().split(".")
+        if "-" in account:
```
mars-lan, line 783: This logic won't work with legacy format that contains regional locator, e.g. `<account>.us-east-1`.

mars-lan (approved):

PR #699, opened 2023-11-22
The fivetran crawler on demo has not fetched

elic-eon:
### 🤔 Why?

https://docs.pydantic.dev/2.5/errors/validation_errors/#string_type A breaking change was introduced since pydantic 2.5.0. Cause the FiveTran crawler fail to parse every connector data

### 🤓 What?

- Make `pydantic` version fixed
- Update the Fivetran model

### 🧪 Tested?

Unit test

Locally run the crawler

### ☑️ Checks


mars-lan (commented):
Thanks for the quick fix. Does this impact other connectors that parse bool as str as well?

elic-eon:
> Thanks for the quick fix. Does this impact other connectors that parse bool as str as well?

I don't think so, model for `PowerBI` is configured properly. Didn't see other crawler fail in stage

usefulalgorithm (approved):
LGTM!

PR #701, opened 2023-11-22
Add Kafka crawler

usefulalgorithm:
### 🤔 Why?

People want a Kafka crawler, people shall receive a Kafka crawler.

### 🤓 What?

Implemented the Kafka crawler.

### 🧪 Tested?

Tested on personal dev instance.
截圖 2023-11-23 凌晨12 37 25

### ☑️ Checks


metaphor/kafka/extractor.py
```diff
@@ -0,0 +1,91 @@
...
+                            schema_type=schema_type,
+                        ),
+                    )
+                )
+
+        return self._datasets
+
+    @staticmethod
+    def init_admin_client(config: KafkaConfig) -> AdminClient:
+        # FIXME how do we get this to break & exit if we cannot connect?
```
usefulalgorithm, line 86: even `kcat` is not able to do this lol, it doesn't throw or break, just keeps on retrying.

Maybe send a raw HTTP request or sth to the endpoint and check if the response is reasonable, if not then break?

metaphor/kafka/README.md
```diff
@@ -0,0 +1,53 @@
+# Kafka Connector
+
+This connector extracts technical metadata from Kafka. Right now only locally created Kafka cluster has been tested.
```
mars-lan, line 3:
```suggestion
This connector extracts technical metadata from Kafka using [Confluent's Python Client](https://github.com/confluentinc/confluent-kafka-python).
```

metaphor/kafka/config.py
```diff
@@ -0,0 +1,40 @@
...
+class KafkaBootstrapServer(BaseModel):
+    host: str
+    port: int
+
+
+@dataclass(config=ConnectorConfig)
+class KafkaConfig(BaseConfig):
+    schema_registry_url: str
+
+    servers: List[KafkaBootstrapServer] = dataclass_field(default_factory=lambda: [])
```
mars-lan, line 21: Consider renaming this to `bootstrap_servers` or `brokers` to avoid confusion.

metaphor/kafka/config.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # Include or exclude specific databases/schemas/tables
+    filter: DatasetFilter = dataclass_field(default_factory=lambda: DatasetFilter())
```
mars-lan, line 24: Might make more sense to take a list of glob patterns for filtering topic names directly?

metaphor/kafka/extractor.py
```diff
@@ -0,0 +1,91 @@
...
+        for topic in cluster_metadata.topics.keys():
+            if topic not in SYSTEM_GENERATED_TOPICS:
+                logger.info(f"Topic: {topic}")
+                name = topic
+                schema_type = SchemaType.SCHEMALESS
+                raw_schema = None
+                if topic in subjects:
+                    # If there's a subject in schema registry equal to the topic, that's
+                    # the schema we're looking for.
+                    registered_schema = self._schema_registry_client.get_latest_version(
```
mars-lan, line 61: Are you able to find any documentation on the convention for the mapping of topic name <-> schema id? If so, please add the link here for posterity.,

metaphor/kafka/extractor.py
```diff
@@ -0,0 +1,91 @@
...
+    def from_config_file(config_file: str) -> "KafkaExtractor":
+        return KafkaExtractor(KafkaConfig.from_yaml_file(config_file))
+
+    def __init__(self, config: KafkaConfig) -> None:
+        super().__init__(config, "Kafka metadata crawler", Platform.KAFKA)
+
+        self._config = config
+        self._admin_client = KafkaExtractor.init_admin_client(self._config)
+        self._schema_registry_client = KafkaExtractor.init_schema_registry_client(
+            self._config
```
mars-lan, line 41: Consider creating a new dict to hold only schema registry-specific configs?

metaphor/kafka/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
+---
+version: '2'
+
+services:
+  zookeeper:
+    image: confluentinc/cp-zookeeper:7.4.3.arm64
```
mars-lan, line 6: Is ZK still needed? I thought the latest Kafka does away with RAFT? Or maybe that's not available in OSS?

metaphor/kafka/README.md
```diff
@@ -0,0 +1,53 @@
...
+output:
+  file:
+    directory: <output_directory>
+```
+
+See [Output Config](../common/docs/output.md) for more information on `output`.
+
+## Testing
+
+Follow the [Installation](../../README.md) instructions to install `metaphor-connectors` in your environment (or virtualenv).
```
mars-lan, line 45: Need to install `kafka` extra?

pyproject.toml
```diff
@@ -50,13 +51,15 @@ sql-metadata = { version = "^2.8.0", optional = true }
 sqllineage = { version = "~=1.3.8", optional = true }
 tableauserverclient = { version = "^0.25", optional = true }
 thoughtspot_rest_api_v1 = { version = "1.5.3", optional = true }
+types-confluent-kafka = { version = "^1.0.1", optional = true } # Stubs for confluent kafka
```
mars-lan, line 54: Move it to `tool.poetry.dev-dependencies`?

metaphor/kafka/docker-compose.yml
usefulalgorithm: I found someone else's example that has zookeeper, and I saw other people saying it's not needed. When I run this compose manifest it runs OK so I didn't worry about whether this is necessary for the broker to run or not.

usefulalgorithm: I think we don't really want this file to exist when this PR gets merged?

mars-lan: Not really. We'll only set up an integration testing based off this.

metaphor/kafka/extractor.py
usefulalgorithm: I think this logic is actually off, here's the topic <-> schema subject description:
https://docs.confluent.io/cloud/current/sr/fundamentals/serdes-develop/index.html#subject-name-strategy

looks like it's something that confluent users wouldn't really touch usually, and isn't enforced. I think I'll add a new config value for the naming strategy and use that to derive subject name from topics?

metaphor/kafka/docker-compose.yml
usefulalgorithm: I think I'll finish the rest of this PR and see if I can remove zookeeper

metaphor/kafka/extractor.py
usefulalgorithm: how datahub does it:
https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/src/datahub/ingestion/source/confluent_schema_registry.py#L79

pyproject.toml
mars-lan: Remove?

metaphor/kafka/config.py
```diff
@@ -0,0 +1,148 @@
...
+    bootstrap_servers: List[KafkaBootstrapServer] = dataclass_field(
+        default_factory=lambda: []
+    )
+    """
+    The Kafka bootstrap servers / brokers. Cannot be empty.
+    """
+
+    filter: TopicFilter = dataclass_field(default_factory=lambda: TopicFilter())
+
+    topic_naming_strategies: Dict[str, KafkaTopicNamingStrategy] = dataclass_field(
```
mars-lan, line 102: I doubt people will adopt topic-specific strategy? Most likely they'll choose one strategy for all topics, or simply have no strategy, which means they'll need to manually specify the topic <-> schema ID mapping in config.

tests/kafka/expected.json
```diff
@@ -0,0 +1,31 @@
+[
+  {
+    "logicalId": {
+      "name": "foo_1_1",
+      "platform": "KAFKA"
+    },
+    "schema": {
+      "rawSchema": "\"string\"",
```
perepelitca, line 8: Can we return this as already parsed json?
https://github.com/MetaphorData/app/blob/6692d60a0c58c50835d98b3035e7c6afaf67727c/%40www/data/schema.gql#L5290-L5293

usefulalgorithm: this is not guaranteed to be json, protobuf is also a possible format

pyproject.toml
usefulalgorithm: ah yes

metaphor/kafka/config.py
usefulalgorithm: I actually don't know the convention for this. The 3 strategies are from confluent, and the default behavior is `TopicNameStrategy`.

Do you think having a topic <-> schema mapping in the config makes more sense to our user?

usefulalgorithm: > I doubt people will adopt topic-specific strategy.

I think most people probably don't really care, but it is still a possible scenario.

> Most likely they'll choose one strategy for all topics

That's the `default_subject_name_strategy`. I will add more in the README about this.

> simply have no strategy, which means they'll need to manually specify the topic <-> schema ID mapping in config.

That would be the `RECORD_NAME_STRATEGY`, where people can manually map to subjects from a topic.

Also it seems to be common practice for a topic `foo` to have subject `foo-key` or `foo-value`, instead of just named `foo`.

tests/kafka/expected.json
perepelitca: Is it will be used anywhere in UI? How UI would know how to parse it if the format is unkown?

usefulalgorithm:
@mars-lan  lets merge this~

mars-lan (approved):
Great work!

PR #704, opened 2023-11-23
If snowflake table type is not a known values, don't parse it into Dataset

usefulalgorithm:
### 🤔 Why?

We found a table type that isn't supposed to be a valid value, and it's causing the crawler to break.

### 🤓 What?

Only parse the recognizable values.

### 🧪 Tested?

Unit test.

### ☑️ Checks

### DIFF
--- a/metaphor/snowflake/extractor.py
+++ b/metaphor/snowflake/extractor.py
@@ -758,15 +758,20 @@ class SnowflakeExtractor(BaseExtractor):
             main_url=SnowflakeExtractor.build_table_url(self._account, normalized_name)
         )
 
+        sql_schema = None
+        if table_type in [item.value for item in SnowflakeTableType]:
+            sql_schema = SQLSchema(
+                materialization=table_type_to_materialization_type[
+                    SnowflakeTableType(table_type)
+                ]
+            )
+        else:
+            logger.warning(f"Unknown table type: {table_type}")
         dataset.schema = DatasetSchema(
             schema_type=SchemaType.SQL,
             description=comment,
             fields=[],
-            sql_schema=SQLSchema(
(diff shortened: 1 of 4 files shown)


metaphor/snowflake/extractor.py
```diff
@@ -758,15 +758,20 @@ def _init_dataset(
             main_url=SnowflakeExtractor.build_table_url(self._account, normalized_name)
         )
 
+        sql_schema = None
+        if table_type in [item.value for item in SnowflakeTableType]:
```
mars-lan, line 762: Interesting to know what undocumented type this is. Please add a link to the best documentation (if not https://docs.snowflake.com/en/sql-reference/info-schema/tables) to `SnowflakeTableType` for posterity.

mars-lan: Also by not setting `sql_schema` it could lead to exception when it's accessed by later part of the code.

metaphor/snowflake/extractor.py
```diff
@@ -758,15 +758,18 @@ def _init_dataset(
             main_url=SnowflakeExtractor.build_table_url(self._account, normalized_name)
         )
 
+        sql_schema = SQLSchema()
+        if table_type in [item.value for item in SnowflakeTableType]:
+            sql_schema.materialization = table_type_to_materialization_type[
+                SnowflakeTableType(table_type)
+            ]
+        else:
+            logger.warning(f"Unknown table type: {table_type}")
```
mars-lan, line 767: add a blank line after to improve readability?

mars-lan (commented):
Otherwise LGTM.

mars-lan (approved):

PR #705, opened 2023-11-24
Add assetStructure to dbt model views

usefulalgorithm:
### 🤔 Why?

To populate `AssetStructure` for DBT model views.

### 🤓 What?

- Added logic to keep track of model's directories and name.

### 🧪 Tested?

Unit test.

### ☑️ Checks

### DIFF
--- a/metaphor/dbt/artifact_parser.py
+++ b/metaphor/dbt/artifact_parser.py
@@ -27,6 +27,7 @@ from metaphor.dbt.util import (
     to_dataset_entity_id,
 )
 from metaphor.models.metadata_change_event import (
+    AssetStructure,
     ColumnTagAssignment,
     DataMonitorStatus,
     DataPlatform,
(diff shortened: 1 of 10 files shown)


metaphor/dbt/artifact_parser.py
```diff
@@ -452,6 +453,12 @@ def _parse_model(
             return
 
         virtual_view = init_virtual_view(self._virtual_views, model.unique_id)
+
+        virtual_view.structure = AssetStructure(
+            directories=model.original_file_path.split("/")[:-1],
```
mars-lan, line 458: For dbt, I believe we should only have a single-level directory, i.e. the project directory name.

usefulalgorithm: project directory name == model.path ?

I think it should be something else but can't figure out what

metaphor/dbt/artifact_parser.py
```diff
@@ -455,8 +455,12 @@ def _parse_model(
 
         virtual_view = init_virtual_view(self._virtual_views, model.unique_id)
 
+        # Extract project directory from the model's unique id
+        directory = get_model_name_from_unique_id(model.unique_id)[
+            : -len(model.name) - 1
```
mars-lan, line 460: Consider split by `.` instead, which seems a bit more intuitive/robust?

mars-lan (commented):
Otherwise LGTM.

mars-lan (approved):

mars-lan (approved):

alyiwang (commented):
LGTM

PR #708, opened 2023-11-27
Extract tableau workbook tags as SystemTags

usefulalgorithm:
### 🤔 Why?

Tableau workbook can be tagged, our connector needs to support that.

### 🤓 What?

- Implemented parsing logic for the tags on workbooks.
- Modified unit test.

### 🧪 Tested?

- Tested locally.
- Unit test.

### ☑️ Checks


mars-lan:
To show clean diff, you can base your PR off `tsung-julii/sc-22604/add-config-option-to-exclude-tableau-personal` instead.

metaphor/tableau/extractor.py
```diff
@@ -282,6 +288,18 @@ def _parse_dashboard(self, workbook: tableau.WorkbookItem) -> None:
 
         self._dashboards[workbook_id] = dashboard
 
+        if workbook.tags:
+            hierarchy = Hierarchy(
+                logical_id=HierarchyLogicalID(
+                    path=[DashboardPlatform.TABLEAU.value, workbook_id],
+                ),
+                system_tags=SystemTags(
+                    tags=[SystemTag(value=tag) for tag in sorted(workbook.tags)]
```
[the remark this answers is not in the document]
usefulalgorithm: `Hierarchy` already contains the source in `logical_id.path`, I think it should be fine?

metaphor/tableau/extractor.py
```diff
@@ -282,6 +288,18 @@ def _parse_dashboard(self, workbook: tableau.WorkbookItem) -> None:
 
         self._dashboards[workbook_id] = dashboard
 
+        if workbook.tags:
+            hierarchy = Hierarchy(
```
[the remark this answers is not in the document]
usefulalgorithm: - `systemTags` is an entity of only `Hierarchy`, not for `Dashboard` and `VirtualView`.
- `workbook` here actually corresponds to a dashboard.
- It is indeed also possible to tag datasources. Do we want to support that?

metaphor/tableau/extractor.py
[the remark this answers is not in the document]
mars-lan: `systemTagSource `is a required field and currently only supports `SNOWFLAKE` .

metaphor/tableau/extractor.py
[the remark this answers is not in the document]
mars-lan: We need to add `systemTags` to all assets (`Dataset`, `Dashboard`, `VirtualView`).

metaphor/tableau/extractor.py
```diff
@@ -332,13 +343,22 @@ def _parse_custom_sql_table(
         except Exception as e:
             logger.error(f"Unable to parse custom query for {custom_sql_table.id}: {e}")
 
+        system_tags = None
+        if custom_sql_table.column_tags:
```
mars-lan, line 347: SystemTags are only for asset-level?

usefulalgorithm: I thought we need to tag the upstream datasets as well?

mars-lan: Are `column_tags` tags for columns?

mars-lan: Yeah, but the point is what you're setting is table-level tags. I don't think we have models for column-level system tags yet.

usefulalgorithm: I see, I'll remove that then. I thought column level tags are considered table tags as well

PR #719, opened 2023-12-08
Add column stats to unity catalog profiler

usefulalgorithm:
### 🤔 Why?

It is also possible to get column stats from databricks sql client, best if we can support that.

### 🤓 What?

Parses column stats with databricks sql client, and stores the info in `Dataset.field_statistics`.

### 🧪 Tested?

Tested locally.

### ☑️ Checks

### DIFF
--- a/metaphor/unity_catalog/profile/extractor.py
+++ b/metaphor/unity_catalog/profile/extractor.py
@@ -105,9 +105,7 @@ class UnityCatalogProfileExtractor(BaseExtractor):
                 platform=DataPlatform.UNITY_CATALOG,
             ),
         )
-        return self._get_statistics(dataset, table_info)
 
-    def _get_statistics(self, dataset: Dataset, table_info: TableInfo) -> Dataset:
         if table_info.table_type is TableType.VIEW:
             return dataset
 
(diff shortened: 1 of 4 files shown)


metaphor/unity_catalog/profile/extractor.py
```diff
@@ -88,40 +99,92 @@ async def extract(self) -> Collection[ENTITY_TYPES]:
...
     def _init_dataset(self, table_info: TableInfo) -> Dataset:
-        return Dataset(
+        dataset = Dataset(
             logical_id=DatasetLogicalID(
                 name=normalize_full_dataset_name(table_info.full_name),
                 platform=DataPlatform.UNITY_CATALOG,
             ),
-            statistics=self._get_dataset_statistics(table_info=table_info),
         )
+        return self._get_statistics(dataset, table_info)
```
elic-eon, line 108: Consider moving `get_statistics` to the top-level?

usefulalgorithm: I'm just gonna put it in `_init_dataset`, since this is the only place that calls it

PR #722, opened 2023-12-09
Allow overriding the source URLs for Unity Catalog datasets

mars-lan:
### 🤔 Why?

Sometimes it's desirable to set a fixed URL for all the datasets instead of using the ones derived from the hostname, e.g. the users are not able to access the workspace used to extract metadata. 

### 🤓 What?

As titled.

### 🧪 Tested?

### ☑️ Checks


metaphor/unity_catalog/extractor.py
```diff
@@ -183,7 +184,10 @@ def _init_dataset(self, table_info: TableInfo) -> Dataset:
         path = urllib.parse.quote(
             f"/explore/data/{database}/{schema_name}/{table_name}"
         )
-        dataset.source_info = SourceInfo(main_url=f"{self._host}{path}")
+        main_url = (
+            f"{self._host}{path}" if self._source_url is None else self._source_url
+        )
```
alyiwang, line 189:
```suggestion
        main_url = self._source_url or f"{self._host}{path}"
```

alyiwang (commented):
Just to confirm, the request is to use one FIXED URL for all entities, not the URL prefix?

mars-lan:
> Just to confirm, the request is to use one FIXED URL for all entities, not the URL prefix?

Correct.

metaphor/unity_catalog/extractor.py
mars-lan: In general it's a good idea to test for None explicitly instead replying on the default boolean evaluation.

mars-lan:
> > Just to confirm, the request is to use one FIXED URL for all entities, not the URL prefix?
> 
> Correct.

Oh second thought, generalized the config so to allow specifying variables to be replaced by actual catalog, schema, and table names to make it more flexible.

alyiwang (approved):
LGTM

PR #727, opened 2023-12-20
Unity catalog table tag

elic-eon:
### 🤔 Why?

To extract table-level unity catalog tag

### 🤓 What?

- Query `{catalog}.information_schema.table_tags` to get tag information
- Extract tags info to `dataset.schema.tags`
- Factor out SDK-related functions into `utils`

### 🧪 Tested?

Verified end-to-end. Ingested the generated MCE, and it show up in the asset page.

### ☑️ Checks


mars-lan (commented):
Does this mean you need `SELECT` privilege for all catalogs?

elic-eon:
> Does this mean you need `SELECT` privileges for all catalogs?

Did some tests. It will need other privileges.

elic-eon:
Can we catch the exception when fetching tags

mars-lan:
> Can we catch the exception when fetching tags

Ideally we want to limit `unity_catalog` to just metadata privilege. I assume there's no REST APIs to get the tags?

mars-lan:
> > Can we catch the exception when fetching tags
> 
> Ideally we want to limit `unity_catalog` to just metadata privilege. I assume there's no REST APIs to get the tags?

Actually never mind, looks like we need `SELECT` privilege to view the lineage anyway: https://docs.databricks.com/en/data-governance/unity-catalog/data-lineage.html#requirements

> To view the lineage of a table or view, users must have the SELECT privilege on the table or view.

elic-eon:
> > Can we catch the exception when fetching tags
> 
> Ideally we want to limit `unity_catalog` to just metadata privilege. I assume there's no REST APIs to get the tags?

No. The REST APIs not provide tag information

alyiwang (approved):
LGTM
Great work

mars-lan (approved):

PR #729, opened 2023-12-22
Fix snowflake profiler parsing issue with case sensitive datasets

usefulalgorithm:
### 🤔 Why?

The snowflake profiler connector was failing when table / column names are case sensitive. This ticket makes it run without crashing midway through the crawling process.

### 🤓 What?

Instead of relying on case insensitive names in information schema, use `SHOW {TABLES|COLUMNS}` query to get case sensitive names.

### 🧪 Tested?

Tested locally with the following datasets:
- table name: M3_CONSENSUS_AGGREGATE, columns: ID, FOO, BAR
- table name: m3_consensus_aggregate, columns: ID, FoO, Phone Number

The resulting MCE:
```json
{
  "fieldStatistics": {
    "fieldStatistics": [
      {
        "fieldPath": "ID",
        "nonnullValueCount": 0.0,
        "nullValueCount": 0.0
      },
      {
        "fieldPath": "FoO",
        "nonnullValueCount": 0.0,
        "nullValueCount": 0.0
      },
      {
        "fieldPath": "Phone Number",
        "nonnullValueCount": 0.0,
        "nullValueCount": 0.0
      },
      {
        "fieldPath": "ID",
        "nonnullValueCount": 0.0,
        "nullValueCount": 0.0
      },
      {
        "fieldPath": "FOO",
        "nonnullValueCount": 0.0,
        "nullValueCount": 0.0
      },
      {
        "fieldPath": "BAR",
        "nonnullValueCount": 0.0,
        "nullValueCount": 0.0
      }
    ]
  },
  "logicalId": {
    "account": "metaphor-dev",
    "name": "demo_db.public.m3_consensus_aggregate",
    "platform": "SNOWFLAKE"
  }
}
```

This ticket only aims to make the connector run without failing. Will deal with the tables being collapsed into a single table issue in another ticket.

### ☑️ Checks


metaphor/snowflake/profile/extractor.py
```diff
@@ -136,38 +139,35 @@ def _fetch_columns_async(
...
-        for normalized_name, column_info in column_info_map.items():
-            dataset_info = tables.get(normalized_name)
-            assert dataset_info is not None
-
-            profile_queries[normalized_name] = QueryWithParam(
+        for table_name, table_catalog, table_schema in table_tuples:
+            full_name = f"{table_catalog}.{table_schema}.{table_name}"
+            column_info: List[Tuple[str, str]] = []
+
+            cursor.execute(f'SHOW COLUMNS IN "{table_name}"')
```
alyiwang, line 151: Instead of issuing one query per each table (could be 100K tables), it's probably more efficient to use one query `SHOW COLUMNS` and then match columns to tables.

alyiwang (commented):
Does this PR handle the case that column name contains space?

metaphor/snowflake/profile/extractor.py
usefulalgorithm: yeah good call... will do that

usefulalgorithm:
> Does this PR handle the case that column name contains space?

yes it does.

elic-eon:
```
table name: M3_CONSENSUS_AGGREGATE, columns: ID, FOO, BAR
table name: m3_consensus_aggregate, columns: ID, FoO, Phone Number
```

Looks like we are trying to merge two Snowflake table into one metaphor dataset in this case?

usefulalgorithm:
> ```
> table name: M3_CONSENSUS_AGGREGATE, columns: ID, FOO, BAR
> table name: m3_consensus_aggregate, columns: ID, FoO, Phone Number
> ```
> 
> Looks like we are trying to merge two Snowflake table into one metaphor dataset in this case?

Well that's basically the only way to do this for the time being. Ideally they should be considered different tables, but since we always normalize the dataset names to lowercase, there's no way to distinguish between these two.

elic-eon:
Do you test this end-to-end on UI? Not sure how UI will show if we have two columns with the same name

usefulalgorithm:
> Do you test this end-to-end on UI? Not sure how UI will show if we have two columns with the same name

Looks like UI will GG
截圖 2023-12-23 凌晨1 25 53

elic-eon:
Yeah. I don't think we handle duplicate fields

usefulalgorithm:
> Yeah. I don't think we handle duplicate fields

I don't think it's because there are duplicate fields, but that there are columns that do not exist in the dataset.

usefulalgorithm:
@elic-eon getting rid of the duplicate columns in the source table actually solved it:

With the following SQL:
```sql
create table m3_consensus_aggregate (
  id NUMBER,
  "FoO" NUMBER,
  "Phone Number" VARCHAR(20)
);

create table "m3_consensus_aggregate" (
  "FOO" TEXT,
  "Full Name" VARCHAR(100)
);
```
It would work:
截圖 2023-12-23 凌晨1 56 13

elic-eon:
> @elic-eon getting rid of the duplicate columns in the source table actually solved it:
> 
> With the following SQL:
> 
> ```sql
> create table m3_consensus_aggregate (
>   id NUMBER,
>   "FoO" NUMBER,
>   "Phone Number" VARCHAR(20)
> );
> 
> create table "m3_consensus_aggregate" (
>   "FOO" TEXT,
>   "Full Name" VARCHAR(100)
> );
> ```
> 
> It would work: 截圖 2023-12-23 凌晨1 56 13

But there could be duplicate columns, we cannot control that part, right?

usefulalgorithm:
> > @elic-eon getting rid of the duplicate columns in the source table actually solved it:
> > With the following SQL:
> > ```sql
> > create table m3_consensus_aggregate (
> >   id NUMBER,
> >   "FoO" NUMBER,
> >   "Phone Number" VARCHAR(20)
> > );
> > 
> > create table "m3_consensus_aggregate" (
> >   "FOO" TEXT,
> >   "Full Name" VARCHAR(100)
> > );
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It would work: 截圖 2023-12-23 凌晨1 56 13
> 
> But there could be duplicate columns, we cannot control that part, right?

correct.

Should the fix be in `app` or in here? In crawler best we can do is to drop the column if there's another one that already exists

mars-lan:
> > > @elic-eon getting rid of the duplicate columns in the source table actually solved it:
> > > With the following SQL:
> > > ```sql
> > > create table m3_consensus_aggregate (
> > >   id NUMBER,
> > >   "FoO" NUMBER,
> > >   "Phone Number" VARCHAR(20)
> > > );
> > > 
> > > create table "m3_consensus_aggregate" (
> > >   "FOO" TEXT,
> > >   "Full Name" VARCHAR(100)
> > > );
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > It would work: 截圖 2023-12-23 凌晨1 56 13
> > 
> > 
> > But there could be duplicate columns, we cannot control that part, right?
> 
> correct.
> 
> Should the fix be in `app` or in here? In crawler best we can do is to drop the column if there's another one that already exists

Let's tackle the column name normalization in a separate PR.

mars-lan (dismissed):

mars-lan (approved):

PR #733, opened 2023-12-25
Make project ids and job ids optional for DBT cloud crawler

usefulalgorithm:
### 🤔 Why?

There can be a lot of job ids to specify. Instead of having them as required options, it is better to be able to only specify project ids.

### 🤓 What?

Modified the configs and dbt cloud crawler to support parsing a project's last successful run.

Also made the entities a dict, so that if a run was already found we won't send the redundant request.

### 🧪 Tested?

Tested locally.

### ☑️ Checks


mars-lan (commented):
I'm not sure this is the desired behavior. What we want is to specify a project ID, and extract metadata from the last successful runs of *each job*, not *one job*?

usefulalgorithm:
> I'm not sure this is the desired behavior. What we want is to specify a project ID, and extract metadata from the last successful runs of _each job_, not _one job_?

Modified the logic to parse all jobs in a given project id.

metaphor/dbt/cloud/client.py
```diff
@@ -11,6 +11,15 @@
 logger = get_logger()
 
 
+class DbtRun(NamedTuple):
+    run_id: int
+    project_id: int
+    job_id: int
```
alyiwang, line 17: maybe sort alphabetically, or by `project -> job -> run`

metaphor/dbt/cloud/config.py
```diff
@@ -17,13 +17,16 @@ class DbtCloudConfig(BaseConfig):
     service_token: str
 
     # dbt cloud job IDs
-    job_ids: List[int]
+    job_ids: List[int] = dataclass_field(default_factory=list)
+
+    # dbt cloud project IDs
+    project_ids: List[int] = dataclass_field(default_factory=list)
```
alyiwang, line 23: update README also?

metaphor/dbt/cloud/extractor.py
```diff
@@ -43,35 +44,42 @@ def __init__(self, config: DbtCloudConfig):
     async def extract(self) -> Collection[ENTITY_TYPES]:
         logger.info("Fetching metadata from DBT cloud")
 
+        for project_id in self._project_ids:
+            self._job_ids.extend(self._client.get_project_jobs(project_id))
```
alyiwang, line 48: Do we need to dedupe after this?

metaphor/dbt/cloud/extractor.py
```diff
@@ -43,35 +44,42 @@ def __init__(self, config: DbtCloudConfig):
...
+
         for job_id in self._job_ids:
-            await self._extract_job(job_id)
+            await self._extract_last_run(job_id)
+
+        return [item for ls in self._entities.values() for item in ls]
 
-        return self._entities
+    async def _extract_last_run(self, job_id: Optional[int]):
```
alyiwang, line 55: In what situation is `job_id=None`?

alyiwang: And if it's None, should we do early-termination instead of running through the following steps?

alyiwang (approved):
LGTM

metaphor/dbt/cloud/config.py
```diff
@@ -17,13 +17,16 @@ class DbtCloudConfig(BaseConfig):
     service_token: str
 
     # dbt cloud job IDs
-    job_ids: List[int]
+    job_ids: List[int] = dataclass_field(default_factory=list)
```
mars-lan, line 20: @usefulalgorithm please add a validator to ensure that only `job_ids` or `project_ids` is specified, but not both.

PR #734, opened 2023-12-25
Allow partial failures in dbt cloud crawler

usefulalgorithm:
### 🤔 Why?

Allow failed dbt manifest extractions. If during a crawler session a single manifest failed to extract, instead of crashing the entire thing, simply proceed with the rest of the manifests. That being said, we still want the run metadata to show FAILURE so that people actually get notified, so there's a bit of refactor to make this possible.

### 🤓 What?

- Added a try catch block to handle the failed parse.
- Removed unused `run` method in `BaseExtractor`.
- Refactored `run_connector` to take the entire extractor instance instead of only the callback function.
- Have `BaseExtractor` store error messages and stacktraces to reflect failures encountered during crawling.

### 🧪 Tested?

Tested locally.

### ☑️ Checks


metaphor/dbt/cloud/extractor.py
```diff
@@ -85,16 +85,19 @@ async def _extract_last_run(self, job_id: Optional[int], project_id: Optional[in
...
+                    account=account,
+                    docs_base_url=docs_base_url,
+                    output=self._output,
+                    meta_ownerships=self._meta_ownerships,
+                    meta_tags=self._meta_tags,
+                )
+            ).extract()
+            self._entities[run.run_id] = entities
+        except Exception:
+            logger.exception(f"Failed to parse artifacts for run {run}")
```
mars-lan, line 103: Need to output the overall failed status if any of the jobs failed.

mars-lan (commented):
Bump version

metaphor/dbt/cloud/extractor.py
usefulalgorithm: I thought we wanted the crawler to not break if we failed to parse any of the manifest?

If we want the whole thing to crash when a manifest cannot be parsed, I believe this is the current behavior.

mars-lan: What we want is for the connector to run to the end despite the failure, but set the `status` in `run-metadata.json` as `FAILURE`. This way the user will get notified about the errors, while any successful jobs will still get ingested.

usefulalgorithm: understood, I'll modify this then

usefulalgorithm:
@mars-lan  can you take another look? thanks

alyiwang (commented):
Could you document the rest of the changes in the PR description, like the code refactoring you did here?
Otherwise, LGTM

alyiwang (approved):

PR #737, opened 2023-12-27
Upgrade to Python 3.10

usefulalgorithm:
### 🤔 Why?

`snowflake-connector-python` has a `urllib3` dependency that's stuck in v1.x, while some other libraries (e.g. tableuserverclient, minio for s3 crawler) depends on urllib3 v2.x.
In https://github.com/snowflakedb/snowflake-connector-python/issues/1743 they got rid of the urllib3 dependency altogether for python >= 3.10, so it makes sense to upgrade our minimum supported python version from 3.8 to 3.10 in order to be able to upgrade other libraries.

### 🤓 What?

- Change python minimum version in pyproject.toml from 3.8 to 3.10
- Fixed some incompatibilities between different versions
- Added guide to install mssql on apple silicon machines

### 🧪 Tested?

- Unit tests are passing.
- Ran several connectors, all worked as intended.

### ☑️ Checks

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -2,7 +2,7 @@
 name: CD
 
 env:
-  PYTHON_VERSION: 3.10
+  PYTHON_VERSION: '3.10'
   POETRY_VERSION: 1.2.2
   NODE_VERSION: 16
   DOCKER_IMAGE: metaphordata/connectors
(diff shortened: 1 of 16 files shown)


usefulalgorithm:
Actually this is probably not necessary... closing this

mars-lan:
> Actually this is probably not necessary... closing this

Yeah prefer to keep python 3.8 for another 6 months till near EOL for maximum compatibility.

PR #738, opened 2023-12-27
S3 crawler

usefulalgorithm:
### 🤔 Why?

Adds support for parsing S3 objects as datasets. Specifically, the following scenarios are implemented:
- Parsing files to datasets
- Parsing directories to datasets
    - this is done by finding the newest file within the directory, and using that as the source of truth
    - to do this, specify `{table}` label in the path specification: `s3://.../.../{table}/*/*.csv`
    - It is also possible to specify other labels to be included in the dataset name. For example, `s3://{dept}/*/{name}/*/{table}/*.csv` will have the crawler name all parsed datasets as `DEPT/NAME/TABLE`
- Parsing partitioned tables to datasets
    - specify the path spec as `s3://.../.../{table}/{partition_key[0]}={partition[0]/*.parquet` to parse partitioned tables as datasets. If there's no partition key, then write `s3://.../.../{table}/{partition[0]}/*.parquet`.

### 🤓 What?

- Implementation
- Readme
- Unit test

### 🧪 Tested?

- Added unit test with actual minio storage
- Tested end to end

截圖 2024-01-02 晚上11 08 51

This is done with https://github.com/MetaphorData/app/pull/7485

### ☑️ Checks


metaphor/common/logger.py
```diff
@@ -24,6 +24,7 @@ def get_logger() -> logging.Logger:
     logger = logging.getLogger("metaphor")
     logger.setLevel(logging.DEBUG)
     logger.addHandler(console)
+    console.setLevel(logging.DEBUG)
```
usefulalgorithm, line 27: no

metaphor/s3/table_data.py
```diff
@@ -0,0 +1,169 @@
...
+    @property
+    def guid(self) -> str:
+        return self.table_path
+
+    @property
+    def logical_id(self) -> DatasetLogicalID:
+        return DatasetLogicalID(
+            name=yarl.URL(self.table_path)
+            .with_scheme("")
+            .human_repr()[1:],  # Drop the first slash
```
elic-eon, line 68: What does this transformation do? Drop the `s3://` and first slash?

tests/s3/test_table_data.py
```diff
@@ -0,0 +1,78 @@
...
+
+    assert merged == TableData(
+        display_name="foo",
+        full_path="bar",
+        partitions=None,
+        timestamp=datetime(2024, 1, 1, 0, 0, 0, 0),
+        table_path="foo",
+        size_in_bytes=5678,
+        number_of_files=1,
+    )
```
elic-eon, line 37: Why merged `TableData` have different `full_path` and `table_path`?

usefulalgorithm: full_path = path to the file
table_path = path to the dataset

a dataset can be a single file or a directory containing other files / directories.

metaphor/s3/table_data.py
usefulalgorithm: yes.

metaphor/s3/path_spec.py
```diff
@@ -0,0 +1,305 @@
...
+        return self.url.with_scheme("").human_repr()[1:]  # Drop the first slash
+
+    @cached_property
+    def path_prefix(self) -> str:
+        """
+        Returns the prefix of the path before the label level. E.g.
+        for uri = `s3://foo/bar/baz/*/*/{table}/*/*.*`, this would be
+        `bar/baz/*/*/`.
+        """
+        index = re.search(r"[\*|\{]", self.object_path)
```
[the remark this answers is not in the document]
usefulalgorithm: the comment was incorrect.

metaphor/s3/extractor.py
```diff
@@ -0,0 +1,197 @@
...
+                folder1 = folder1.rsplit("=", 1)[-1]
+                folder2 = folder2.rsplit("=", 1)[-1]
+
+        num_folder1 = int(folder1)
+        num_folder2 = int(folder2)
+        if num_folder1 == num_folder2:
+            return 0
+        else:
+            return 1 if num_folder1 > num_folder2 else -1
+    except Exception:
```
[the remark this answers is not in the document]
usefulalgorithm: `num_folder1 = int(folder1)` can throw

metaphor/s3/extractor.py
```diff
@@ -0,0 +1,197 @@
...
+        self._path_specs = config.path_specs
+
+    def _resolve_templated_folders(
+        self, bucket_name: str, path_prefix: str
+    ) -> Iterable[str]:
+        folder_split: List[str] = path_prefix.split("*", 1)
+        # If the len of split is 1 it means we don't have * in the prefix
+        if len(folder_split) == 1:
+            yield path_prefix
+            return
```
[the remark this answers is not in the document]
usefulalgorithm: actually no, we have to explicitly return after yielding here

metaphor/s3/extractor.py
```diff
@@ -0,0 +1,197 @@
...
+        iterator = list_folders(
+            bucket_name=bucket_name,
+            prefix=folder,
+            config=self._config,
+        )
+        iterator = peekable(iterator)
+        if iterator:
+            sorted_dirs = sorted(
+                iterator,
+                key=functools.cmp_to_key(
```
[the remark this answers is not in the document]
usefulalgorithm: can't think of a better way to compare two directories...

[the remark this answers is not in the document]
elic-eon: nvm, use `cmp_to_key` is correct.

tests/s3/test_table_data.py
elic-eon: Merging two `TableData` with different `table_path` didn't make sense in the crawler? So this is only for the test case?

My real question is the logic looks different between `test_merge_table_data` and `test_merge_table_data_with_partitions`.

usefulalgorithm: I see what you mean now. Yes, in a real scenario we would never merge two table datas with different table paths. The objects here are only for testing.

I will add more comment in the merge method, see if it makes more sense.

elic-eon (approved):
LGTM

PR #741, opened 2024-01-02
Sort the snowflake tags so that inherited tags can be overriden

usefulalgorithm:
### 🤔 Why?

Sort the tags returned by querying the tag references so that it's possible to override inherited tags.

See https://docs.snowflake.com/en/user-guide/object-tagging#tag-lineage.

### 🤓 What?

Added a sort clause to the query.

### 🧪 Tested?

Tested locally.

### ☑️ Checks


mars-lan (commented):
Bump up version?

mars-lan (commented):
Should we also write to the new `system_tags` field?

mars-lan (approved):

PR #745, opened 2024-01-07
Update README.md

mars-lan:
### 🤓 What?

Improve the README file for Kafka.

### DIFF
--- a/metaphor/kafka/README.md
+++ b/metaphor/kafka/README.md
@@ -4,28 +4,7 @@ This connector extracts technical metadata from Kafka using [Confluent's Python 
 
 ## Setup
 
-To run a Kafka cluster locally, follow the instructions below:
-
-1. Start a Kafka cluster (broker + schema registry + REST proxy) locally via docker-compose:
-```shell
-$ docker-compose --file metaphor/kafka/docker-compose.yml up -d
-```
-  - Broker is on port 9092.
-  - Schema registry is on port 8081.
-  - REST proxy is on port 8082.
-2. Find the cluster ID:  
-```shell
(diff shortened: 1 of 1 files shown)


metaphor/kafka/README.md
```diff
@@ -4,76 +4,55 @@ This connector extracts technical metadata from Kafka using [Confluent's Python
...
-3. Register a new topic via the REST proxy:
-```shell
-curl -X POST -H "Content-Type: application/json" http://localhost:8082/v3/clusters/<YOUR CLUSTER ID>/topics -d '{"topic_name": "<YOUR TOPIC NAME>"}'| jq .
-```
-4. Register a schema to the registry:
-```shell
-curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" --data '{"schema": <SCHEMA AS STRING>}' http://localhost:8081/subjects/<YOUR TOPIC NAME>-<key|value>/version
-```
-  - It is possible to have schema with name different to the topic. See `Topic <-> Schema Subject Mapping` section below for more info.
+If [ACL](https://docs.confluent.io/platform/current/security/rbac/authorization-acl-with-mds.html) is enabled, the credentials used by the crawler must be allowed to perform [Describe operation](https://docs.confluent.io/platform/current/kafka/authorization.html#topic-resource-type-operations) on the topics of interest.
```
mars-lan, line 7: @usefulalgorithm please verify if `Describe` for Cluster resource is sufficient, or if there's an easy way to specify ACL for `*` Topic resource.

usefulalgorithm: The Kafka admin client only does `ListTopics`, `Describe` permission should be enough. Setting up ACL in my local environment is not so straightforward though, so I didn't set it up.

usefulalgorithm (approved):
lgtm

PR #747, opened 2024-01-08
Parse table and column stats for hive

usefulalgorithm:
### 🤔 Why?

Parse table and column statistics for hive tables.

### 🤓 What?

- Added logic for parsing table and column stats.
- Added config value for collecting these stats.
- Unit test.

### 🧪 Tested?

Unit test.

### ☑️ Checks


mars-lan (commented):
Please also update README with requirements on privileges.

metaphor/hive/README.md
```diff
@@ -13,8 +13,16 @@ host: <host>
 port: <port>
 auth_user: <auth user for hiveserver>
 password: <password for the auth user>
+
+output:
+  file:
+    directory: <output_directory>
 ```
 
+If `auth_user` and `password` are provided, ensure the specified user has `SELECT` privilege. See [https://cwiki.apache.org/confluence/display/Hive/SQL+Standard+Based+Hive+Authorization](https://cwiki.apache.org/confluence/display/Hive/SQL+Standard+Based+Hive+Authorization) for more information.
```
mars-lan, line 22: Add a `Setup` section like others. Also is do you need `SELECT` privilege for ALL tables, or some. Do you need that even for SHOW & DESCRIBE?

usefulalgorithm: >  Do you need that even for SHOW & DESCRIBE?

Yes, see the table in the `Privileges Required for Hive Operations` section in the linked page.

mars-lan: Alos use `SQL Standard Based Hive Authorization` as the title of the link.

mars-lan: Looks like we also need `WITH GRANT OPTION`: https://cwiki.apache.org/confluence/display/Hive/SQL+Standard+Based+Hive+Authorization#:~:text=Y%20(input%20location)-,SHOW%20CREATE%20TABLE,-Y%2BG

metaphor/hive/README.md
```diff
@@ -11,10 +15,31 @@ Create a YAML config file based on the following template.
...
+```
+
+See [Output Config](../common/docs/output.md) for more information on `output`.
+
+### Optional Configurations
+
+#### Authenticated User
+
+```yaml
 auth_user: <auth user for hiveserver>
```
mars-lan, line 31: Chances are these will be required for any reasonable production deployment?

metaphor/hive/extractor.py
```diff
@@ -85,8 +95,87 @@ def _extract_table(
...
+                data_size_bytes=table_size,
+                record_count=num_rows,
+            )
+            field_statistics: List[FieldStatistics] = []
+            for field in fields:
+                cursor.execute(
+                    f"describe formatted {database}.{table} {field.field_path}"
+                )
+                stats_col_names = {
+                    "num_nulls": "nullValueCount",
```
mars-lan, line 141: What bout avg & std dev (which are expensive to compute)? Maybe we should have a config like https://github.com/MetaphorData/connectors/blob/main/metaphor/common/docs/column_statistics.md?

usefulalgorithm: Hive does not compute those AFAIK. Don't think we want to compute these by hand?

metaphor/hive/README.md
usefulalgorithm: Yes. How should I write this?

metaphor/hive/extractor.py
usefulalgorithm: There are indeed functions that calculate avg and std. Added those and checked that they work.

mars-lan: std & avg are expensive to calculate so I'd suggest giving users the choice to selectively enable each like https://github.com/MetaphorData/connectors/tree/main/metaphor/snowflake/profile#column-statistics

mars-lan (approved):

Issue #621, opened 2023-10-10
Breaking changes in 0.14


mars-lan:
- https://github.com/MetaphorData/connectors/pull/854
- https://github.com/MetaphorData/connectors/pull/857

PR #749, opened 2024-01-09
Filter dbt cloud jobs by environment id

usefulalgorithm:
### 🤔 Why?

To filter collected dbt jobs by their environment ID.

### 🤓 What?

If user specified a list of environment ids to include, and the crawler sees a job with an environment id not in that list, then the job will be ignored.

### 🧪 Tested?

- Unit test.
- Tested locally.

### ☑️ Checks

### DIFF
--- a/metaphor/dbt/cloud/client.py
+++ b/metaphor/dbt/cloud/client.py
@@ -26,10 +26,17 @@ class DbtAdminAPIClient:
     See https://docs.getdbt.com/dbt-cloud/api-v2 for more details.
     """
 
-    def __init__(self, base_url: str, account_id: int, service_token: str):
+    def __init__(
+        self,
+        base_url: str,
+        account_id: int,
+        service_token: str,
+        included_env_ids: Set[int] = set(),
+    ):
         self.admin_api_base_url = f"{base_url}/api/v2"
         self.account_id = account_id
         self.service_token = service_token
+        self.included_env_ids = included_env_ids
 
     def _get(self, path: str, params: Optional[Dict] = None):
         url = f"{self.admin_api_base_url}/accounts/{self.account_id}/{path}"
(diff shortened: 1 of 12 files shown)


usefulalgorithm:
Yikes, merged another PR into this one and now there's a lot of changes :/

mars-lan:
> Yikes, merged another PR into this one and now there's a lot of changes :/

Rebase?

mars-lan:
> Yikes, merged another PR into this one and now there's a lot of changes :/

Argh I see. You didn't create https://github.com/MetaphorData/connectors/pull/750 against main.

metaphor/dbt/cloud/config.py
```diff
@@ -22,6 +22,9 @@ class DbtCloudConfig(BaseConfig):
     # dbt cloud project IDs
     project_ids: List[int] = dataclass_field(default_factory=list)
```
mars-lan, line 23: Consider changing `project_ids` & `job_ids` to sets as well in a follow-up PR?

metaphor/dbt/cloud/client.py
```diff
@@ -71,6 +78,15 @@ def get_project_jobs(self, project_id: int) -> List[int]:
             jobs |= new_jobs
             offset += page_size
 
+    def job_is_included(self, job_id: int) -> bool:
```
mars-lan, line 81:
```suggestion
    def is_job_included(self, job_id: int) -> bool:
```

metaphor/dbt/cloud/extractor.py
```diff
@@ -53,6 +54,10 @@ async def extract(self) -> Collection[ENTITY_TYPES]:
         return [item for ls in self._entities.values() for item in ls]
 
     async def _extract_last_run(self, job_id: int):
+        if not self._client.job_is_included(job_id):
```
mars-lan, line 57: I wonder if we already have the env_id from `self._client.get_project_jobs` so we can avoid making this extra call per job?

metaphor/dbt/cloud/config.py
usefulalgorithm: I can do that in this PR

metaphor/dbt/cloud/extractor.py
usefulalgorithm: we can't skip this call, since we might have jobs not specified by the project ids, and we still need to check if those are included by the env ids

mars-lan (approved):

metaphor/dbt/cloud/extractor.py
mars-lan: We should make `project_ids` required and using `job_ids` & `evnironment_ids` as filters in https://github.com/MetaphorData/connectors/issues/621

mars-lan (approved):

PR #754, opened 2024-01-11
Add configs to ignore MC timeout errors by default

mars-lan:
### 🤔 Why?

When MC monitors time out, it is treated as an error, which can lead to confusion.

### 🤓 What?

As titled.

### 🧪 Tested?

### ☑️ Checks

### DIFF
--- a/metaphor/monte_carlo/config.py
+++ b/metaphor/monte_carlo/config.py
@@ -1,4 +1,5 @@
-from typing import Optional
+from dataclasses import field
+from typing import List, Optional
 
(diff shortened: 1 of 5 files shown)


alyiwang (commented):
Do we have a test to cover `Monitor timed out`?

mars-lan:
> Do we have a test to cover `Monitor timed out`?

Yes. I used a different error message to test the config as well.

alyiwang (approved):
LGTM

PR #757, opened 2024-01-12
Create scorecard.yml

mars-lan:
### 🤓 What?

To add support for [OpenSSF Scorecard](https://github.com/ossf/scorecard).


usefulalgorithm (approved):
go go go

PR #763, opened 2024-01-24
Make `output` optional in configuration

usefulalgorithm:
### 🤔 Why?

Provide a reasonable default output directory when the `output` field isn't provided. This hopefully creates a more seamless user experience.

### 🤓 What?

- Made `output` optional
- When config file does not have a `output` field, `from_yaml_file` method manually adds it.

### 🧪 Tested?

Tested locally.
### ☑️ Checks

### DIFF
--- a/metaphor/common/base_config.py
+++ b/metaphor/common/base_config.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 from typing import Optional, Type, TypeVar
 
 import yaml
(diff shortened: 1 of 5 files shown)


tests/common/test_config.py
```diff
@@ -24,8 +24,9 @@ def test_yaml_config(test_root_dir):
 
 
 def test_yaml_config_with_missing_config(test_root_dir):
-    with pytest.raises(ValidationError):
-        BaseConfig.from_yaml_file(f"{test_root_dir}/common/configs/missing.yml")
+    missing = BaseConfig.from_yaml_file(f"{test_root_dir}/common/configs/missing.yml")
```
elic-eon, line 27: change the test name to `test_missing_output_config` and rename `missing.yml` to `missing_output.yml`?

metaphor/common/base_config.py
```diff
@@ -27,10 +28,33 @@ class BaseConfig:
...
+    """
 
     @classmethod
     def from_yaml_file(cls: Type[T], path: str) -> T:
         with open(path, encoding="utf8") as fin:
             obj = yaml.safe_load(fin.read())
+
+            # So that user can just ignore this field in their config file.
+            if "output" not in obj:
+                obj["output"] = None
```
elic-eon, line 48: Why not new a `FileSinkConfig` here instead of creating new instance each time when calling the `file_sink_cinfig` prop

tests/common/test_config.py
usefulalgorithm: ok

elic-eon (approved):
LGTM!

mars-lan (commented):
Please also simplify all the README files by remove the example `output` config, e.g. https://github.com/MetaphorData/connectors/blob/main/metaphor/bigquery/README.md?plain=1#L78-L83

metaphor/common/base_config.py
```diff
@@ -28,9 +29,27 @@ class BaseConfig:
     """
 
     output: OutputConfig
+    """
+    No default value here, otherwise dataclass inheritance would not work.
```
mars-lan, line 33: So `pydantic.Field` doesn't work in this case?

usefulalgorithm: no

usefulalgorithm: If I do it like this:
截圖 2024-01-26 下午6 23 31

it would be:
截圖 2024-01-26 下午6 23 08

mars-lan: What I meant is using a factory method to create the default value for `output` so you don't need to make it optional:  https://docs.pydantic.dev/latest/concepts/fields/#default-values

PR #766, opened 2024-01-26
Make tableau connector compatible with older versions of Tableau Server

mars-lan:
### 🤔 Why?

When running against older version of Tableau Server (e.g. `2021.4`), the connector throws the following error:
```
Error when querying workbooksConnection: [{'message': "Validation error of type FieldUndefined: Field 'projectLuid' in type 'Workbook' is undefined @ 'workbooksConnection/nodes/projectLuid'", 'locations': [{'line': 7, 'column': 7, 'sourceName': None}], 'description': "Field 'projectLuid' in type 'Workbook' is undefined", 'validationErrorType': 'FieldUndefined', 'queryPath': ['workbooksConnection', 'nodes', 'projectLuid'], 'errorType': 'ValidationError', 'extensions': None, 'path': None}]
17:11:26  2024-01-22 15:11:26:ERROR:metaphor:'NoneType' object is not subscriptable
```

This is because the `projectLuid` field is not available on older versions:

Screenshot 2024-01-25 at 4 37 20 PM

### 🤓 What?

Extract workbook's project ID from Tableau [REST API](https://tableau.github.io/server-client-python/docs/api-ref#workbookitem-class) instead.

### 🧪 Tested?

Ran against a production instance and verified that there's no change in the MCEs.

### ☑️ Checks


usefulalgorithm (approved):
lgtm

PR #767, opened 2024-01-26
Use pytest-testmon to speed up precommit hook

mars-lan:
### 🤔 Why?

Some of the docker-based tests take too long to run. [pytest-testmon](https://github.com/tarpas/pytest-testmon/) runs only tests affected by the commit, making pre-commit snappy again.

### 🤓 What?

As titled.

### 🧪 Tested?

Verified locally.


usefulalgorithm (approved):
nice

PR #769, opened 2024-01-26
Add delimiter after key prefix when matching dataset keys

usefulalgorithm:
### 🤔 Why?

Need to make sure the key prefix for tag is the entire name, i.e. if it's a database tag, then the tag's key prefix should represent the entire database name instead of a prefix of the name.

### 🤓 What?

Fix the aforementioned bug.

### 🧪 Tested?

Tested locally.

### ☑️ Checks


mars-lan (approved):
Thanks for the quick fix!

PR #775, opened 2024-02-01
Emit system tags for datasets, virtual views, dashboards and metrics

usefulalgorithm:
### 🤔 Why?

Now that `system_tags` is being shown by UI, we can just use that instead of each asset aspect's specific tag fields.

### 🤓 What?

Emit tags thru `system_tags` field in each asset.

### 🧪 Tested?

Updated unit test.

### ☑️ Checks


elic-eon (commented):
Can you also test it by ingesting the generated MCE and see if we are showing the tag correctly?

usefulalgorithm:
> Can you also test it by ingesting the generated MCE and see if we are showing the tag correctly?

Verified, the tags are indeed showing correctly.

mars-lan (approved):

mars-lan (approved):

PR #776, opened 2024-02-02
[sc-24431] Replace FileSink with StreamSink

usefulalgorithm:
### 🤔 Why?

In order to reduce crawler memory usage, instead of having all parsed MCE hogging up memory it's perhaps more memory efficient to pipe a MCE to the file once it's been parsed. In order to do that, the sink has to support writing a single MCE at a time.

The new `StreamSink` has the exact same chunking mechanism as `FileSink`, but it has to be context managed:
```python
with StreamSink(config, metadata) as sink:
    for event in events: # TODO: in the near future `events` will be replaced by `extractor.run_async()`, which is a iterator that yields a MCE once at a time.
        sink.write_event(event)
```

When the context ends, the sink finalizes the last batch file, and writes the execution logs and metadata.

### 🤓 What?

- Implemented new `StreamSink` class to support piped MCEs from crawler classes.
- Removed `Sink` ABC and `FileSink`.
- Modified unit test.

### 🧪 Tested?

Tested on personal dev env with Snowflake crawler:

crawler output:
截圖 2024-02-02 下午6 18 42
ingestion logs: 
[live-tail-results.csv](https://github.com/MetaphorData/connectors/files/14138755/live-tail-results.csv)

All 427 MCEs are ingested successfully.

### ☑️ Checks


usefulalgorithm:
looks like S3 simply does not allow append:
https://stackoverflow.com/questions/41783903/append-data-to-an-s3-object

usefulalgorithm:
figured out a new way to do this, closing this PR

PR #777, opened 2024-02-02
[sc-24431] Create generator based extract method specifically for `QueryLogs`

usefulalgorithm:
### 🤔 Why?

It is possible for large amount of query logs to exhaust crawler's memory in the serverless instance. To mitigate this issue, I created a route for connectors to extract query logs as an iterator, and the corresponding method in `FileSink` to export those parsed query logs one by one. The chunking is taken care of in `FileSink`.

Note: the boundary check for `batch_size_bytes` is not 100% precise, but should be close enough.

### 🤓 What?

- Implemented `collect_query_logs` method in `BaseExtractor` and `SnowflakeExtractor`
- Implemented `QueryLogSink` in `file_sink.py` to support piping an iterator of `QueryLog` to files.
- Modified unit test.

### 🧪 Tested?

Tested on personal dev env with Snowflake crawler:

crawler output:
截圖 2024-02-03 凌晨3 39 48
ingestion logs:
截圖 2024-02-03 凌晨3 43 21

Notice the `query_logs-0.json` is correctly ingested.

### ☑️ Checks

### DIFF
--- a/metaphor/common/base_extractor.py
+++ b/metaphor/common/base_extractor.py
@@ -1,7 +1,7 @@
 import asyncio
 import traceback
 from abc import ABC, abstractmethod
-from typing import Collection, Generator, Optional
+from typing import Collection, Iterator, Optional
 
 from metaphor.common.base_config import BaseConfig
 from metaphor.common.event_util import ENTITY_TYPES
@@ -38,7 +38,7 @@ class BaseExtractor(ABC):
     def run_async(self) -> Collection[ENTITY_TYPES]:
         return asyncio.run(self.extract())
 
-    def collect_query_logs(self) -> Generator[QueryLog, None, None]:
+    def collect_query_logs(self) -> Iterator[QueryLog]:
         """
         Collects only the query logs. By default collects nothing,
         crawler class needs to implement this method if it wishes
--- a/metaphor/common/base_extractor.py
+++ b/metaphor/common/base_extractor.py
@@ -1,11 +1,12 @@
 import asyncio
 import traceback
 from abc import ABC, abstractmethod
-from typing import Collection, Optional
+from typing import Collection, Generator, Optional
 
 from metaphor.common.base_config import BaseConfig
 from metaphor.common.event_util import ENTITY_TYPES
 from metaphor.models.crawler_run_metadata import Platform, RunStatus
+from metaphor.models.metadata_change_event import QueryLog
 
 
 class BaseExtractor(ABC):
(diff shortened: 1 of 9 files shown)


metaphor/common/base_extractor.py
```diff
@@ -37,6 +38,14 @@ def status(self) -> RunStatus:
     def run_async(self) -> Collection[ENTITY_TYPES]:
         return asyncio.run(self.extract())
 
+    def collect_query_logs(self) -> Generator[QueryLog, None, None]:
```
mars-lan, line 41: `Iterator` works too.

metaphor/snowflake/extractor.py
```diff
@@ -527,7 +528,7 @@ def _fetch_tags(self, cursor: SnowflakeCursor) -> None:
                     key_prefix, key, value, column if object_type == "COLUMN" else None
                 )
 
-    def _fetch_query_logs(self) -> None:
+    def _fetch_query_logs(self):
```
mars-lan, line 531: Update return type?

metaphor/common/file_sink.py
```diff
@@ -47,6 +49,79 @@ class FileSinkConfig:
...
+            json.dumps(self.mces),
+        )
+        self.completed_batches += 1
+        self.mces_count = 0
+        self.mces.clear()
+        self.batch_bytes = 0
+
+    def write_query_log(self, query_log: QueryLog) -> None:
+        if not self._entered:
+            raise ValueError()
```
mars-lan, line 109: Add a meaningful error message like "must be used in a with statement"?

metaphor/common/file_sink.py
```diff
@@ -47,6 +49,79 @@ class FileSinkConfig:
...
+        self,
+        path: str,
+        storage: BaseStorage,
+        batch_size_count: int,
+        batch_size_bytes: int,
+        logs_per_mce: int,
+    ) -> None:
+        self.path = path
+        self.storage = storage
+        self.mces_per_batch = batch_size_count
```
mars-lan, line 63: I don't think we need to overcomplicate the chunking. We can stick up to `batch_size_count` query logs in a single MCE up to `bach_size_bytes`. Each output file will therefore contain only 1 MCE.

mars-lan (commented):
Great work. A few comments.

metaphor/common/file_sink.py
usefulalgorithm: I was trying to preserve the original behavior, there is a `DEFAULT_QUERY_LOG_OUTPUT_SIZE` that controls how many query logs can we stuff in a single MCE. `batch_size_count` then controls how many MCEs can we stuff in a single file.

usefulalgorithm: I think the current implementation makes more sense to me though, or we'd have to find another way to tell users that `batch_size_count` means different things for different targets:
- if the crawler you're using does not separate query logs from other things, then `batch_size_count` controls how many MCEs you can fit in a single file
- if your crawler separates query log collection from other things, then for the other things it's how many MCEs are going to be in a file, whereas for query logs it's how many logs you can stuff in a single MCE (and there's going to be 1 MCE only for each of the files)

I guess for me the less explaining we have to do the better. If my implementation in this PR is legit I think it's better to just keep it as is.

@mars-lan  wdyt?

mars-lan (approved):

metaphor/common/file_sink.py
mars-lan: Yeah this is a bit tricky. Ideally we want to have a separate file format for the query logs but for now let's live with the legacy.

PR #791, opened 2024-03-08
refresh time missing for snowflake tables

elic-eon:
### 🤔 Why?

Some of the Snowflake tables are missing refresh time because the loop to get table info stops iterating if there is an error in `get_table_info`.

### 🤓 What?

- Reduce query compilation error ( skip secure view )
- Add try-catch in the for loop
- Partition table by schema because it's likely we can't get table information under a certain schema.

### 🧪 Tested?

- Add tests

### ☑️ Checks

Run crawler against testing Snowflake account


usefulalgorithm (approved):

alyiwang (dismissed):
LGTM

alyiwang (approved):

PR #793, opened 2024-03-11
Set user_id field in BigQuery query log for service accounts

mars-lan:
### 🤔 Why?

When encountering queries issued by a [GCP service account](https://cloud.google.com/compute/docs/access/service-accounts), we should identify it as such by setting the `user_id` field instead of `email`.

### 🤓 What?

As titled.

### 🧪 Tested?

### ☑️ Checks


alyiwang (approved):
LGTM

alyiwang (approved):

PR #794, opened 2024-03-13
Exceptions thrown in crawler constructors not handled gracefully

elic-eon:
### 🤔 Why?

We can't initialize the class before entering `run_cli` or we can't catch the exceptions

### 🤓 What?

- Revert the change to passing a function to `run_cli`

### 🧪 Tested?

- Add unit tests
- Modify snowflake connector to throw exception locally and able to get the output files

### ☑️ Checks

### DIFF
--- a/metaphor/common/runner.py
+++ b/metaphor/common/runner.py
@@ -24,7 +24,7 @@ def run_connector(
     Parameters
     ----------
     connector_func : Callable[[], Collection[ENTITY_TYPES]]
-        The function to create the instance of connector and exec run_async
+        The function to create and return an instance of connector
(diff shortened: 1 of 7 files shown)


metaphor/common/runner.py
```diff
@@ -13,7 +13,7 @@
 
 
 def run_connector(
-    connector: BaseExtractor,
+    connector_func: Callable[[], BaseExtractor],
```
usefulalgorithm, line 16: call it `make_container` or something?

elic-eon: SG. Updated

PR #795, opened 2024-03-13
[sc-25097] Add `column_statistics` field to postgresql.profile config

usefulalgorithm:
### 🤔 Why?

Some statistics cost a lot of resources to calculate, it's better to give user the ability to decide what and what not to calculate.

This is mostly to improve AWS Redshift profiling crawler performance.

Also fixes the llama-index dependency versions.

### 🤓 What?

- Add `column_statistics` field to `postgresql.profile` config.
- Fix `llama-index` versions to latest
  - If I don't do this it would not resolve to the correct versions, and some imports get messed up

### 🧪 Tested?

- Unit tests modified and passed
- Tested locally, with the default being not calculating unique counts and average:
截圖 2024-03-14 凌晨2 36 04

### ☑️ Checks


mars-lan (approved):
Thanks for the quick fix!

PR #797, opened 2024-03-14
Optimize queries used in Snowflake lineage connector

mars-lan:
### 🤔 Why?

It takes an exceedingly long time (> 1.5 hours) to run the Snowflake lineage crawler against large `Snowflake.AccountUnage.QUERY_HISTORY` views.

Turns out that [similar to `ACCESS_HISTORY`](https://docs.snowflake.com/en/user-guide/access-history#querying-the-access-history-view), we must also specify a filter against `START_TIME` in order to query `QUERY_HISTORY` efficiently. 

### 🤓 What?

- Specify the `START_TIME` filter for `QUERY_HISTORY` and use inner JOIN when joining with  `ACCESS_HISTORY` in Snowflake lineage connector
- Qualify the filters in the base Snowflake connector to avoid confusion.

### 🧪 Tested?

Verified that the MCEs before & after the changes are identical. Observed an order of magnitude improvement in performance:

Before:
```
Ended running with RunStatus.SUCCESS at 2024-03-14 08:54:28.949943, fetched 78 entities, took 217.1s
```

After:
```
Ended running with RunStatus.SUCCESS at 2024-03-14 08:47:54.532785, fetched 78 entities, took 27.6s
```

### ☑️ Checks

### DIFF
--- a/metaphor/snowflake/extractor.py
+++ b/metaphor/snowflake/extractor.py
@@ -743,8 +743,8 @@ class SnowflakeExtractor(BaseExtractor):
                 JOIN SNOWFLAKE.ACCOUNT_USAGE.ACCESS_HISTORY a
                   ON a.QUERY_ID = q.QUERY_ID
                 WHERE EXECUTION_STATUS = 'SUCCESS'
-                  AND START_TIME > %s AND START_TIME <= %s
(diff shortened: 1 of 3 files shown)


alyiwang (approved):
LGTM

PR #813, opened 2024-04-01
[sc-25545] Exclude personal space dashboards for tableau

usefulalgorithm:
### 🤔 Why?

For the Workbook class returned by the rest api, its `project_name` would be `None` instead of `Personal Space` if it's from the personal space. To address that, we always use the project name returned by the metadata api to check.

Also fixes https://app.shortcut.com/metaphor-data/story/25549/make-config-for-including-tableau-projects, logic for including / excluding specific project based on id is implemented in this PR as well.

### 🤓 What?

- Excludes dashboards from personal space properly.
- Adds logic for including & excluding specific project based on its id.

### 🧪 Tested?

- Modified unit test.
- Tested locally:

```
2024-04-01 14:58:16:INFO:metaphor:Ignoring dashboard from workbook in excluded project: Personal Space
2024-04-01 14:58:29:INFO:metaphor:Querying customSQLTablesConnection with offset 0
2024-04-01 14:58:29:INFO:metaphor:Found 1 custom SQL tables.
2024-04-01 14:58:29:INFO:metaphor:Ignoring datasources from workbook in excluded project: Personal Space
```
Both dashboard and datasources are ignored.

### ☑️ Checks


mars-lan (commented):
Otherwise LGTM

mars-lan (approved):

PR #814, opened 2024-04-01
[sc-25548] Ignore duplicate snowflake tags

usefulalgorithm:
### 🤔 Why?

Turns out Snowflake's account usage table can have duplicate tags in them, so we need to weed out the duplicate ones when updating object tags.

### 🤓 What?

- Check if tag already exists in object's tag field (`systemTags` for datasets, `tags` for columns`), only insert new tag if it is not there.

### 🧪 Tested?

- Added new test case.
- Tested locally.

### ☑️ Checks

### DIFF
--- a/metaphor/snowflake/extractor.py
+++ b/metaphor/snowflake/extractor.py
@@ -484,18 +484,20 @@ class SnowflakeExtractor(BaseExtractor):
             hierarchy_key = dataset_normalized_name(database, object_name)
         logical_id = HierarchyLogicalID(path=path)
 
-        self._hierarchies.setdefault(
+        system_tags = self._hierarchies.setdefault(
             hierarchy_key,
             Hierarchy(
                 logical_id=logical_id, system_tags=SystemTags(tags=[])
             ),  # SystemTags.tags should never be empty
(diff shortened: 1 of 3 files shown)


metaphor/snowflake/extractor.py
```diff
@@ -546,21 +550,21 @@ def _append_tag(
...
-            ORDER BY case when DOMAIN = 'DATABASE' then 1
-                          when DOMAIN = 'SCHEMA' then 2
-                          when DOMAIN = 'TABLE' then 3
-                          when DOMAIN = 'COLUMN' then 4
-                          end asc;
-            """
-        )
+        query = """
+        SELECT TAG_NAME, TAG_VALUE, DOMAIN, OBJECT_DATABASE, OBJECT_SCHEMA, OBJECT_NAME, COLUMN_NAME
+        FROM snowflake.account_usage.tag_references
```
mars-lan, line 559: Maybe the reason why there's repeated tags is because deleted objects? https://docs.snowflake.com/en/sql-reference/account-usage/tag_references#:~:text=parent%20table%20name.-,OBJECT_DELETED,-TIMESTAMP_LTZ

PR #818, opened 2024-04-02
[sc-25569] dbt cloud client infinite loop should update offset in payload

usefulalgorithm:
### 🤔 Why?

In the infinite loop I forgot to update payload's offset field, so it's always getting the same thing if no successful run is found. If no successful run was found for a job, then we simply skip it.

### 🤓 What?

Fix the abovementioned bug.

### 🧪 Tested?

- Added unit test.
- Tested locally.

### ☑️ Checks


elic-eon (approved):
LGTM!

PR #823, opened 2024-04-09
Auto-renew Power BI API access tokens

mars-lan:
### 🤔 Why?

Currently the connector only retrieves the access token once and reuses it throughout the entire run. This can lead to the following error when the token expires:

```
AssertionError: GET https://api.powerbi.com/v1.0/myorg/admin/groups/39fdb7eb-dd2c-4c93-9f6e-71b500723895/dashboards failed: 403
{"error":{"code":"TokenExpired","message":"Access token has expired, resubmit with a new access token"}}
```

### 🤓 What?

According to [the documentation](https://learn.microsoft.com/en-us/entra/msal/python/getting-started/acquiring-tokens#acquire-token-for-client), we should use [acquire_token_for_client](https://learn.microsoft.com/en-us/python/api/msal/msal.application.confidentialclientapplication?view=msal-py-latest#msal-application-confidentialclientapplication-acquire-token-for-client), which automatically handles token caching and renewal with MSAL 1.23+.

- Retrieve tokens (either from cache or from AAD) on every request.
- Update MSAL to the latest version.

### 🧪 Tested?

### ☑️ Checks


alyiwang (approved):
LGTM
thanks for fixing

PR #829, opened 2024-04-17
User get notifications when dbt crawler run

elic-eon:
### 🤔 Why?

Some dbt models are imported using [project-dependencies](https://docs.getdbt.com/docs/collaborate/govern/project-dependencies). That means the dbt crawler may generate two entities with the same logical ID but we should not generate the imported one.

### 🤓 What?

- Filter models the `package_name` is not equal to `project_name` if `project_name` is set.

### 🧪 Tested?

Add test case.

### ☑️ Checks

### DIFF
--- a/metaphor/dbt/artifact_parser.py
+++ b/metaphor/dbt/artifact_parser.py
@@ -1,6 +1,6 @@
 import json
 from copy import deepcopy
-from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Type, Union
 
 from metaphor.common.entity_id import EntityId
 from metaphor.common.logger import get_logger
@@ -335,6 +335,7 @@ class ArtifactParser:
         datasets: Dict[str, Dataset],
         virtual_views: Dict[str, VirtualView],
         metrics: Dict[str, Metric],
+        referenced_virtual_views: Set[str],
     ):
         self._platform = platform
         self._docs_base_url = config.docs_base_url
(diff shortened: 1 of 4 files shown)


metaphor/dbt/artifact_parser.py
```diff
@@ -427,6 +429,8 @@ def parse_manifest(
             .rsplit("/", 1)[-1]
             .split(".")[0]
         )
+        project_name = manifest_metadata.get("project_name", "")
```
mars-lan, line 432: If `project_name` is not set in metadata, all models will be filtered out?

elic-eon: no, if `project_name` is not set, all models will be extracted

elic-eon: [if project_name and node.package_name != project_name:](https://github.com/MetaphorData/connectors/pull/829/files#diff-a1df7e7750265ceca906dcd5962a36e71a550fa20dea40d2729a4ccde150bc4dR535)

metaphor/dbt/artifact_parser.py
```diff
@@ -528,6 +532,8 @@ def parse_manifest(
                 node.unique_id,
                 node_name_getter(node),
             )
+            if project_name and node.package_name != project_name:
```
usefulalgorithm, line 535: I think we want to include referenced views, not the other way around?

metaphor/dbt/extractor.py
```diff
@@ -67,11 +68,20 @@ async def extract(self) -> Collection[ENTITY_TYPES]:
...
-        entities.extend(self._virtual_views.values())
+        entities.extend(
+            map(
+                lambda item: item[1],
+                filter(
+                    lambda item: item[0] not in self._referenced_virtual_views,
+                    self._virtual_views.items(),
+                ),
+            )
+        )
```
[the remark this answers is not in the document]
elic-eon: 👍

usefulalgorithm (approved):
lgtm

PR #830, opened 2024-04-18
Snowflake crawler no longer fetches tables [sc-25711]

alyiwang:
### 🤔 Why?

Snowflake `SHOW SHARES` command now returns columns:
created_on    | kind     | owner_account           | name            | database_name  |  ....

Previously we were using `row[3]` to get the database name, now it should be `row[4]`.

### 🤓 What?

- Fix the `SHOW SHARES` results handling
- Also dedupe the shared databases from both `SHOW SHARES` and `information_schema` table

### 🧪 Tested?

tested on metaphor instance.

### ☑️ Checks


elic-eon (approved):

usefulalgorithm (approved):
lgtm, but the test needs fixing

PR #832, opened 2024-04-18
Not use SHOW SHARES [sc-25711]

alyiwang:
### 🤔 Why?

The `SHOW SHARES` will return all inbound/outbound shared databases, also including the inbound shared databases that have not been accepted by the receiver.

### 🤓 What?

- Not use `SHOW SHARES`, and use only `information_schema` instead. This does require the access to the imported databases to be granted to the metaphor role.

### 🧪 Tested?

tested on metaphor snowflake instance.

### ☑️ Checks

### DIFF
--- a/metaphor/snowflake/extractor.py
+++ b/metaphor/snowflake/extractor.py
@@ -186,13 +186,10 @@ class SnowflakeExtractor(BaseExtractor):
 
     @staticmethod
     def _fetch_shared_databases(cursor: SnowflakeCursor) -> List[str]:
-        cursor.execute("SHOW SHARES")
-        shared_database = [db[4].lower() for db in cursor if db[1] == "INBOUND"]
         cursor.execute(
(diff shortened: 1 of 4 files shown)


metaphor/snowflake/extractor.py
```diff
@@ -186,13 +186,10 @@ def fetch_databases(cursor: SnowflakeCursor) -> List[str]:
 
     @staticmethod
     def _fetch_shared_databases(cursor: SnowflakeCursor) -> List[str]:
-        cursor.execute("SHOW SHARES")
```
mars-lan, line 189: I think you also need to remove the `type != 'IMPORTED DATABASE'` filter on line 183 in order by the crawler to actually fetch the tables from shared DBs?

alyiwang: Good catch! Fixed

mars-lan (approved):

PR #839, opened 2024-04-22
Crawler filters aren t merged properly [sc-25950]

alyiwang:
### 🤔 Why?

The previous merge filter function will take two database filters and merge together with overwriting, i.e. if two filters has same key, then the second value will overwrite the first value. What we want instead is to merge the two values (schema filters) together. 

### 🤓 What?

- change the merge logic to actually merge the database filter values instead of overwriting. However, the schema filter level, we still do merge with overwriting.

### 🧪 Tested?

unit tests

### ☑️ Checks


mars-lan (commented):
Also need to merge the table patterns?

alyiwang:
> Also need to merge the table patterns?

updated

tests/common/test_filter.py
```diff
@@ -207,6 +207,64 @@ def test_merge():
...
+            "*": {
+                "foo": {"f1", "f2", "f3"},
+                "bar": {"b1"},
+            },
+        }
+    )
+
+    f1 = DatasetFilter(
+        excludes={
+            "*": {"foo": {"f1", "f2"}, "bar": None},
```
mars-lan, line 255: Consider replacing `*` with an actual schema name to show that it's not just limited to `*`?

mars-lan (commented):
Otherwise LGTM

tests/common/test_filter.py
alyiwang: updated

mars-lan (approved):

PR #840, opened 2024-04-22
Drop crawler-side SQL parsing from Metabase connector

mars-lan:
### 🤔 Why?

As we have moved native SQL parsing to the server side, there's no need to parse it on the connector side any more.

### 🤓 What?

As titled.

### 🧪 Tested?

### ☑️ Checks


alyiwang (approved):
LGTM

PR #843, opened 2024-04-23
looker crawler stuck in LookML parsing [sc-25902]

alyiwang:
### 🤔 Why?

sql-metadata parser may stuck at certain SQL queries.

### 🤓 What?

- don't do SQL parsing at crawler, but send the raw query in MCE for server-side processing.

### 🧪 Tested?

Unit tests
tested against Looker instance, and verified the MCE

### ☑️ Checks

### DIFF
--- a/metaphor/looker/lookml_parser.py
+++ b/metaphor/looker/lookml_parser.py
@@ -3,13 +3,11 @@ import glob
 import logging
 import operator
 import os
-import re
 from dataclasses import dataclass
 from typing import Dict, List, Optional, Set, Tuple
 
 try:
     import lkml
-    import sql_metadata
 except ImportError:
     print("Please install metaphor[looker] extra\n")
     raise
@@ -30,6 +28,7 @@ from metaphor.models.metadata_change_event import (
     LookerView,
     LookerViewDimension,
     LookerViewMeasure,
+    LookerViewQuery,
     SystemTag,
     SystemTags,
     SystemTagSource,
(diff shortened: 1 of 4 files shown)


metaphor/looker/lookml_parser.py
```diff
@@ -159,49 +161,29 @@ def _get_upstream_datasets(
             if base_view_name is None:
                 base_view_name = explore_name
 
-            upstreams.update(
-                _get_upstream_datasets(base_view_name, raw_model, connection)
+            query = LookerViewQuery(
```
usefulalgorithm, line 164: is it possible to have both `sql` and `explore_source` keys in `derived_table`? If so then `query` gets replaced, is this ok?

metaphor/looker/lookml_parser.py
```diff
@@ -159,49 +161,29 @@ def _get_upstream_datasets(
...
+                query="_",
```
usefulalgorithm, line 165: what does this mean? is this liquid syntax?

alyiwang: this is meant to be fixed by https://github.com/MetaphorData/app/pull/8269

metaphor/looker/lookml_parser.py
alyiwang: The derived table should have either one, but shouldn't be both.

metaphor/looker/lookml_parser.py
alyiwang: updated

usefulalgorithm (approved):
nice!

PR #845, opened 2024-04-23
Looker SQL server side parsing [sc-25997]

alyiwang:
### 🤔 Why?

Instead of putting the base view of a looker view in `query` field, we should treat is as the upstream of the looker view.

### 🤓 What?

- convert the base view to virtual view id and put it in the upstream of the looker view

### 🧪 Tested?

unit tests
tested against looker instance

### ☑️ Checks


usefulalgorithm (approved):
LGTM! go go go

PR #846, opened 2024-04-24
[sc-25976] Add `database_defaults` to Metabase crawler config

usefulalgorithm:
### 🤔 Why?

Metabase does not know a SQL query's default schema on Snowflake, RedShift and BigQuery. This causes our lineage parser unable to locate the correct upstream dataset, and in some instances the lineage parser to straight up fail.

Since Metabase API does not provide us with the default schema, best we can do is to specify this in the configs.

### 🤓 What?

- Add config for `database_defaults`, which contains `default_schema`
- Use a database's `default_schema` when parsing database info from metabase api

### 🧪 Tested?

- Unit test
- Tested locally

### ☑️ Checks

### DIFF
--- a/metaphor/metabase/README.md
+++ b/metaphor/metabase/README.md
@@ -32,7 +32,7 @@ Metabase's API does not provide information on the default schema used to execut
 ```yaml
 database_defaults:
     - id: <id of the database in Metabase>
-      default_schema: <default schema in the SQL platform>
+      default_schema: <default schema for the database>
 ```
 
 #### Output Destination
--- a/metaphor/metabase/README.md
+++ b/metaphor/metabase/README.md
@@ -27,7 +27,7 @@ password: <password>
 
 #### Database Defaults
 
-Metabase parses the SQL queries executed on Redshift, Snowflake and BigQuery, but it is unaware of a query's default schema during its execution. Metaphor's in-app SQL lineage parser will not be able to locate the correct dataset if the query's schema is not provided. To set the SQL platform's default schema, set the `database_defaults` configuration:
+Metabase's API does not provide information on the default schema used to execute [native queries](https://www.metabase.com/glossary/native_query). This makes it difficult to parse the lineage precisely. When this happens, use `database_defaults` to manually set the [database](https://www.metabase.com/docs/latest/databases/start)'s default schema:
 
 ```yaml
 database_defaults:
--- a/metaphor/metabase/README.md
+++ b/metaphor/metabase/README.md
@@ -25,6 +25,16 @@ password: <password>
 
 ### Optional Configurations
 
+#### Database Defaults
+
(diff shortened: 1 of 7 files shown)


metaphor/metabase/extractor.py
```diff
@@ -210,21 +211,36 @@ def _parse_database(self, database: Dict) -> None:
...
             # not able to get connection details, possibly due to lack of Admin permission
             return
 
         if platform == DataPlatform.SNOWFLAKE:
             self._databases[database_id] = DatabaseInfo(
-                platform, details.get("db"), None, details.get("account")
+                platform,
+                details.get("db"),
+                default_schema,
+                details.get("account"),
```
elic-eon, line 232: Didn't Metabase provide any schema information? Then, do we need to specify the default schema for every (Metabase) database?

elic-eon: what if the query is using the full table name, `SELECT * from db.schema.table`. Can we get the schema in the lineage parser?

elic-eon: Metabase may know which RS schema is queried against but we can tell from the API. There is another possibility, there is some default schema in the RS system. When we use this SQL, `SELECT * from TABLE`, there is a default schema.

mars-lan: The default schema is tied to the account used so it's not possible to tell that from Metabase side.

metaphor/metabase/README.md
```diff
@@ -25,6 +25,16 @@ password: <password>
 
 ### Optional Configurations
 
+#### Database Defaults
+
+Metabase parses the SQL queries executed on Redshift, Snowflake and BigQuery, but it is unaware of a query's default schema during its execution. Metaphor's in-app SQL lineage parser will not be able to locate the correct dataset if the query's schema is not provided. To set the SQL platform's default schema, set the `database_defaults` configuration:
```
mars-lan, line 30:
```suggestion
Metabase's API does not provide information on the default schema used to execute [native queries](https://www.metabase.com/glossary/native_query). This makes it difficult to parse the lineage precisely. When this happens, use `database_defaults` to manually set the [database](https://www.metabase.com/docs/latest/databases/start)'s default schema:
```

metaphor/metabase/README.md
```diff
@@ -25,6 +25,16 @@ password: <password>
...
+
+```yaml
+database_defaults:
+    - id: <id of the database in Metabase>
+      default_schema: <default schema in the SQL platform>
```
mars-lan, line 35:
```suggestion
      default_schema: <default schema for the database>
```

metaphor/metabase/README.md
usefulalgorithm: noice

metaphor/metabase/extractor.py
usefulalgorithm: > what if the query is using the full table name, SELECT * from db.schema.table. Can we get the schema in the lineage parser?

then it is ok

> Metabase may know which RS schema is queried against but we can tell from the API. ...

The tricky part is that MB doesn't know about RS schema, instead it only knows RS database.

PR #849, opened 2024-04-26
[sc-26028] Override inherited system tags

usefulalgorithm:
### 🤔 Why?

Instead of deduping system tags, what we need to do is to properly override inherited system tags: 

>It is possible to override an inherited tag on a given object. For example, if a table column inherits the tag named cost_center with a string value called sales, the tag can be updated with a more specific tag string value such as sales_na, to specify the North America sales cost center. Additionally, a new tag can be applied to the table column. Use an [ALTER TABLE … ALTER COLUMN](https://docs.snowflake.com/en/sql-reference/sql/alter-table-column) statement to update the tag string value on the column and to set one or more additional tags on a column.

https://docs.snowflake.com/en/user-guide/object-tagging#tag-lineage

### 🤓 What?

- Refactor
- Implement logic for overriding inherited system tags.

### 🧪 Tested?

- Added unit test.
- Tested on our Snowflake instance.

### ☑️ Checks

### DIFF
--- a/metaphor/snowflake/extractor.py
+++ b/metaphor/snowflake/extractor.py
@@ -65,8 +65,9 @@ from metaphor.snowflake.utils import (
     DatasetInfo,
     QueryWithParam,
     SnowflakeTableType,
+    append_column_system_tag,
+    append_dataset_system_tag,
     check_access_history,
-    dedup_dataset_system_tags,
     exclude_username_clause,
     fetch_query_history_count,
     str_to_source_type,
@@ -165,10 +166,6 @@ class SnowflakeExtractor(BaseExtractor):
 
         datasets = list(self._datasets.values())
         tag_datasets(datasets, self._tag_matchers)
-
-        # Dedup system tags
-        for dataset in datasets:
-            dedup_dataset_system_tags(dataset)
 
         entities: List[ENTITY_TYPES] = []
         entities.extend(datasets)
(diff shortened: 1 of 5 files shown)


metaphor/snowflake/utils.py
```diff
@@ -220,3 +223,11 @@ def fetch_query_history_count(
     if result is not None:
         return result[0]
     return 0
+
+
+def dedup_dataset_system_tags(dataset: Dataset) -> None:
+    assert dataset.system_tags and dataset.system_tags.tags is not None
+    dataset.system_tags.tags = [
+        SystemTag.from_dict(json.loads(x))
+        for x in set(json.dumps(tag.to_dict()) for tag in dataset.system_tags.tags)
```
mars-lan, line 232: I assume you don't need to convert the dict to json to dedupe?

usefulalgorithm: yes you do:
```
    def dedup_dataset_system_tags(dataset: Dataset) -> None:
        assert dataset.system_tags and dataset.system_tags.tags is not None
        dataset.system_tags.tags = [
            SystemTag.from_dict(x)
>           for x in set(tag.to_dict() for tag in dataset.system_tags.tags)
        ]
E       TypeError: unhashable type: 'dict'
```

mars-lan: Got it. Deduping using JSON-serialized string seems like a hack. Given that there are only two fields (`key` & `value`), why not do a field-by-field comparison?

mars-lan (approved):

metaphor/snowflake/utils.py
```diff
@@ -220,3 +223,50 @@ def fetch_query_history_count(
...
+    other_field_tags = [t for t in field.tags if t.split("=", 1)[0] != system_tag.key]
+    field.tags = other_field_tags + [_stringify_system_tag(system_tag)]
+
+
+def append_dataset_system_tag(dataset: Dataset, system_tag: SystemTag) -> None:
+    assert (
+        dataset.schema is not None
+        and dataset.system_tags
+        and dataset.system_tags.tags is not None
+    )
```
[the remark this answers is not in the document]
usefulalgorithm: This is to make pyright happy, otherwise there will be a lot of warnings in VSCode

metaphor/snowflake/extractor.py
```diff
@@ -567,30 +513,60 @@ def _fetch_tags(self, cursor: SnowflakeCursor) -> None:
...
-            else:
-                self._append_tag(
-                    key_prefix, key, value, column if object_type == "COLUMN" else None
+                    if dataset_key.startswith(prefix):
+                        append_dataset_system_tag(
+                            self._datasets[dataset_key], system_tag
+                        )
+
+            elif object_type == "TABLE":
+                dataset_key = dataset_normalized_name(
```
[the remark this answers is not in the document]
usefulalgorithm: it is the key to look for in `self._datasets`, I think calling it `dataset_key` is less likely to cause confusion for future me XD

elic-eon (approved):

PR #852, opened 2024-05-02
Support Databricks cluster based extractors

elic-eon:
### 🤔 Why?

Currently, we run SQL on a SQL warehouse for metadata. This PR is to support running SQL on a compute cluster.

### 🤓 What?

- Add an optional configuration to set the cluster path for a cluster

### 🧪 Tested?

Verified against Metaphor Databricks env.

### ☑️ Checks

### DIFF
--- a/metaphor/unity_catalog/README.md
+++ b/metaphor/unity_catalog/README.md
@@ -31,16 +31,6 @@ See [Output Config](../common/docs/output.md) for more information.
 
 See [Filter Configurations](../common/docs/filter.md) for more information on the optional `filter` config.
 
-#### Warehouse ID
-
-To run the queries using a specific warehouse, simply add its ID in the configuration file:
-
-```yaml
-warehouse_id: <warehouse_id>
-```
-
-If no warehouse id nor cluster path is provided, the connector automatically uses the first discovered warehouse.
-
 #### Cluster Path
 
 To run the queries using a compute cluster, add its cluster path in the configuration file:
@@ -78,6 +68,18 @@ query_log:
   # (Optional) Limit the number of results returned in one page of query log history. The default is 100.
   max_results: <count>
 ```
+
+#### Warehouse ID
+
+Note: we encourage using cluster, this connector will deprecate the SQL warehouse support.
+
+To run the queries using a specific warehouse, simply add its ID in the configuration file:
+
+```yaml
+warehouse_id: <warehouse_id>
+```
+
+If no warehouse id nor cluster path is provided, the connector automatically uses the first discovered warehouse.
 
 ## Testing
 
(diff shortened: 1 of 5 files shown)


metaphor/unity_catalog/extractor.py
```diff
@@ -107,7 +107,11 @@ def __init__(self, config: UnityCatalogRunConfig):
         self._source_url = config.source_url
         self._api = create_api(self._host, self._token)
         self._connection = create_connection(
-            self._api, self._token, config.warehouse_id
+            self._api,
+            self._token,
+            config.warehouse_id,
+            cluster_url=config.host,
```
mars-lan, line 113: Since `config.host` is required, there doesn't seem a need for `warehouse_id`?

elic-eon: We will support SQL warehouse and compute cluster for the current implementation.

metaphor/unity_catalog/utils.py
```diff
@@ -101,20 +101,36 @@ def escape_special_characters(name: str) -> str:
 
 
 def create_connection(
-    client: WorkspaceClient, token: str, warehouse_id: Optional[str] = None
+    client: WorkspaceClient,
+    token: str,
+    warehouse_id: Optional[str] = None,
+    cluster_url: Optional[str] = None,
```
mars-lan, line 107: Should rename this to `cluster_hostname` to avoid confusion?

metaphor/unity_catalog/utils.py
```diff
@@ -101,20 +101,36 @@ def escape_special_characters(name: str) -> str:
...
+    cluster_path: Optional[str] = None,
 ) -> Connection:
     endpoints = list(client.warehouses.list())
-    if not endpoints:
-        raise ValueError("No valid warehouse found")
-    endpoint_info: EndpointInfo = endpoints[0]
+    if not endpoints and cluster_url is None and cluster_path is None:
```
mars-lan, line 111: This logic is rather difficult to understand. Maybe check `cluster_hostname` & `cluster_path` first and connect via cluster and only fallback to SQL warehouses otherwise?

metaphor/unity_catalog/README.md
```diff
@@ -39,7 +39,17 @@ To run the queries using a specific warehouse, simply add its ID in the configur
 warehouse_id: <warehouse_id>
 ```
 
-If no warehouse id is provided, the connector automatically uses the first discovered warehouse.
+If no warehouse id nor cluster path is provided, the connector automatically uses the first discovered warehouse.
```
mars-lan, line 42: Consider adding a deprecation note asking people to set `cluster_path` instead. Also move this section to the end?

mars-lan (approved):

PR #857, opened 2024-05-06
Refactor config for Unity Catalog connector

mars-lan:
### 🤔 Why?

Refactor the config to support both Databricks cluster & SQL warehouse and to better match Unity Catalog's terminology: https://docs.databricks.com/en/integrations/compute-details.html

### 🤓 What?

- Replace `host`, `cluster_path`, and `warehouse_id` with `hostname` & `http_path`.
- Update README files

### 🧪 Tested?

Verified against a production instance using the new config fields.

### ☑️ Checks

- [x] Not bumping up to 0.14 until all breaking changes are merged.

### DIFF
--- a/metaphor/unity_catalog/README.md
+++ b/metaphor/unity_catalog/README.md
@@ -17,9 +17,12 @@ Create a YAML config file based on the following template.
 ### Required Configurations
 
 ```yaml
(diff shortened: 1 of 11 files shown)


alyiwang (approved):
LGTM

PR #859, opened 2024-05-07
[sc-26213] Lower max query text length limit to 100K

usefulalgorithm:
### 🤔 Why?

The Snowflake crawler parses query logs from the `QUERY_HISTORY` view, whose `query_text` column is potentially truncated if the value is too long: https://docs.snowflake.com/en/sql-reference/account-usage/query_history

(The link says the limit is 100K characters, but in our Snowflake instance we can see a query is parsed that's longer than 100K:
截圖 2024-05-07 晚上7 39 04
)

From the lineage parser's perspective a truncated expression is an invalid expression. To make lineage parsing possible, just drop the sql queries that are longer than 100K characters long, since it's very likely to be a `INSERT INTO` expression.

### 🤓 What?

Lower max query text length limit to 100_000.

### 🧪 Tested?

- Crawler run locally against metaphor's Snowflake instance.

### ☑️ Checks


mars-lan:
I'm confused, the example you gave shows that a sql_text of length 144k charts got truncated. Isn't that the correct expectation?

usefulalgorithm:
> I'm confused, the example you gave shows that a `query_text` of length 144k chars got truncated. Isn't that the correct expectation?

ack I saw wrong, wait wait

usefulalgorithm:
@mars-lan  can you take another look? updated the pr description and some names in the logic

mars-lan:
> @mars-lan can you take another look? updated the pr description and some names in the logic

Got it. In that case, let's just ignore queries longer than 100k w/o doing any regex magic?

metaphor/snowflake/config.py
```diff
@@ -12,8 +12,8 @@
 # number of query logs to fetch from Snowflake in one batch
 DEFAULT_QUERY_LOG_FETCH_SIZE = 100000
 
-# By default ignore queries larger than 512KiB
-DEFAULT_MAX_QUERY_SIZE = 512 * 1024
+# By default ignore queries longer than 100K characters
```
mars-lan, line 15: Add reference URL?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.14.0"
+version = "0.14.1"
```
mars-lan, line 3: Update brach and bump again?

mars-lan (commented):
Otherwise LGTM

mars-lan (approved):

PR #862, opened 2024-05-20
[sc-26422] Use `DIRECT_OBJECTS_ACCESSED` instead of `BASE_OBJECTS_ACCESSED` in `ACCESS_HISTORY`

usefulalgorithm:
### 🤔 Why?

This is so that we don't collapse the views along the lineage. 

Also includes `STREAM` in the object domains filter.

### 🤓 What?

- Use `DIRECT_OBJECTS_ACCESSED` instead of `BASE_OBJECTS_ACCESSED` in `ACCESS_HISTORY`
- Includes `STREAM` in the object domains filter.

### 🧪 Tested?

- SnowflakeLineage crawler run locally
 
### ☑️ Checks

### DIFF
--- a/metaphor/snowflake/lineage/extractor.py
+++ b/metaphor/snowflake/lineage/extractor.py
@@ -33,6 +33,7 @@ SUPPORTED_OBJECT_DOMAIN_TYPES = (
     "TABLE",
     "VIEW",
     "MATERIALIZED VIEW",
+    "STREAM",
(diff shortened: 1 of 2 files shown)


usefulalgorithm:
Currently trying to setup my environment so that I can verify crawled lineage is right

mars-lan (approved):

PR #869, opened 2024-05-23
Fix missing logs in output

mars-lan:
### 🤔 Why?

Connectors are not outputting the complete log to files.

### 🤓 What?

Refactor `run_connector` so that the logs are written at the very end. Also simplify the logic by creating a `FileConfig` that writes to a local temp directory if not specified.

### 🧪 Tested?

Verified that the output stays identical (other than the additional log entries) before & after the change.

### ☑️ Checks

### DIFF
--- a/metaphor/common/runner.py
+++ b/metaphor/common/runner.py
@@ -1,9 +1,10 @@
+import tempfile
 import traceback
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


alyiwang (approved):
LGTM

PR #870, opened 2024-05-27
Add debug info for snowflake crawler

elic-eon:
### 🤔 Why?

As title

### 🤓 What?

- log the summary of comment extracting

### 🧪 Tested?

Unit test, verified against our snowflake

### ☑️ Checks


metaphor/snowflake/extractor.py
```diff
@@ -957,13 +956,29 @@ def _fetch_database_comment(self, cursor: SnowflakeCursor) -> None:
...
 
+        database_count, comment_count = 0, 0
         for database_name, comment in cursor:
+            database_count += 1
+            if comment:
+                comment_count += 1
             self._add_hierarchy_comment(database_name, comment)
 
-    def _fetch_schemas_comment(self, cursor: SnowflakeCursor) -> None:
+        logger.debug(f"Fetch {comment_count}/{database_count} database comment")
```
mars-lan, line 966: Instead of printing the count, might as well print out each database name in the loop?

metaphor/snowflake/extractor.py
```diff
@@ -957,13 +956,29 @@ def _fetch_database_comment(self, cursor: SnowflakeCursor) -> None:
...
+        schema_count, comment_count = 0, 0
+        for catalog_name, schema, comment in cursor:
+            schema_count += 1
+            if comment:
+                comment_count += 1
+            self._add_hierarchy_comment(
+                database=catalog_name, schema=schema, comment=comment
+            )
+
+        logger.debug(
```
mars-lan, line 982: Ditto

metaphor/snowflake/extractor.py
elic-eon: Updated

metaphor/snowflake/extractor.py
elic-eon: Updated

mars-lan (approved):

PR #872, opened 2024-05-30
Allow assigning additional dbt tags via meta field

mars-lan:
### 🤔 Why?

We're moving away from treating tags declared in dbt's [meta](https://docs.getdbt.com/reference/resource-configs/meta) field as Metaphor's governance tag. Instead, they should be treated just as additional "dbt tags".

### 🤓 What?

- Add `meta_key_tags` to dbt connector config to point to the key name in `meta` field where the additional dbt tags are defined.
- Combining and deduping additional dbt tags defined in `meta` with the [tags](https://docs.getdbt.com/reference/resource-configs/tags) defined at the model level into `SystemTags` in MCE.
- Rename `get_tags_from_meta` to `get_metaphor_tags_from_meta` to disambiguate.

TODOs in follow-up PRs:
- Migrate Metaphor ownerships declared in `meta` into `SystemContacts`.
- Update models to support column-level tags.

> Note: `tests/dbt/data/ride_share/manifest.json` & `tests/dbt/data/ride_share/run_results.json` are regenerated based on the latest code in https://github.com/MetaphorData/dbt, which led to a substantial diff.

### 🧪 Tested?

Verified end to end against a production instance.

### ☑️ Checks

### DIFF
--- a/metaphor/dbt/artifact_parser.py
+++ b/metaphor/dbt/artifact_parser.py
@@ -12,11 +12,13 @@ from metaphor.dbt.util import (
     build_metric_docs_url,
     build_model_docs_url,
     build_source_code_url,
+    build_system_tags,
     dataset_normalized_name,
+    get_dbt_tags_from_meta,
+    get_metaphor_tags_from_meta,
     get_model_name_from_unique_id,
     get_ownerships_from_meta,
     get_snapshot_name_from_unique_id,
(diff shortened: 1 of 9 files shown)


elic-eon (dismissed):
LGTM!

elic-eon (approved):

PR #877, opened 2024-06-11
Move text_from_HTML out of metaphor.common

mars-lan:
### 🤔 Why?

The `text_from_HTML` util function depends on the rather heavy `beautifulsoup` library. Since we don't expect it to be used by connectors other than `static_web` & `sharepoint`, it makes sense to move it out of `metaphor.common.utils`. 

### 🤓 What?

- Move `text_from_HTML` to from `metaphor.common.utils` to`metaphor.static_web.util`, along with `title_from_HTML`.
- Refactor `static_web` & `sharepoint` connectors to make use of these util functions.

### 🧪 Tested?

Unit tests.

### ☑️ Checks

### DIFF
--- a/metaphor/common/utils.py
+++ b/metaphor/common/utils.py
@@ -3,8 +3,6 @@ from datetime import datetime, time, timedelta, timezone
 from hashlib import md5
 from typing import Any, Callable, Dict, Iterable, List, Optional, Union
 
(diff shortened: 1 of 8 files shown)


usefulalgorithm (approved):
nice

PR #879, opened 2024-06-12
[sc-26791][sc-26630] Parse test run results via discovery api, and fill in `DataMonitor.last_run` for dbt data monitors

usefulalgorithm:
### 🤔 Why?

It is possible for `run_results.json` to not have the results of previously run tests. Since dbt cloud provides the discovery API for metadata discovery, we should use that instead of only relying on `run_results.json`. We should not overwrite the data monitors collected by DbtCoreExtractor though; the monitors we collected via discovery API are only added to the collected entities if it does not exist before. 

### 🤓 What?

- Added logic in `dbt.cloud.DbtCloudExtractor` to extend core's parsed entities with test results parsed from discovery api.

### 🧪 Tested?

- Crawler run locally, seems normal
- Successfully ingested data monitors into andy.dev.metaphor.io.
- Added unit test.

### ☑️ Checks

### DIFF
--- a/metaphor/dbt/cloud/discovery_api.py
+++ b/metaphor/dbt/cloud/discovery_api.py
@@ -1,0 +1,69 @@
+from typing import Any, Dict
+
+from requests import post
+
+from metaphor.common.entity_id import dataset_normalized_name
+
+
+class DiscoveryAPI:
+    def __init__(self, url: str, token: str) -> None:
+        self.url = url
+        self.token = token
+
+    def _send(self, query: str, variables: Dict[str, Any]):
+        resp = post(
+            url=self.url,
+            headers={
+                "authorization": f"Bearer {self.token}",
+                "content-type": "application/json",
+            },
+            json={"query": query, "variables": variables},
+            timeout=15,
+        )
+        return resp.json()["data"]
+
+    def get_model_dataset_name(self, job_id: int, model_unique_id: str):
+        query = """
+query Model($uniqueId: String!, $jobId: BigInt!) {
+    job(id: $jobId) {
+        model(uniqueId: $uniqueId) {
+            alias
+            database
+            name
+            schema
+        }
+    }
+}
+        """
+        variables = {
+            "uniqueId": model_unique_id,
+            "jobId": job_id,
+        }
+
+        model = self._send(query, variables)["job"]["model"]
+        database, schema, name = (
+            model.get("database"),
+            model.get("schema"),
+            model.get("alias") or model.get("name"),
+        )
+        return dataset_normalized_name(database, schema, name)
+
+    def get_test_status(self, job_id: int, test_unique_id: str):
+        query = """
+query Test($uniqueId: String!, $jobId: BigInt!) {
+  job(id: $jobId) {
+    test(uniqueId: $uniqueId) {
+      status
+    }
+  }
+}
+        """
+        variables = {
+            "jobId": job_id,
+            "uniqueId": test_unique_id,
+        }
+
+        res = self._send(query, variables)["job"]["test"]
+        if res is None:
+            return "skipped"
+        return res["status"]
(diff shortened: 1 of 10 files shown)


metaphor/dbt/cloud/discovery_api.py
```diff
@@ -0,0 +1,76 @@
...
+        return dataset_normalized_name(database, schema, name)
+
+    def get_test_status(self, job_id: int, test_unique_id: str):
+        """
+        If the test cannot be found, we consider it as `skipped`.
+        """
+        query = """
+query Test($uniqueId: String!, $jobId: BigInt!) {
+  job(id: $jobId) {
+    test(uniqueId: $uniqueId) {
```
mars-lan, line 62: Instead of querying a single test at a time, it seems more efficient to just dump all the tests for a specific job using the `tests` field?

metaphor/dbt/cloud/discovery_api.py
```diff
@@ -0,0 +1,80 @@
...
+            model.get("alias") or model.get("name"),
+        )
+        return dataset_normalized_name(database, schema, name)
+
+    def get_all_test_status(self, job_id: int):
+        query = """
+query Tests($jobId: BigInt!) {
+  job(id: $jobId) {
+    tests {
+        uniqueId
```
mars-lan, line 68: Let's also capture `executeCompletedAt`?

metaphor/dbt/cloud/extractor.py
```diff
@@ -102,7 +115,74 @@ async def _extract_last_run(self, job_id: int):
...
+    ):
+        logger.info("Parsing test run results")
+
+        discovery_api = DiscoveryAPI(self._discovery_api_url, self._service_token)
+
+        new_monitor_datasets: List[Dataset] = list()
+
+        # Collect all datasets we've found so far, they're the data monitors.
+        # We do this because we don't really want to override previously collected
+        # data monitors.
```
mars-lan, line 141: Rather than doing this, why not rely on discovery API for all the data monitors? For dbt core, we can explicitly specify `run_results.json` and use that as the source for data monitors.

metaphor/dbt/cloud/extractor.py
```diff
@@ -94,15 +99,68 @@ async def _extract_last_run(self, job_id: int):
...
+                platform=platform,
+                account=account,
+            )
+
+            dataset = Dataset(
+                logical_id=dataset_logical_id,
+            )
+            new_monitor_datasets.append(dataset)
+
+            for test in discovery_api.get_all_job_tests(job_id):
```
mars-lan, line 150: This still makes GraphQL calls N times instead of once?

metaphor/dbt/cloud/extractor.py
```diff
@@ -94,15 +99,68 @@ async def _extract_last_run(self, job_id: int):
...
+            dataset_logical_id = DatasetLogicalID(
+                name=discovery_api.get_model_dataset_name(job_id, model_unique_id),
+                platform=platform,
+                account=account,
+            )
+
+            dataset = Dataset(
+                logical_id=dataset_logical_id,
+            )
+            new_monitor_datasets.append(dataset)
```
mars-lan, line 148: We should only create a new dataset if there's a monitor? Not every dbt model will have tests.

metaphor/dbt/cloud/extractor.py
usefulalgorithm: no, just once

usefulalgorithm:
```python
    def get_all_job_tests(self, job_id: int):
        query = """
query Tests($jobId: BigInt!) {
  job(id: $jobId) {
    tests {
        uniqueId
        name
        status
        columnName
        executeCompletedAt
    }
  }
}
        """
        variables = {
            "jobId": job_id,
        }
        tests = self._send(query, variables)["job"]["tests"]
        return [DiscoveryTestNode.model_validate(test) for test in tests]
```

mars-lan: Since this is inside a for loop, wouldn't it get called once per entity? If so, that's very inefficient.

usefulalgorithm: good catch, thanks

mars-lan (approved):

PR #881, opened 2024-06-19
[sc-26934] Expose `query_log_batch_size_count` as a config value

usefulalgorithm:
### 🤔 Why?

Ovative's lineage parser is timing out. Currently the lineage parser is able to handle individual queries, but the lineage lambda actually writes all parsed lineage into MongoDB one *all* queries have been parsed, which causes a single batch that contains multiple very long queries not being written to MongoDB at all.

One way to solve this is to reduce the number of query logs in a single batch file. Before it's hard-coded to 100,  but we probably want to expose this config value and try out different possible values to see if that'll fix the issue.

### 🤓 What?

Exposed `query_log_batch_size_count` in `FileSinkConfig`. Set it like
```yaml
output:
    file:
        directory: ...
        query_log_batch_size_count: ... /// Default is 100
```

### 🧪 Tested?

Tested locally, verified that with different `query_log_batch_size_count` the number of queries in a single batch file does change.

### ☑️ Checks

### DIFF
--- a/metaphor/common/docs/output.md
+++ b/metaphor/common/docs/output.md
@@ -22,6 +22,9 @@ output:
 
     # (Optional) IAM role to assume. Default using the current AWS credential.
     assume_role_arn: <iam_role_arn>
+
+    # (Optional) Maximum number of query logs to store in one batch file. Default to 100.
+    query_log_batch_size_count: <query_logs_per_file>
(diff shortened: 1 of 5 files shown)


elic-eon (approved):
LGTM!

PR #882, opened 2024-06-19
Snowflake crawler to populate sourceInfo

elic-eon:
### 🤔 Why?

Since we deprecate `statistics.lastUpdated`, we need to populate `sourceInfo.lastUpdated`

### 🤓 What?

- Update util `to_dataset_statistics`
- Refactor time util function

### 🧪 Tested?

Unit tests passed

### ☑️ Checks

### DIFF
--- a/metaphor/common/utils.py
+++ b/metaphor/common/utils.py
@@ -176,20 +176,14 @@ def is_email(email: str) -> bool:
     return True
 
 
-def to_utc_datetime(
+def to_utc_datetime_from_timestamp(
     timestamp: Optional[Union[float, int]] = None,
-    timestamp_ms: Optional[Union[float, int]] = None,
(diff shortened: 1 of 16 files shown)


metaphor/common/utils.py
```diff
@@ -172,3 +174,22 @@ def is_email(email: str) -> bool:
     except (ValueError, AssertionError):
         return False
     return True
+
+
+def to_utc_datetime(
```
mars-lan, line 179: The signature of this function is rather confusing. I'd name it `to_utc_time_from_timestamp` and take only ms version of the timestamp.

elic-eon: Updated!

alyiwang (approved):
LGTM

PR #885, opened 2024-06-27
[sc-27082] Split up `stl_querytext` table with OVER before aggregating query text rows

usefulalgorithm:
### 🤔 Why?

The Redshift crawler could fail while collecting query logs if there are a lot of queries in the cluster:
```
Invalid operation: Result size exceeds LISTAGG limit
```
To avoid that, we split `stl_querytext` with `OVER` clause before doing `LISTAGG`, so that we're doing `LISTAGG` on single queries instead of the whole column.

Reference: https://thedataguy.in/redshift-reconstructing-sql-from-sql-querytext/

### 🤓 What?

- Split up `stl_querytext` table with OVER before aggregating query text rows

### 🧪 Tested?

Run the updated query against our own Redshift cluster. Our cluster does not have much going on though, so it's hard to say how this change will impact the performance in production.

### ☑️ Checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.14.20"
+version = "0.14.21"
 license = "Apache-2.0"
 description = "A collection of Python-based 'connectors' that extract metadata from various sources to ingest into the Metaphor app."
 authors = ["Metaphor <dev@metaphor.io>"]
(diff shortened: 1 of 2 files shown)


alyiwang (approved):
Great fix!

alyiwang:
Need to bump the version again

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "metaphor-connectors"
-version = "0.14.20"
+version = "0.14.21"
```
alyiwang, line 3: Need to bump the version again

alyiwang (changes requested):

alyiwang (approved):

PR #889, opened 2024-07-02
[sc-27082] Filter out the query texts that go beyond the size limit

usefulalgorithm:
### 🤔 Why?

The `LISTAGG` expression was failing because there are SQL queries that are longer than 65535 characters. These really long queries are generated by things like DBT or Hex, and very likely not have any lineage in them.

Instead of trying to concatenate these queries in crawler code, it is perhaps better to just ignore them.

This PR stems from a Slack conversation with BioRender. _**This PR should not be merged until someone else gives the green light!**_

### 🤓 What?

Filter out the queries that go beyond 65535 characters.

### 🧪 Tested?

Tested on our own Redshift cluster:
- Import a large dataset
- Do a bunch of selects
- Set the size limit to 100
- Run the modified query in console
- Observe there are only a few queries that are below the size limit

### ☑️ Checks

### DIFF
--- a/metaphor/redshift/access_event.py
+++ b/metaphor/redshift/access_event.py
@@ -73,6 +73,11 @@ GROUP BY
     q.sequence
 ORDER BY ss.endtime DESC;
 """
+"""
+The condition `q.length < 65536` is because Redshift's LISTAGG method
+is unable to process the query if it is over 65535 characters long.
+See https://docs.aws.amazon.com/redshift/latest/dg/r_WF_LISTAGG.html#r_WF_LISTAGG-data-types
+"""
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


alyiwang (dismissed):
LGTM
Thanks for the fix

mars-lan (approved):

metaphor/redshift/access_event.py
```diff
@@ -38,11 +45,12 @@
 FROM stl_scan ss
     JOIN svv_table_info sti ON ss.tbl = sti.table_id
     JOIN stl_query sq ON ss.query = sq.query
-    JOIN stl_querytext sqt ON ss.query = sqt.query
+    JOIN queries q ON ss.query = q.query
     JOIN svl_user_info sui ON sq.userid = sui.usesysid
 WHERE ss.starttime >= '{start_time}'
     AND ss.starttime < '{end_time}'
     AND sq.aborted = 0
+    AND q.length < 65536
```
theseyi, line 53: For posterity, consider adding a comment on why this limit exists

mars-lan (dismissed):

mars-lan (approved):

PR #891, opened 2024-07-03
[sc-27082] Do the query text concatenation as CTE

usefulalgorithm:
### 🤔 Why?

I don't know why it is still failing, I guess it has to do with the order of execution within the entire SELECT statement?

Anyway I think the safest approach is to split the query collection into multiple stages:
1. Get the length of the queries
2. Filter out the queries that are too long
3. Do the LISTAGG
We then combine the queries back into the main SELECT.

### 🤓 What?

Fix the access event query.

### 🧪 Tested?

- Tested against our own redshift cluster.
- Tested on production: 
截圖 2024-07-04 下午5 01 22

### ☑️ Checks

### DIFF
--- a/metaphor/redshift/access_event.py
+++ b/metaphor/redshift/access_event.py
@@ -5,41 +5,28 @@ from typing import AsyncIterator
 from asyncpg import Connection, Record
 
 REDSHIFT_USAGE_SQL_TEMPLATE = """
-WITH queries as (
+WITH queries AS (
     SELECT
         *,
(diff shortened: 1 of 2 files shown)


alyiwang (dismissed):
If possible, please test against production data to make sure this works.

usefulalgorithm:
> If possible, please test against production data to make sure this works.

That is the plan

usefulalgorithm:
Finally got it to work...

mars-lan (approved):

PR #893, opened 2024-07-08
[sc-27305] Regenerate dbt manifest models

usefulalgorithm:
### 🤔 Why?

There's a new field `cumulative_type_params`, without it in our generated manifest models our crawlers will break.

### 🤓 What?

Regenerate the dbt manifest models by running `generate_all_models.sh`.

### 🧪 Tested?

Tested with QA configs locally.

### ☑️ Checks


mars-lan (approved):
Thanks for fixing

PR #896, opened 2024-07-09
[sc-27368] Allow other embedding models

rishimo:
### 🤔 Why?

[SC-27368](https://app.shortcut.com/metaphor-data/story/27368/allow-openai-embedding-models-in-document-connectors)

So document embedding connectors support the regular OpenAI models in addition to Azure OpenAI services. Other embedding models can be introduced as desired with the simplified configuration.

### 🤓 What?

* Introduces the `EmbeddingModelConfig` to minimize configuration fields that have to be manually specified and make it easier to add new embedding models in the future
* Updates document connectors to make use of new configuration
* Added support for OpenAI API embedding models
* Documentation about embedding models improved

### 🧪 Tested?

Connectors were run locally using OpenAI and Azure OpenAI successfully.

### ☑️ Checks


rishimo:
Could also refactor embedding model configs more generically, e.g.:

```
embed_source: "azure"
embed_model_configs:
  azure...:
  azure...:
```

to make more compatible with other embedding models. Would be a breaking change, but these connectors are not used in production yet.

docs/embeddings.md
```diff
@@ -0,0 +1,53 @@
...
+
+The following models are known to work:
+* `text-embedding-ada-002`
+* `text-embedding-3-small`
+
+### Configuration
+
+```yaml
+# REQUIRED CONFIGS
+embed_model_config:
```
mars-lan, line 20: 1. the suffix `_config` seems redundant since everything is a config
2. Since this is already a breaking change, let's mak it more generic? Something like
```
embedding_model:
  azure_open_ai:
     key:
     endpoint:
     ...
  
  open_ai:
    key:
    model:
    ...
```

rishimo: This makes sense - I will stick to `azure-openai` and `openai` since those are the `aiProvider` conventions we already have in the app.

mars-lan: Please drop the `_config` suffix.

rishimo: Missed those docs; should be good now.

docs/embeddings.md
```diff
@@ -0,0 +1,54 @@
...
+### Configuration
+
+```yaml
+# REQUIRED CONFIGS
+embedding_model:
+  azure_openai:
+    key: <key>
+    endpoint: <endpoint>
+    # DEFAULTS; DON'T NEED TO BE CONFIGURED
+    version: <version> # "2024-03-01-preview"
```
mars-lan, line 25:
```suggestion
    version: <version> # Default to "2024-03-01-preview" if not specified
```

docs/embeddings.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Configuration
+
+```yaml
+# REQUIRED CONFIGS
+embedding_model:
+  azure_openai:
+    key: <key>
+    endpoint: <endpoint>
+    # DEFAULTS; DON'T NEED TO BE CONFIGURED
```
mars-lan, line 24: MIght be less confusing to add defaults in the comments below?

metaphor/common/embeddings_config.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@dataclass
+class EmbeddingModelConfig:
+    azure_openai: AzureOpenAIConfig = field(default_factory=AzureOpenAIConfig)
+    openai: OpenAIConfig = field(default_factory=OpenAIConfig)
+    chunk_size: int = 512
+    chunk_overlap: int = 50
+
+    def update(self, config_dict: Dict[str, Dict]):
```
mars-lan, line 29: I assume this can be significantly simplified using validators? https://docs.pydantic.dev/latest/concepts/validators/

metaphor/common/embeddings.py
```diff
@@ -49,16 +45,35 @@ def embed_documents(
...
-    embed_model = AzureOpenAIEmbedding(
-        model=azure_openAI_model,
-        deployment_name=azure_openAI_model_name,
-        api_key=azure_openAI_key,
-        azure_endpoint=azure_openAI_endpoint,
-        api_version=azure_openAI_ver,
-    )
+    # Determine the source from the embedding_model configuration
+    source = None
+    for src in supported_sources:
```
mars-lan, line 50: Instead of this rather complicated business logic, why not just make `azure_openai` & `openai` fields optional and take the first not None config?

metaphor/common/embeddings_config.py
```diff
@@ -1,50 +1,56 @@
...
-    openai: OpenAIConfig = field(default_factory=OpenAIConfig)
+class EmbeddingModelConfig(BaseModel):
+    azure_openai: Optional[AzureOpenAIConfig] = None
+    openai: Optional[OpenAIConfig] = None
     chunk_size: int = 512
     chunk_overlap: int = 50
 
-    def update(self, config_dict: Dict[str, Dict]):
+    @model_validator(mode="before")
+    def apply_defaults(cls, values: Dict[str, Any]) -> Dict[str, Any]:
```
mars-lan, line 28: Still not sure why is this needed in the first place? Shouldn't Pydantic handles the default value automatically? The only thing that requires validation is that `azure_openai` or `openai` can't be both None or both set?

rishimo: 🤦 yes, I was overthinking it. Updated validation

metaphor/common/embeddings_config.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class EmbeddingModelConfig(BaseModel):
+    azure_openai: Optional[AzureOpenAIConfig] = None
+    openai: Optional[OpenAIConfig] = None
+    chunk_size: int = 512
+    chunk_overlap: int = 50
+
+    @model_validator(mode="before")
+    def check_only_one_config(cls, values: Dict[str, Any]) -> Dict[str, Any]:
```
mars-lan, line 28: Pls leverage the existing util function: https://github.com/MetaphorData/connectors/blob/559a4ec546edd372a64d5d26c3b68d3c0eb88837/metaphor/common/git.py#L32

mars-lan (approved):
Otherwise LGTM. Much cleaner now

mars-lan (dismissed):

rishimo:
needed to bump version 😢

PR #897, opened 2024-07-10
Should not fill out last update time in the UC profile crawler

elic-eon:
### 🤔 Why?

Because the the updated time is moved to `SourceInfo` so we need to remove these lines of code. We can't fill the `sourceInfo.lastUpdated` because the two crawler will overwrite this aspect. The UC connector already has the update time of tables so I think it's safe to remove it from the profile connector

### 🤓 What?

- Stop filling `statistics.lastUpdated`

### 🧪 Tested?

Unit tests

### ☑️ Checks

### DIFF
--- a/metaphor/unity_catalog/profile/extractor.py
+++ b/metaphor/unity_catalog/profile/extractor.py
@@ -179,12 +179,6 @@ class UnityCatalogProfileExtractor(BaseExtractor):
                 )
                 field_statistics.field_statistics.append(stats)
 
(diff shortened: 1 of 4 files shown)


mars-lan (commented):
Is this the only crawler left?

elic-eon:
Yes. We use a util function to create `DatasetStatistics` except this one.

mars-lan (approved):

PR #898, opened 2024-07-11
Support BigQuery table snapshots

elic-eon:
### 🤔 Why?

Currently throwing error when processing a snapshot table we need to support this type of table

### 🤓 What?

- Parse the snapshot table and fill in the snapshot time.
- Our lineage stage should get the table/column lineage from the DDL.

### 🧪 Tested?

Create a snapshot and run the crawler with new change.

### ☑️ Checks


usefulalgorithm (changes requested):
Otherwise LGTM

usefulalgorithm (approved):

PR #902, opened 2024-07-15
All crawlers emit normalized fieldPath [sc-27452]

alyiwang:
### 🤔 Why?

to support more efficient regex matching on column path

### 🤓 What?

- all crawlers emit column `fieldpath` as lowercase, and use `fieldName` to store the original string

### 🧪 Tested?

tested various crawlers against metaphor instances

### ☑️ Checks

### DIFF
--- a/metaphor/glue/extractor.py
+++ b/metaphor/glue/extractor.py
@@ -78,7 +78,7 @@ class GlueExtractor(BaseExtractor):
                 columns.append(
                     SchemaField(
                         field_name=column.get("Name"),
-                        field_path=column.get("Name"),
+                        field_path=column.get("Name").lower(),
                         native_type=column.get("Type"),
                         description=column.get("Comment"),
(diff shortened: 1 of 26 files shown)


metaphor/glue/extractor.py
```diff
@@ -78,7 +78,7 @@ def _get_columns(self, storageDescriptor: Any) -> List[SchemaField]:
                 columns.append(
                     SchemaField(
                         field_name=column.get("Name"),
-                        field_path=column.get("Name"),
+                        field_path=column.get("Name").lower(),
```
mars-lan, line 81: Perhaps we should consolidate the logic into a `build_schema_field` util method instead of littering the `lower()` everywhere?

alyiwang: updated

mars-lan (approved):

PR #905, opened 2024-07-18
[sc-27536] Allow ignoring `INSERT INTO ... VALUES` queries

usefulalgorithm:
### 🤔 Why?

`INSERT INTO ... VALUES` means we're inserting static, explicit values into a table. There's no lineage for these queries, so we should just exclude them from the crawled query logs.

### 🤓 What?

Allow ignoring `INSERT INTO ... VALUES` queries. By default these are ignored.

### 🧪 Tested?

Added unit test.

In our own environment there's no `INSERT INTO ... VALUES` query, so can't really test the crawler...

### ☑️ Checks

### DIFF
--- a/metaphor/common/docs/process_query.md
+++ b/metaphor/common/docs/process_query.md
@@ -8,7 +8,7 @@ process_query:
 
   redacted_literal_placeholder: <redacted literal placeholder>  # The redacted values will be replaced by this placeholder string. Default is '<REDACTED>'.
 
-  ignore_insert_values_into: <ignore_insert_values_into>  #  Ignore `INSERT INTO ... VALUES` expressions. These expressions don't have any lineage information, and are often very large in size. Default is `true`.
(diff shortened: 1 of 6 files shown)


metaphor/common/docs/process_query.md
```diff
@@ -7,8 +7,11 @@ process_query:
   redact_literal_values_in_where_clauses: <true | false>  # Whether to redact all literal values in WHERE clauses. Default is `false`.
 
   redacted_literal_placeholder: <redacted literal placeholder>  # The redacted values will be replaced by this placeholder string. Default is '<REDACTED>'.
+
+  ignore_insert_values_into: <ignore_insert_values_into>  #  Ignore `INSERT INTO ... VALUES` expressions. These expressions don't have any lineage information, and are often very large in size. Default is `true`.
```
mars-lan, line 11: Might want to set the default to false until we're certain the performance implication?

mars-lan (approved):

PR #906, opened 2024-07-18
[sc-27537] Redact PII in `MERGE ... WHEN NOT MATCHED INSERT`

usefulalgorithm:
### 🤔 Why?

Consider this SQL query:
```sql
    MERGE TargetProducts AS Target
    USING SourceProducts	AS Source
    ON Source.ProductID = Target.ProductID

    WHEN NOT MATCHED BY Target THEN
        INSERT (ProductID,ProductName, Price)
        VALUES (1,'awesome product', 0.1)  -- These might be PII, we should be able to redact them

    WHEN MATCHED THEN UPDATE SET
        Target.ProductName	= Source.ProductName,
        Target.Price		= Source.Price;
```

### 🤓 What?

- Support redacting explicit, static values in `WHEN NOT MATCHED INSERT` expression
- Refactor redact config

### 🧪 Tested?

Added unit test.

### ☑️ Checks

### DIFF
--- a/metaphor/common/docs/process_query.md
+++ b/metaphor/common/docs/process_query.md
@@ -11,7 +11,7 @@ process_query:
 
     placeholder_literal: <placeholder literal>  # The redacted values will be replaced by this placeholder string. Default is '<REDACTED>'.
 
-  ignore_insert_values_into: <true | false>  #  Ignore `INSERT INTO ... VALUES` expressions. These expressions don't have any lineage information, and are often very large in size. Default is `true`.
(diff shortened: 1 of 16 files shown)


metaphor/common/docs/process_query.md
```diff
@@ -4,14 +4,18 @@ Controls whether the crawler should process the SQL queries before storing the M
 
 ```yaml
 process_query:
-  redact_literal_values_in_where_clauses: <true | false>  # Whether to redact all literal values in WHERE clauses. Default is `false`.
+  redact_literals:
```
mars-lan, line 7: This is a breaking change but okay because no one is using this yet.

PR #908, opened 2024-07-18
Support Databricks shallow clone

elic-eon:
### 🤔 Why?

There are more table types we did not capture in the model, and we want to support them.

### 🤓 What?

- Support all table types
- Safely map from `UC.TableType` to `UnityCatalogTableType` in case there is more table type in the future.

### 🧪 Tested?

Unit tests. Verified against dev with shallow clone table.

### ☑️ Checks


alyiwang (approved):
LGTM

PR #915, opened 2024-07-20
Use system.query.history table to extract query logs from Unity Catalog

mars-lan:
### 🤔 Why?

It seems that using [Query history system table](https://docs.databricks.com/en/admin/system-tables/query-history.html) is the recommended way for batch extraction, as opposed to the [REST API](https://docs.databricks.com/api/workspace/queryhistory/list).

### 🤓 What?

- Refactor query extraction code to use `system.query.history` table.
- Drop system schema checks, as we should fail loudly if the required schemas are missing.
- Use filter.include_database method to properly include/exclude catalogs.
- Replace deprecated `logger.warn` with `logger.warning`.
- Update databricks-sql-connector to latest version.

TODO in a follow-up PR: Use sqlglot to extract the sources & targets for the query

### 🧪 Tested?

Verified against a production instance with 250k+ queries. Took < 1 min to complete (5x faster than REST API).

### ☑️ Checks


elic-eon (approved):
LGTM

PR #917, opened 2024-07-20
Run DESCRIBE HISTORY in parallel to improve performance

mars-lan:
### 🤔 Why?

Each [DESCRIBE HISTORY](https://docs.databricks.com/en/sql/language-manual/delta-describe-history.html) command takes 1~5 seconds to complete. The time adds up quickly when running against a large number of tables.

### 🤓 What?

- Batch run `DESCRIBE HISTORY` commands in a thread pool to improve performance.
- Add `create_connect_pool` util method for code sharing.
- Bump up `LIMIT` for `DESCRIBE HISTORY` from 50 to 100 since we're less contained by time.

### 🧪 Tested?

Verified against a production instance with ~3000 tables and saw close to 10x speed up (took 7 mins to get all last refresh dates).

### ☑️ Checks


alyiwang (approved):
LGTM

PR #918, opened 2024-07-20
Resolve service principal owners to their display names

mars-lan:
### 🤔 Why?

If a Unity Catalog table is owned by a group, the [API](https://docs.databricks.com/api/workspace/tables/list) will automatically resolve the group's name. However, if the table is owned by a service principal, the API returns its [application ID](https://docs.databricks.com/en/admin/users-groups/service-principals.html) instead. 

### 🤓 What?

Use the [API](https://docs.databricks.com/api/workspace/serviceprincipals/list) to build a list of service principals and resolve the application ID to the corresponding display name when processing the ownership info.

### 🧪 Tested?

### ☑️ Checks

### DIFF
--- a/tests/unity_catalog/expected.json
+++ b/tests/unity_catalog/expected.json
@@ -374,7 +374,7 @@
     "systemContacts": {
       "contacts": [
         {
-          "email": "user3@foo.com",
+          "email": "service principal 1",
           "systemContactSource": "UNITY_CATALOG"
         }
       ]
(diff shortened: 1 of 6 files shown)


tests/unity_catalog/expected.json
```diff
@@ -374,7 +374,7 @@
     "systemContacts": {
       "contacts": [
         {
-          "email": "user3@foo.com",
+          "email": "service principal 1",
```
alyiwang, line 377: Is this correct?

mars-lan: Yes, the API is mocked to return "sp1", which in turns get resolved into "service principal 1": https://github.com/MetaphorData/connectors/blob/4548d9494acfdc09474edd976bd473454cb6db57/tests/unity_catalog/test_extractor.py#L146

alyiwang (approved):
lgtm

PR #919, opened 2024-07-21
[sc-27604] Parse TLL for UnityCatalog query logs

usefulalgorithm:
### 🤔 Why?

UnityCatalog does not compute the sources and targets of each query log, so we need to do it ourselves.

The TLL parser is largely based on our in-house lineage parser, with some incompatible methods stripped off.

### 🤓 What?

- Implemented TLL parser to find sources and targets for UC queries.
- Look for fully qualified datasets if the found sources and targets are not fully qualified.
- Added `poetry export` step in CD's `docker` job. This step creates the `requirements.txt` file, which is populated during CD and is used to build the image.
- Modified Dockerfile to install dependencies first and then install the module. This reduces image build time drastically.

### 🧪 Tested?

- Run locally
- Tested end-to-end, took about 14 minutes to process 233k queries on andy.dev.metaphor.io
截圖 2024-07-22 上午8 50 36

### ☑️ Checks


metaphor/unity_catalog/extractor.py
```diff
@@ -459,38 +458,14 @@ def collect_query_logs(self) -> Iterator[QueryLog]:
 
         logger.info(f"Fetching queries from the last {lookback_days} days")
 
-        count = 0
-        for row in list_query_log(
+        for query_log in get_query_logs(
             self._connection,
             self._query_log_config.lookback_days,
             self._query_log_config.excluded_usernames,
+            self._max_concurrency,
```
mars-lan, line 465: Might want to add a separate config for this. 10 is an overkill for CPU-bound tasks.

mars-lan: Or better yet autodetect the number of cores

mars-lan:
Pls also measure the time it takes for max_workers=2 to simulate the typical crawler environment.

.pre-commit-config.yaml
```diff
@@ -60,3 +60,11 @@ repos:
...
         pass_filenames: false
         always_run: true
+
+  - repo: https://github.com/python-poetry/poetry
+    rev: 1.8.3
+    hooks:
+      - id: poetry-lock
+        args: [--no-update]
+      - id: poetry-export
+        args: [--all-extras, --output=requirements.txt]
```
mars-lan, line 70: This takes a long time? If not, would prefer generating this on-the-fly during Docker build time.

usefulalgorithm: if you don't change any dependency then this finishes instantly

usefulalgorithm: i changed it so that it exports to requirements.txt in cd before building the img

.github/workflows/ci.yml
```diff
@@ -77,8 +77,28 @@ jobs:
...
+
+      - name: Setup Poetry
+        uses: abatilo/actions-poetry@v2.1.4
+        with:
+          poetry-version: ${{ env.POETRY_VERSION }}
+
+      - name: Export requirements.txt
+        run: poetry export -E all -f requirements.txt --output requirements.txt
+
+      - name: ls
```
mars-lan, line 96: ?

PR #920, opened 2024-07-22
Improve unity catalog profiler speed part 2

elic-eon:
### 🤔 Why?

Running `ANALYZE TABLE` every time is too costly, add config flag to en/disable it

### 🤓 What?

- Creat a connection pool after getting all the table
- Don't run `ANALYZE` by default
- Only profile tables that the table type is in the white list

### 🧪 Tested?

Test against prod env

### ☑️ Checks


metaphor/unity_catalog/profile/README.md
```diff
@@ -40,6 +40,15 @@ The max number of concurrent queries to the databricks compute node can be confi
 max_concurrency: <max_number_of_queries> # Default to 10
 ```
 
+#### Analyze table
+
+To run `ANALYZE TABLE` query before the connector getting the statistics, or only for tables have no statistics.
+
+```yaml
+analyze: true # Default is false
```
mars-lan, line 48: Not sure why you need 2 flags? I can only think of 3 situations:

1. Has pre-cached statistics available, return immediately (ignore config)
2. No pre-cached statistics available, run analyze (`analyze_if_no_statistics` = true)
3. No pre-cached statistics available, skip analyze (`analyze_if_no_statistics` = false)

elic-eon: 1. We can still run analyze if we want to update the cached statistics
2. run analyze
3. skip analyze, getting null value

elic-eon: Basically, with `analyze=true` will behave like previous implementation.

mars-lan: Also update README?

metaphor/unity_catalog/profile/extractor.py
```diff
@@ -130,9 +148,11 @@ def _get_tables(self) -> List[TableInfo]:
                 ]
                 for table_info in table_infos:
                     table_type = table_info.table_type
-                    if table_type is None or table_type in (
-                        TableType.VIEW,
-                        TableType.MATERIALIZED_VIEW,
+                    if table_type not in (
+                        TableType.EXTERNAL,
```
mars-lan, line 152: Let's define this as a const at the top for easier modification in the future.

mars-lan (commented):
Otherwise LGTM

mars-lan (commented):
Otherwise LGTM

mars-lan (approved):

PR #921, opened 2024-07-22
bugfix: Drop partially qualified dataset found during UC TLL

usefulalgorithm:
### 🤔 Why?

`QueriedDataset(db="", schema="", table="something")` causes the MCE validation to fail in our server application.

If a dataset isn't fully qualified, it is not an actually existing dataset. We'll just drop those from the TLL.

### 🤓 What?

- Drop partially qualified dataset found during UC TLL.

### 🧪 Tested?

Modified unit test.

### ☑️ Checks

### DIFF
--- a/metaphor/common/sql/table_level_lineage/table.py
+++ b/metaphor/common/sql/table_level_lineage/table.py
@@ -16,8 +16,8 @@ class Table:
     @classmethod
     def from_sqlglot_table(cls, table: exp.Table):
         return cls(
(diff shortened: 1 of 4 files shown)


mars-lan (commented):
Add test?

usefulalgorithm:
> Add test?

there's already a test for `find_qualified_dataset`, I modified it to match this fix

mars-lan (approved):

PR #922, opened 2024-07-22
Map service principals to user_id field

mars-lan:
### 🤔 Why?

Similar to https://github.com/MetaphorData/connectors/pull/918, the service principal IDs should be resolved to the display names and stored in the correct field (`user_id` instead of `email`) in query logs.

### 🤓 What?

As titled.

### 🧪 Tested?

### ☑️ Checks


alyiwang (approved):
LGTM

PR #923, opened 2024-07-22
Use safe_float in UC profiler to avoid outputting INF in MCEs

mars-lan:
### 🤔 Why?

Unity Catalog profiler can output INF in MCEs, causing validation errors.

### 🤓 What?

As titled.

### 🧪 Tested?

### ☑️ Checks


alyiwang (approved):
LGTM

PR #924, opened 2024-07-23
Exception is not handled causing the crawler stuck 

elic-eon:
### 🤔 Why?

We try to format `None` as a float which will raise an exception in the threaded function. So the crawler stuck

### 🤓 What?

- Handle the exception
- Fix the logging format

### 🧪 Tested?

- Change the code to raise a exception in the helper function, and the crawler can finish
- Test against prod 
```
2024-07-23 14:34:31:INFO:metaphor:Ended running with RunStatus.SUCCESS at 2024-07-23 14:34:31.223049, fetched 2169 entities, took 2654.9s
```

### ☑️ Checks

### DIFF
--- a/metaphor/unity_catalog/profile/extractor.py
+++ b/metaphor/unity_catalog/profile/extractor.py
@@ -125,10 +125,14 @@ class UnityCatalogProfileExtractor(BaseExtractor):
 
         entities = []
         with ThreadPoolExecutor(max_workers=connection_pool.maxsize) as executor:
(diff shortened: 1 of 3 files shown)


mars-lan (approved):

mars-lan (changes requested):
Bump version

elic-eon:
> Bump version

Bumped the version

mars-lan (approved):

PR #926, opened 2024-07-23
[sc-27549] Switch to using DiscoveryAPI for dbt.cloud

usefulalgorithm:
### 🤔 Why?

DBT's discovery API exposes more information than manifest.json, plus it's more flexible for us to control what we want to get instead of having to download the entire file every time.

### 🤓 What?

- Switch to using discovery API for dbt.cloud extractor.
- Added script to download `schema.graphql` from DBT's Apollo server and generate graphql client code automatically.
- Updated CI to respect coverage configurations in `pyproject.toml`
- Ignore all generated files in coverage reporting.
- Updated pre-commit hook to ignore bandit issues in generated files

#### Difference between using manifest.json and using discovery api

- With manifest.json we could not get column types most of the time. With discovery api it is always available (at least during the tests i've run)
- Discovery API is picking up columns that are not covered by parsing manifest.json. These models don't have any field when parsed with manifest.json, but have all the fields when done with DiscoveryAPI:
  - jaffle_shop.metricflow_time_spine
  - jaffle_shop.order_items
  - london_bike_analysis.cleaned_bike_rides_from_snapshot
  - london_bike_analysis.cycle_hire_snapshot
  - london_bike_analysis.raw_bike_hires
  - london_bike_analysis.rides_by_month_2017
  - london_bike_analysis.rides_by_month_start_station_2017
- Parsing with manifest.json does not support metrics for v10 and up. With DiscoveryAPI metrics is available.

Script used for comparing MCEs: https://pastebin.com/zc9zE2GG 

### 🧪 Tested?

- Tested locally with staging configuration
- Tested with production configuration
- Unit test

### ☑️ Checks


mars-lan:
Let's exclude `metaphor/dbt/cloud/discovery_api/*` from coverage to avoid a big drop?

metaphor/common/entity_id.py
```diff
@@ -121,3 +121,17 @@ def dataset_normalized_name(
     return normalize_full_dataset_name(
         ".".join([part for part in [db, schema, table] if part is not None])
     )
+
+
+def parts_to_dataset_entity_id(
```
mars-lan, line 126: Pls add doc

metaphor/dbt/cloud/discovery_api/graphql_client/__init__.py
mars-lan, line 1: If possible, output all generated files in a `generated` directory so it's obvious

usefulalgorithm:
> Let's exclude `metaphor/dbt/cloud/discovery_api/*` from coverage to avoid a big drop?

coverage on generated files looks alright, it's the parser not parsing macros and sources for jaffle_shop. I'll add a new test case today.

.pre-commit-config.yaml
```diff
@@ -44,7 +44,7 @@ repos:
     rev: 1.7.8
     hooks:
       - id: bandit
-        args: ['--skip=B101,B106,B404,B603,B607,B608']
+        args: ['--skip=B101,B106,B404,B603,B607,B608', -x, metaphor/dbt/cloud/discovery_api/generated]
```
mars-lan, line 47: Doesn't read these from pyproject.toml?

metaphor/dbt/cloud/discovery_api/codegen.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+APOLLO_IOS_CLI_VERSION=1.14.0
```
mars-lan, line 3: Consider adding a comment why ios for posterity?

.pre-commit-config.yaml
usefulalgorithm: no, pre-commit bandit hook uses these arguments instead of pyproject.toml

metaphor/dbt/cloud/discovery_api/codegen.sh
usefulalgorithm: the apollo program is called `apollo-ios-cli`, I don't know why it's ios either

mars-lan: Please add a comment to avoid future confusion.

mars-lan: I also saw swift-related config in apollo-codegen-config.json, which suggests that this is indeed for iOS. I assume there are no Mac/Linux/Python-specific versions?

usefulalgorithm: I think this is the only one.

.pre-commit-config.yaml
mars-lan: Would this work? https://github.com/PyCQA/bandit/issues/902#issuecomment-1132318369

usefulalgorithm: I modified it, now it should pick up pyproject.toml

elic-eon (dismissed):
Let's go!

elic-eon (approved):

usefulalgorithm:
ci is failling because setuptools + poetry is broken:

https://github.com/pypa/setuptools/issues/4519
https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/16289

elic-eon (approved):

.github/workflows/cd.yml
```diff
@@ -75,6 +75,8 @@ jobs:
 
       - name: Test
         run: |
+          echo "setuptools<72" > constraints.txt
```
mars-lan, line 78: Explicitly adding setuptool to pyproject.toml doesn't work?

usefulalgorithm: I've tried that, it wouldn't work

usefulalgorithm: I think the reason is because it doesn't matter if we're explicitly setting setuptools version in pyproject.toml, when poetry is building pyhive it's going to use whatever version that's specified in pyhive's dependencies, and once they resolve it to use setuptools>=72 it will break.

I have tried this both in CI and locally, none of which worked.

PR #936, opened 2024-08-04
Add account_usage config to query a alternate schema

mars-lan:
### 🤔 Why?

Granting access to the entire SNOWFLAKE database may not always be desirable. The connector should be able to query from a different schema that mirrors the minimally needed views (QUERY_HISTORY, ACCESS_HISTORY, TAG_REFERENCES).

### 🤓 What?

As titled.

### 🧪 Tested?

Verified against a production instance with mirrored views.

### ☑️ Checks


elic-eon (approved):
LGTM!

PR #939, opened 2024-08-06
[sc-27547] Point embedded datasource url to workbook

usefulalgorithm:
### 🤔 Why?

An embedded datasource is always in a workbook, so for its URL we just use the workbook's.

### 🤓 What?

For embedded tableau datasources, their url should point to the workbooks containing them.

### 🧪 Tested?

- Updated unit test.
- Ran crawler locally.

### ☑️ Checks

### DIFF
--- a/metaphor/tableau/extractor.py
+++ b/metaphor/tableau/extractor.py
@@ -525,6 +525,7 @@ class TableauExtractor(BaseExtractor):
                         custom_sql_source.account if custom_sql_source else None
                     ),
                     source_datasets=source_datasets or None,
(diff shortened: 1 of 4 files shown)


alyiwang (approved):

metaphor/tableau/extractor.py
```diff
@@ -525,6 +525,7 @@ def _parse_workbook_query_response(
                         custom_sql_source.account if custom_sql_source else None
                     ),
                     source_datasets=source_datasets or None,
+                    url=f"{self._base_url}/workbooks/{workbook.vizportalUrlId}",
```
mars-lan, line 528: Not sure this alone will work. Should use `source_info.main_url` instead?

PR #940, opened 2024-08-06
[sc-27955] QueryLog.platform needs to be a valid value for Trino

usefulalgorithm:
### 🤔 Why?

Without it, Metaphor app would fail to ingest the record.

### 🤓 What?

Added platform=DataPlatform.TRINO when creating QueryLog in trino extractor.

### 🧪 Tested?

N/A.

### ☑️ Checks

### DIFF
--- a/metaphor/trino/extractor.py
+++ b/metaphor/trino/extractor.py
@@ -109,6 +109,7 @@ class TrinoExtractor(BaseExtractor):
                     else None
                 ),
                 user_id=user,
+                platform=DataPlatform.TRINO,
             )
             yield log
             query_log_count += 1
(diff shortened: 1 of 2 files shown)


usefulalgorithm:
> Surprised that the JSON schema validator didn't catch this? Or maybe it was only applied to non-query log MCEs?

Yeah query logs aren't validated at the moment :/

mars-lan:
> > Surprised that the JSON schema validator didn't catch this? Or maybe it was only applied to non-query log MCEs?
> 
> Yeah query logs aren't validated at the moment :/

Please add a ticket to track that.

PR #949, opened 2024-08-13
Pull catalog schema ownership info from unity catalog

elic-eon:
### 🤔 Why?

To pull the owner of catalog and schema as a system contact

### 🤓 What?

- Fill `systemContacts` for hierarchies
- DRY refactor for getting service principal display name
- DRY refactor for initialize a hierarchy

### 🧪 Tested?

Unit tests and verified the generated MCEs are ingested without error

### ☑️ Checks


metaphor/unity_catalog/extractor.py
```diff
@@ -467,34 +492,31 @@ def collect_query_logs(self) -> Iterator[QueryLog]:
         ):
             yield query_log
 
+    def _assert_hierarchy(
```
usefulalgorithm, line 495: this looks a lot like those `init_dataset` methods in other crawlers, name it `_init_hierarchy` instead?

mars-lan (approved):

PR #958, opened 2024-08-21
Add validator to catch datetime azure_openai.version

rishimo:
### 🤔 Why?

If the `azure_openai.version` set in the app is a regular version e.g. `2024-06-01` instead of a `-preview` version it can be misinterpreted as a datetime when the config is loaded, causing validation errors.

### 🤓 What?

Adds a validator to force datetime to string to catch pesky validation errors.

### 🧪 Tested?

Connector tested locally with a YAML file generated by `app`, runs successfuly.

### ☑️ Checks


mars-lan (approved):
YAML FTW!

mars-lan (approved):

PR #959, opened 2024-08-22
[sc-28233] Redacted values should be strings

usefulalgorithm:
### 🤔 Why?

If a literal value should be redacted, we use the placeholder string `<REDACTED>`. However, if we are redacting a numeric value we need to turn it into a string, otherwise SQLGlot will think `<` is LT, and would not parse the thing.

For example, SQLGlot will not parse this:
```sql
INSERT INTO target_table (first_name, last_name, email, status)
SELECT
    first_name,
    last_name,
    email,
    CASE
        WHEN age < <REDACTED> THEN '<REDACTED>'
        WHEN age >= <REDACTED> AND age < <REDACTED> THEN '<REDACTED>'
        ELSE '<REDACTED>'
    END AS status
FROM source_table
WHERE NOT email IS NULL
```

On the other hand, this works:
```sql
INSERT INTO target_table (first_name, last_name, email, status)
SELECT
    first_name,
    last_name,
    email,
    CASE
        WHEN age < '<REDACTED>' THEN '<REDACTED>'
        WHEN age >= '<REDACTED>' AND age < '<REDACTED>' THEN '<REDACTED>'
        ELSE '<REDACTED>'
    END AS status
FROM source_table
WHERE NOT email IS NULL
```
```

### 🤓 What?

<!--
  Summary of the changes committed. How does your PR fix the above issue?
-->

- When redacting a literal, set its `is_string` arg to `True`.
- Refactored tests.

### 🧪 Tested?

<!--
  Describe how the change was tested end-to-end.
-->

Modified unit tests.

### ☑️ Checks

<!--
  Some sanity checks to make sure your PR is good to go. Make sure all checkboxes
  are ticked!
-->

### DIFF
--- a/metaphor/common/sql/process_query/process_query.py
+++ b/metaphor/common/sql/process_query/process_query.py
@@ -19,6 +19,7 @@ def _redact_literal_values_in_where_clauses(
     for where in expression.find_all(exp.Where):
         for lit in where.find_all(exp.Literal):
             lit.args["this"] = config.redact_literals.placeholder_literal
+            lit.args["is_string"] = True
 
 
 def _redact_literal_values_in_case_clauses(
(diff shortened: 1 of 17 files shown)


elic-eon (approved):
LGTM!

Issue #858, opened 2024-05-06
Breaking changes in 0.15


mars-lan:
- Rename all `snowflake_account` config to `account`

usefulalgorithm:
Drop `where_clauses`, `case_clauses` and `when_not_matched_insert_clauses` configs in `RedactPIILiteralsConfig` in favor of `redact`

PR #961, opened 2024-08-22
[sc-28238] Just redact all literals in the query, and don't print the entire query in log

usefulalgorithm:
### 🤔 Why?

- Printing the entire query in the logs when Sqlglot can't parse it makes it look very scary especially when the query is very very long. Instead of printing the entire thing, just print where it failed to be parsed and its id if there is one.
- When we are redacting literals we're actually traversing the query 3 times (looking for `WHEN`, `WHERE` and `CASE` expressions, each is a BFS). Might as well just define a new Sqlglot dialect that turns whatever literal into `'<REDACTED>'`, so that we don't traverse the query multiple times.

### 🤓 What?

- Only query id and exact location of failure are printed in the logs.
- Instead of traversing the query multiple times, use a dialect that redacts literals to speed things up.
- Replaced config for where, when and case in process query configs with a single switch.
- Updated process_query.md.

### 🧪 Tested?

Existing unit tests are passing.

### ☑️ Checks


metaphor/common/sql/process_query/config.py
```diff
@@ -11,20 +11,7 @@ class RedactPIILiteralsConfig:
     Config to control whether we want to redact literal values. Useful if you want to remove PII from your queries.
     """
 
-    where_clauses: bool = False
```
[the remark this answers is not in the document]
mars-lan: And add an entry to https://github.com/MetaphorData/connectors/issues/858 for eventual removal

metaphor/common/docs/process_query.md
```diff
@@ -5,11 +5,7 @@ Controls whether the crawler should process the SQL queries before storing the M
 ```yaml
 process_query:
   redact_literals:
-    where_clauses: <true | false>  # Whether to redact all literal values in WHERE clauses. Default is `false`.
-
-    case_clauses: <true | false>  # Whether to redact all literal values in CASE clauses. Default is `false`.
-
-    when_not_matched_insert_clauses: <true | false>  # Whether to redact literal values in WHEN NOT MATCHED INSERT clauses. If set to `True`, all literal values will be redacted to a predefined string value. Default is `false`.
+    redact: <true | false> # Whether to redact the literal values. Default it `false`.
```
mars-lan, line 8:
```suggestion
    enabled: <true | false> # Whether to redact the literal values. Default it `false`.
```

metaphor/common/sql/process_query/config.py
```diff
@@ -11,23 +15,32 @@ class RedactPIILiteralsConfig:
...
 
+    redact: bool = False
+
     placeholder_literal: str = "<REDACTED>"
 
+    @field_validator("where_clauses", "case_clauses", "when_not_matched_insert_clauses")
+    @classmethod
+    def _warn_deprecated_fields(cls, value: bool, info: ValidationInfo) -> bool:
+        if value:
+            logger.warning(f"Deprecated field set to True: {info.field_name}")
```
mars-lan, line 41:
```suggestion
            logger.warning(f"Deprecated config {info.field_name} will be removed in v0.15.")
```

mars-lan (commented):
Otherwise LGTM

metaphor/common/docs/process_query.md
```diff
@@ -18,6 +14,5 @@ process_query:
 
 If any of the following boolean values is set to true, crawler will process the incoming SQL queries:
 
-- `redact_literals.where_clauses`
-- `redact_literals.when_not_matched_insert_clauses`
+- `redact_literals.redact`
```
mars-lan, line 17:
```suggestion
- `redact_literals.enabled`
```

mars-lan (commented):
Final nit.

mars-lan (approved):
Go go go!

PR #962, opened 2024-08-23
dbt cloud crawler is not extract test status

elic-eon:
### 🤔 Why?

The status of the test run is stored in the `tests.status`, and our current implementation uses the model run status.

### 🤓 What?

- Modify the gql query to get test status.
- Refactor `_parse_test_run_result`

### 🧪 Tested?

Verified against dbt projects. Also, disable `Generate doc on run` option in the dbt job setting, otherwise, the status will always be success

### ☑️ Checks


usefulalgorithm (approved):
Good catch, thanks!

PR #966, opened 2024-08-27
Not able to parse char length for a column

elic-eon:
### 🤔 Why?

"character varying" is a valid type string we should throw warning log on this.

### 🤓 What?

- Better parsing char length and precision

### 🧪 Tested?

Unit test updated.

### ☑️ Checks

### DIFF
--- a/metaphor/postgresql/extractor.py
+++ b/metaphor/postgresql/extractor.py
@@ -1,6 +1,8 @@
-from typing import Callable, Collection, Dict, List, Optional, Tuple
+import re
+from typing import Collection, Dict, List, Optional, Tuple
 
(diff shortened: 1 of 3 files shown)


mars-lan (commented):
Can we use the opportunity to add some tests?

elic-eon:
> Can we use the opportunity to add some tests?

test for parsing? we do have

mars-lan:
> > Can we use the opportunity to add some tests?
> 
> test for parsing? we do have

My bad. The other parts of Postgres crawler have very low coverage.

mars-lan (approved):

mars-lan (commented):
Bump version?

usefulalgorithm (approved):

PR #969, opened 2024-08-28
[sc-28404] Add config to drop or pass through unparseable queries

usefulalgorithm:
### 🤔 Why?

If we can't parse some queries, we want to be able to decide whether we want these queries to pass thru or be ignored altogether.

### 🤓 What?

Added config to control whether to skip or pass through unparseable queries.

### 🧪 Tested?

Added unit test.

### ☑️ Checks

### DIFF
--- a/metaphor/common/docs/process_query.md
+++ b/metaphor/common/docs/process_query.md
@@ -11,7 +11,7 @@ process_query:
 
   ignore_insert_values_into: <true | false>  # Ignore `INSERT INTO ... VALUES` expressions. These expressions don't have any lineage information, and are often very large in size. Default is `false`.
 
-  skip_unparsable_queries: <true | false>  # If this is set to `True`, when Sqlglot fails to parse a query we skip it from the collected MCE. Default is `false`, meaning we pass through any query we are unable to parse.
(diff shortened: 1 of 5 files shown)


metaphor/common/docs/process_query.md
```diff
@@ -9,7 +9,9 @@ process_query:
 
     placeholder_literal: <placeholder literal>  # The redacted values will be replaced by this placeholder string. Default is '<REDACTED>'.
 
-  ignore_insert_values_into: <true | false>  #  Ignore `INSERT INTO ... VALUES` expressions. These expressions don't have any lineage information, and are often very large in size. Default is `false`.
+  ignore_insert_values_into: <true | false>  # Ignore `INSERT INTO ... VALUES` expressions. These expressions don't have any lineage information, and are often very large in size. Default is `false`.
+
+  skip_unparsable_queries: <true | false>  # If this is set to `True`, when Sqlglot fails to parse a query we skip it from the collected MCE. Default is `false`, meaning we pass through any query we are unable to parse.
```
elic-eon, line 14:
```suggestion
  skip_unparsable_queries: <true | false>  # If this is set to `true`, when Sqlglot fails to parse a query we skip it from the collected MCE. Default is `false`, meaning we pass through any query we are unable to parse.
```

elic-eon (commented):
Otherwise, LGTM!

elic-eon (approved):

PR #973, opened 2024-09-02
[sc-28469] Enable `process_query` for BQ, Postgres, Redshift and UC

usefulalgorithm:
### 🤔 Why?

We added SQL preprocessor, but currently only Snowflake crawler makes use of it. The other data platform that we get query history from should also have the ability to preprocess their queries.

### 🤓 What?

Add `process_query` function call in the following crawler's `collect_query_logs` method:
- BigQuery
- Redshift
- Postgresql
- Unity Catalog

### 🧪 Tested?

- Modified UC unit test
- Run all four crawlers with stage configuration with `redact_literals.enabled` set to True, all run without error and the literals are all redacted.

### ☑️ Checks

### DIFF
--- a/metaphor/bigquery/README.md
+++ b/metaphor/bigquery/README.md
@@ -157,6 +157,10 @@ query_log:
   fetch_job_query_if_truncated: <boolean>
 ```
 
+##### Process Query Config
+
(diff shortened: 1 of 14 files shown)


elic-eon (approved):
LGTM

PR #974, opened 2024-09-02
Fix: MCE validation error

elic-eon:
### 🤔 Why?

Encounter MCE validation when processing the query log.

### 🤓 What?

The `database` and `schema` can not be empty strings or null so add a function to drop query logs that don't have `schema`

### 🧪 Tested?

So MCE error after the patch

### ☑️ Checks

### DIFF
--- a/metaphor/common/sql/utils.py
+++ b/metaphor/common/sql/utils.py
@@ -8,6 +8,6 @@ def is_valid_queried_datasets(datasets: List[QueriedDataset]) -> bool:
     database and schema must exist and not an empty string
     """
     for dataset in datasets:
-        if not dataset.database or not dataset.schema:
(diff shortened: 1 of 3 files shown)


metaphor/common/sql/utils.py
```diff
@@ -0,0 +1,13 @@
...
+
+from metaphor.models.metadata_change_event import QueriedDataset
+
+
+def is_valid_queried_datasets(datasets: List[QueriedDataset]) -> bool:
+    """
+    database and schema must exist and not an empty string
+    """
+    for dataset in datasets:
+        if not dataset.database or not dataset.schema:
```
alyiwang, line 11: might as well check the table name?

usefulalgorithm (approved):

PR #978, opened 2024-09-05
[sc-27944] Do not populate system tags when there is no tag for entity

usefulalgorithm:
### 🤔 Why?

Having `systemTags.tags` as empty array in a crawler will make Metaphor ingestion overwrite the existing system tags found by other crawlers.

### 🤓 What?

If no system tags exist for an entity, do not populate its `systemTags` field.

### 🧪 Tested?

Modified unit test.

### ☑️ Checks


mars-lan (commented):
The original reason we did this is to handle tags removal, i.e. an asset had some tags that were later removed. Would this be handled correctly?

PR #984, opened 2024-09-18
[sc-28750] Set `ignore_command_statement` default to `False`

usefulalgorithm:
### 🤔 Why?

`ignore_command_statement` was defaulted to True, which made all crawlers that parses SQL queries in all environments started processing queries.

Instead of doing it by default, our users should explicitly opt-in to query processing.

The actual SQL query that could not be parsed by sqlglot is tracked in a [separate ticket](https://app.shortcut.com/metaphor-data/story/28750/metaphor-crawler-big-query-integrations-error)

### 🤓 What?

- Set `ignore_command_statement` default to `False`
- Modified unit test
- Updated readme

### 🧪 Tested?

Modified unit test.

### ☑️ Checks

### DIFF
--- a/metaphor/postgresql/config.py
+++ b/metaphor/postgresql/config.py
@@ -20,7 +20,9 @@ class QueryLogConfig:
 
     # Config to control query processing
     process_query: ProcessQueryConfig = field(
-        default_factory=lambda: ProcessQueryConfig()
+        default_factory=lambda: ProcessQueryConfig(
+            ignore_command_statement=True
+        )  # Ignore COMMAND statements by default
     )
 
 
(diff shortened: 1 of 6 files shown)


metaphor/postgresql/config.py
```diff
@@ -20,7 +20,9 @@ class QueryLogConfig:
 
     # Config to control query processing
     process_query: ProcessQueryConfig = field(
-        default_factory=lambda: ProcessQueryConfig()
+        default_factory=lambda: ProcessQueryConfig(
+            ignore_command_statement=True
```
mars-lan, line 24: Why?

usefulalgorithm: @elic-eon  this is the intention right?

elic-eon (approved):
Gogogo

PR #985, opened 2024-09-19
QuickSight crawler

elic-eon:
### 🤔 Why?

To extract QuickSight dataset and dashboard information

### 🤓 What?

- Extract QuickSight DataSet as virtual views with CLL (not include `CustomSql`)
- Extract QuickSight Dashboard

### 🧪 Tested?

Tested end-to-end on local (UI is still in progress)

image

### ☑️ Checks


metaphor/quick_sight/cll.py
```diff
@@ -0,0 +1,312 @@
...
+
+    data_platform = DATA_SOURCE_PLATFORM_MAP.get(data_source.Type)
+    if not data_platform:
+        return None
+
+    database = get_database(data_source)
+    account = get_account(data_source)
+    query = custom_sql.SqlQuery
+
+    tll = extract_table_level_lineage(
```
usefulalgorithm, line 108: We can probably just use `RelationalTable` as the source?

elic-eon: The `RelationalTable` and `CustomSql` are mutually exclusive. But we still want table-level lineage for a `CustomSql` table.

elic-eon: Update the logic, should be easier to understand now.

elic-eon:
Change this PR to draft because we need to process lineage on the server side

mars-lan:
> Change this PR to draft because we need to process lineage on the server side

Let's get all other parts in first and handle the lineage in a follow-up PR?

metaphor/quick_sight/extractor.py
```diff
@@ -0,0 +1,246 @@
...
+    _description = "Quick Sight metadata crawler"
+    _platform = Platform.QUICK_SIGHT
+
+    @staticmethod
+    def from_config_file(config_file: str) -> "QuickSightExtractor":
+        return QuickSightExtractor(QuickSightRunConfig.from_yaml_file(config_file))
+
+    def __init__(self, config: QuickSightRunConfig) -> None:
+        super().__init__(config)
+        self._datasets: Dict[str, Dataset] = {}
```
usefulalgorithm, line 72: don't need this anymore

metaphor/quick_sight/extractor.py
```diff
@@ -0,0 +1,246 @@
...
+        entities.extend(self._extract_dashboard())
+
+        return entities
+
+    def _get_resources(self):
+        self._get_dataset_detail()
+        self._get_dashboard_detail()
+        self._get_data_source_detail()
+
+    def _extract_virtual_view(self):
```
usefulalgorithm, line 102:
```suggestion
    def _extract_virtual_views(self):
```

metaphor/quick_sight/extractor.py
```diff
@@ -0,0 +1,246 @@
...
+
+        self._get_resources()
+
+        self._extract_virtual_view()
+
+        entities: List[ENTITY_TYPES] = []
+        entities.extend(self._virtual_views.values())
+        entities.extend(self._extract_dashboard())
+
+        return entities
```
usefulalgorithm, line 95:
```suggestion
        return self._extract_virtual_views() + self._extract_dashboards()
```

`_extract_dashboard` returns a list of dashboards, but `_extract_virtual_view` returns a None. I think we want to have both return a list of entities, and then return these two lists combined.

metaphor/quick_sight/extractor.py
```diff
@@ -0,0 +1,246 @@
...
+                    created_at_source=data_set.CreatedTime,
+                    last_updated=data_set.LastUpdatedTime,
+                ),
+            )
+
+            process_dataset_column_lineage(self._resources, data_set, view)
+
+            self._virtual_views[data_set.Arn] = view
+
+    def _extract_dashboard(self) -> List[MetaphorDashboard]:
```
usefulalgorithm, line 123:
```suggestion
    def _extract_dashboards(self) -> List[MetaphorDashboard]:
```

metaphor/quick_sight/extractor.py
```diff
@@ -0,0 +1,246 @@
...
+                    type=VirtualViewType.QUICK_SIGHT,
+                ),
+                structure=AssetStructure(name=data_set.Name),
+                source_info=SourceInfo(
+                    created_at_source=data_set.CreatedTime,
+                    last_updated=data_set.LastUpdatedTime,
+                ),
+            )
+
+            process_dataset_column_lineage(self._resources, data_set, view)
```
usefulalgorithm, line 119: I think instead of making the virtual view up front and then have `process_dataset_column_lineage` populate `entity_upstream`, it is easier to follow to put this in a `_init_virtual_view` and then have it fill out fields in a virtual view.

metaphor/quick_sight/cll.py
```diff
@@ -0,0 +1,313 @@
...
+
+            _process_transformation(columns, logical_table.DataTransforms or [])
+
+            tables[table_id] = columns
+
+        logical_tables = unresolved
+    return columns
+
+
+def process_dataset_column_lineage(
```
usefulalgorithm, line 255:
```suggestion
def process_dataset_lineage(
```

It is also doing TLL

metaphor/quick_sight/extractor.py
usefulalgorithm: sure, that sounds good.

metaphor/quick_sight/cll.py
usefulalgorithm: And rename the file from cll.py to lineage.py ?

usefulalgorithm (approved):
讚讚

PR #986, opened 2024-09-23
[sc-28901] Validate query logs as they're parsed

usefulalgorithm:
### 🤔 Why?

We should validate the query logs as they're parsed, otherwise bugs are harder to find when they get to the lineage parser.

### 🤓 What?

Validate the parsed query logs before writing them to file.

Tested this against our own Snowflake instance (13,589 queries)
- No validation: ﻿36 secs
截圖 2024-09-23 中午12 53 10

- Validated: 1 min 37 secs
截圖 2024-09-23 中午12 51 24

### 🧪 Tested?

Tested against
- BigQuery
- Snowflake (see above)
- Unity Catalog

### ☑️ Checks

### DIFF
--- a/metaphor/common/file_sink.py
+++ b/metaphor/common/file_sink.py
@@ -99,14 +99,17 @@ class QueryLogSink:
             )
 
     def _finalize_current_mce(self) -> None:
-        self._mces.append(EventUtil.build_then_trim(QueryLogs(logs=self._logs)))
+        event = EventUtil.build_then_trim(QueryLogs(logs=self._logs))
+        if EventUtil().validate_message(event) is None:
+            raise ValueError(f"Cannot validate query log batch: {event}")
+        self._mces.append(event)
         self._logs_count = 0
         self._logs.clear()
         self._mces_count += 1
         self.total_mces_wrote += 1
 
     def _finalize_current_batch(self) -> None:
-        # No need to validate mce
+        # self._mces is already validated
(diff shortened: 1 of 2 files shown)


elic-eon (commented):
Didn't we validate them when ingestion?

usefulalgorithm:
> Didn't we validate them when ingestion?

No, ingestion does not look at what's in the query log files.

elic-eon:
> > Didn't we validate them when ingestion?
> 
> No, ingestion does not look at what's in the query log files.

Got it, can we also valid entities in the MCE as well? We encounter the MCE validation error when ingestion many times, if we can catch the error earlier, it is better.

usefulalgorithm:
> > > Didn't we validate them when ingestion?
> > 
> > 
> > No, ingestion does not look at what's in the query log files.
> 
> Got it, can we also valid entities in the MCE as well? We encounter the MCE validation error when ingestion many times, if we can catch the error earlier, it is better.

MCEs are partially validated: https://github.com/MetaphorData/connectors/blob/main/metaphor/common/sink.py#L23C1-L28C14

It only validates against the json schema though, so it's kind of limited.

elic-eon (approved):

usefulalgorithm:
@elic-eon  @mars-lan  let's not merge this, it takes too long, almost 2x the original time.

Without: 6 mins a batch
main

With: 14 mins a batch
with

PR #987, opened 2024-09-24
[sc-28944] Provide config to link user name to email

usefulalgorithm:
### 🤔 Why?

It's not possible to get the correspondence between user names and emails from Redshift or Postgres, we need user to provide this as a config value.

### 🤓 What?

Add config field `username_to_email` in query log conifgs.

### 🧪 Tested?

- Modified unit tests.
- Ran crawler locally.

### ☑️ Checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
           poetry run isort --check .
           poetry run mypy . --explicit-package-bases
           poetry run bandit -r . -c pyproject.toml
-2
+
       - name: Test
         run: |
           poetry run coverage run -m pytest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
           poetry run isort --check .
           poetry run mypy . --explicit-package-bases
           poetry run bandit -r . -c pyproject.toml
-
+2
       - name: Test
         run: |
           poetry run coverage run -m pytest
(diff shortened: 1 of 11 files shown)


metaphor/postgresql/extractor.py
```diff
@@ -498,12 +498,15 @@ def _process_cloud_watch_log(
 
         if sql:
             sql_hash = md5_digest(sql.encode("utf-8"))
+            user_id = parsed.user
+            email = self._query_log_config.username_to_email.get(user_id)
```
elic-eon, line 502: Do we need to normalize `user_id` or `email` here?

metaphor/redshift/extractor.py
```diff
@@ -184,6 +184,8 @@ def _process_record(self, access_event: AccessEvent):
             )
 
         if sql:
+            user_id = access_event.usename
+            email = self._query_log_config.username_to_email.get(user_id)
```
elic-eon, line 188: Also here

metaphor/postgresql/extractor.py
usefulalgorithm: Good catch, `user_id` probably doesn't need normalization, neither Redshift nor Postgres say anything about normalizing the user name; but `email` should be normalized so that it can match what we store in Metaphor.

elic-eon (approved):

.github/workflows/ci.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           poetry run isort --check .
           poetry run mypy . --explicit-package-bases
           poetry run bandit -r . -c pyproject.toml
-
+2
```
elic-eon, line 52: `2`?

usefulalgorithm: typo orz

elic-eon (approved):

PR #991, opened 2024-09-30
[sc-29051] MongoDB crawler

usefulalgorithm:
### 🤔 Why?

This is the implementation of the MongoDB crawler.

### 🤓 What?

- Implemented the MongoDB crawler.
- Added MongoDB crawler readme.
- Added tests.

### 🧪 Tested?

- Added unit test
  - This test the whole connector by setting up a MongoDB container, importing some sample data from MongoDB's official sample data, and running the crawler against the container.
- Crawler was run locally.

### ☑️ Checks


metaphor/mongodb/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Setup
+
+### Required Configurations
+
+```yaml
+uri: <uri> # The connection URI.
+auth_mechanism: <auth_mechanism> # The authentication mechanism. Allowed values are "GSSAPI", "MONGODB-CR", "MONGODB-OIDC", "MONGODB-X509", "MONGODB-AWS", "PLAIN", "SCRAM-SHA-1", "SCRAM-SHA-256", "DEFAULT". Default is "DEFAULT".
+tls: <boolean> # Whether to set TLS when connecting to MongoDB. Default is False.
+
+infer_schema_sample_size: <int> # Number of documents to sample in a collection in order to infer the schema. Set this to `null` to disable sampling and use all documents in the collections. Default is 1000.
```
alyiwang, line 14: I would imagine there are customers that don't want the crawler to sample any real data, i.e. don't infer at all. Do they set this to 0 in that case?

metaphor/mongodb/extractor.py
```diff
@@ -0,0 +1,121 @@
...
+        pipeline = []
+        if self._sample_size:
+            pipeline.append(
+                {
+                    "$sample": {
+                        "size": self._sample_size,
+                    }
+                }
+            )
+        docs = collection.aggregate(pipeline).to_list()
```
alyiwang, line 69: Does this work if sample size is 0?

metaphor/common/infer_schema.py
```diff
@@ -0,0 +1,191 @@
...
+    if len(field_types.keys()) == 1:
+        field_type = next(iter(field_types))
+    elif set(field_types.keys()) == {int, float}:
+        field_type = float
+    elif set(field_types.keys()) == {datetime, str}:
+        field_type = datetime
+    else:
+        logger.warning(
+            f"Multiple types found in field: field = {field_path}, types = {list(field_types.keys())}"
+        )
```
alyiwang, line 88: If multiple types, do we want to guess the first one or leave it empty?

alyiwang (commented):
Otherwise, LGTM
Great work!

alyiwang (approved):
LGTM

metaphor/mongodb/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Setup
+
+### Required Configurations
+
+```yaml
+uri: <uri> # The connection URI.
+auth_mechanism: <auth_mechanism> # The authentication mechanism. Allowed values are "GSSAPI", "MONGODB-CR", "MONGODB-OIDC", "MONGODB-X509", "MONGODB-AWS", "PLAIN", "SCRAM-SHA-1", "SCRAM-SHA-256", "DEFAULT". Default is "DEFAULT".
+tls: <boolean> # Whether to set TLS when connecting to MongoDB. Default is False.
+
+infer_schema_sample_size: <int> # Number of documents to sample in a collection in order to infer the schema. Set this to `null` to disable sampling and use all documents in the collections. To disable schema inference altogether, set this to 0. Default is 1000.
```
alyiwang, line 14: On second thought, do you think it's more natural to have `NULL` for no infer and 0 for using all records? @mars-lan

usefulalgorithm: I think either should be alright

usefulalgorithm: Gonna merge this first, if we want to change it we can do that in a separate PR.

PR #993, opened 2024-10-01
Extract iceberg table

elic-eon:
### 🤔 Why?

To support the Snowflake iceberg table.

### 🤓 What?

- Update the readme to give the crawler permission to extract the iceberg tables.
- Fill extra information for iceberg table

### 🧪 Tested?

Create an iceberg table and the crawler can extract it.

### ☑️ Checks


tests/mongodb/expected_datasets.json
mars-lan, line 1: ?

elic-eon: @usefulalgorithm Is it related to #991 ? After `pytest`, this file is changed

elic-eon: Fixed, but this JSON will be modified

metaphor/snowflake/extractor.py
```diff
@@ -709,6 +712,49 @@ def _to_dataset_eid(x: str) -> str:
...
+            return
+
+        count = 0
+        for entry in cursor:
+            (
+                table_name,
+                external_volume_name,
+                iceberg_table_type,
+            ) = (
+                entry[1],
```
mars-lan, line 732: I believe you can use `row[<col_nme>]` instead of index-based access?

elic-eon: It's tuple here, because we can't specifying columns for this query

usefulalgorithm (approved):

PR #1000, opened 2024-10-07
[sc-18790] Deprecate Snowflake lineage crawler

usefulalgorithm:
### 🤔 Why?

We're already doing lineage parsing, no need to have a separate Snowflake lineage parsing connector.

### 🤓 What?

- Modified main crawler to include lineage crawler behavior:
  - Moved `fetch_direct_object_dependencies` logic into the main crawler.
  - The TLL contained in query logs will be handled in a separate PR.
- Deleted the implementation of Snowflake lineage crawler, and marked it for deletion once 0.15.0 is out.
- Dropped tests.
- Modified main crawler tests
- Updated readme.
- Added MongoDB crawler to main readme.

### 🧪 Tested?

- Run locally with stage configs.
- Updated unit test.

### ☑️ Checks


rishimo:
pr 1000 🎉

mars-lan:
CI failing?

metaphor/snowflake/extractor.py
```diff
@@ -1048,3 +1068,81 @@ def _fetch_schemas_comment(self, cursor: SnowflakeCursor, database: str) -> None
...
+                transformation,
+            )
+
+    def _fetch_direct_object_dependencies(self, cursor: SnowflakeCursor):
+        logger.info("Fetching direct object dependencies")
+        cursor.execute(
+            """
+            SELECT REFERENCED_DATABASE, REFERENCED_SCHEMA, REFERENCED_OBJECT_NAME, REFERENCED_OBJECT_DOMAIN,
+                REFERENCING_DATABASE, REFERENCING_SCHEMA, REFERENCING_OBJECT_NAME, REFERENCING_OBJECT_DOMAIN
+            FROM SNOWFLAKE.ACCOUNT_USAGE.OBJECT_DEPENDENCIES;
```
mars-lan, line 1121: Must filter by date or else it'll be exceedingly slow (see original code).

usefulalgorithm: Don't think that's possible, the OBJECT_DEPENDENCIES view has no datetime column for me to filter: https://docs.snowflake.com/en/sql-reference/account-usage/object_dependencies

The original code doesn't have it either
截圖 2024-10-09 凌晨12 07 41

mars-lan: Is this the correct view to use? We used https://docs.snowflake.com/en/sql-reference/account-usage/access_history before?

mars-lan: Argh nvm, this is for view lineage.

metaphor/snowflake/extractor.py
```diff
@@ -179,6 +186,17 @@ async def extract(self) -> Collection[ENTITY_TYPES]:
             if self._config.collect_tags:
                 self._fetch_tags(cursor)
 
+            # Fetch query logs now since we need the dataset entity upstream lineage here,
+            # and in `collect_query_logs` we can't touch `self._datasets` anymore.
+            self._query_logs = (
```
mars-lan, line 191: I don't think we can hold this in memory. We'll most likely need to process it in a streaming fashion?

metaphor/snowflake/extractor.py
```diff
@@ -179,6 +186,9 @@ async def extract(self) -> Collection[ENTITY_TYPES]:
             if self._config.collect_tags:
                 self._fetch_tags(cursor)
 
+            # Fetch direct object dependencies
```
mars-lan, line 189:
```suggestion
            # Fetch direct object dependencies for view lineage
```

mars-lan (commented):
Otherwise LGTM

mars-lan (approved):

PR #1004, opened 2024-10-09
Support Postgres log extraction with log_duration disabled

elic-eon:
### 🤔 Why?

Since `log_duration` does not always default to `true`, we should be able to extract query history without duration.

### 🤓 What?

- Add `log_duration_enabled` flag to default to `false` to extract log without duration
- Add `filter_pattern` to speed up query log extraction
- fix skipping `execute` statement and add testcase

### 🧪 Tested?

Update unit tests, verified against dev postgres instance.

### ☑️ Checks

### DIFF
--- a/metaphor/common/sql/process_query/process_query.py
+++ b/metaphor/common/sql/process_query/process_query.py
@@ -22,6 +22,15 @@ def _is_insert_values_into(expression: Expression) -> bool:
     )
 
 
+ALLOW_EXPRESSION_TYPES = (
+    exp.Alter,
+    exp.DDL,
+    exp.DML,
+    exp.Merge,
+    exp.Query,
+)
+
+
 def process_query(
     sql: str,
     data_platform: DataPlatform,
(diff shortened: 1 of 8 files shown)


metaphor/common/sql/process_query/process_query.py
```diff
@@ -82,6 +82,7 @@ def process_query(
         isinstance(expression, exp.Show)
         or isinstance(expression, exp.Set)
         or isinstance(expression, exp.Rollback)
+        or isinstance(expression, exp.Commit)
```
usefulalgorithm, line 85: Maybe it's easier to opt-in rather than opt-out.

Change this to
```python
if not isinstance(expression, (exp.Alter, exp.Query, exp.DML, exp.DDL)):
```

This covers ALTER, SELECT, CREATE, INSERT, COPY, etc. Not sure if there are other things we care about though.

elic-eon: @mars-lan WDYT about opt-in?

mars-lan (approved):

metaphor/common/sql/process_query/process_query.py
```diff
@@ -78,11 +87,7 @@ def process_query(
     if config.ignore_command_statement and isinstance(expression, exp.Command):
         return None
 
-    if (
-        isinstance(expression, exp.Show)
-        or isinstance(expression, exp.Set)
-        or isinstance(expression, exp.Rollback)
-    ):
+    if not isinstance(expression, ALLOW_EXPRESSION_TYPES):
```
usefulalgorithm, line 90: then we don't need `ignore_command_statement` right?

elic-eon: Let's deprecate this later because we will need to make a breaking change to the configuration.

elic-eon: #858

PR #1009, opened 2024-10-16
[sc-29305] Merge dbt cloud metadata before writing to file

usefulalgorithm:
### 🤔 Why?

It is possible for a single dbt model to be referenced by tests in jobs in different projects. To avoid writing to different virtual view MCEs, we want to write to the same virtual view and only update the necessary bits.

### 🤓 What?

- Pass crawler-wide entities to dbt cloud sub-parsers.
- Added dedup method.

### 🧪 Tested?

- Updated unit test.
- Ran the crawler locally.

The model `jaffle_shop.stg_supplies` has two tests:
截圖 2024-10-16 晚上7 45 02

This model is referenced in 2 projects: `jaffle_shop` and `jaffle_shop_postgres` (basically a clone of the first one, just run on top of postgres.) Both projects are included in my crawler config.

When we run the extractor, in the model there's only 1 `not_null_stg_supplies_supply_uuid` test.

But when we look at data quality for the referenced dataset, we see there are two monitors. This is because the test is being run in two different projects:

截圖 2024-10-16 晚上7 45 13

### ☑️ Checks

### DIFF
--- a/metaphor/dbt/cloud/extractor.py
+++ b/metaphor/dbt/cloud/extractor.py
@@ -3,7 +3,7 @@ from typing import Collection, Dict, Set
 import httpx
 
 from metaphor.common.base_extractor import BaseExtractor
-from metaphor.common.event_util import ENTITY_TYPES, EventUtil
+from metaphor.common.event_util import ENTITY_TYPES
 from metaphor.common.logger import get_logger
 from metaphor.dbt.cloud.client import DbtAdminAPIClient
 from metaphor.dbt.cloud.config import DbtCloudConfig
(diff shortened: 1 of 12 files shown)


metaphor/dbt/cloud/extractor.py
```diff
@@ -38,7 +40,13 @@ def __init__(self, config: DbtCloudConfig):
         self._discovery_api_url = config.discovery_api_url
 
         self._project_accounts: Dict[int, str] = {}
-        self._entities: List[ENTITY_TYPES] = []
+
+        self.datasets: Dict[str, Dataset] = {}
+        self.virtual_views: Dict[str, VirtualView] = {}
+        self.metrics: Dict[str, Metric] = {}
```
alyiwang, line 46: nit: use `self._dataset` etc to be consistent

alyiwang (approved):
Otherwise, LGTM!

elic-eon (approved):
LGTM

PR #1017, opened 2024-10-22
Redshift profiler missing tables [sc-29514]

alyiwang:
### 🤔 Why?

The `include_view` config was used incorrectly, doing the exact opposite.

### 🤓 What?

- fix the dataset filtering based on `include_view`, also excluding `external`, `stream`, `snapshot` types
- also remove the dataset entities with empty field statistics from the MCE

### 🧪 Tested?

tested with `include_view` set to True and False, verified the results.

### ☑️ Checks

### DIFF
--- a/metaphor/postgresql/profile/extractor.py
+++ b/metaphor/postgresql/profile/extractor.py
@@ -1,6 +1,6 @@
 import asyncio
 import traceback
-from typing import Collection, Iterable, List
+from typing import Collection, List
 
 try:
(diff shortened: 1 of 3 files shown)


mars-lan (commented):
Can we also add some tests to prevent similar regression in the future?

alyiwang:
> Can we also add some tests to prevent similar regression in the future?

added

mars-lan (approved):

PR #1022, opened 2024-10-25
Refactor Unity Catalog to fetch catalog/schema/table metadata from System tables

mars-lan:
### 🤔 Why?

Fetching metadata from [system.information_schema](https://docs.databricks.com/en/sql/language-manual/sql-ref-information-schema.html) is preferred over individual catalog's `information_schema`, as the former doesn't require granting `SELECT` permission to all tables.

Note that `SELECT` permission is still required to retrieve a table's properties & last refresh date as they're not available from various `system` tables.

### 🤓 What?

- Replace all REST API calls with queries against `system.information_schema`. The only exception is IAM-related metadata, which is only available via REST APIs.
- Consolidate all queries into `queries.py` to make testing easier and better organization.
- Update databricks dependencies to the latest version

### 🧪 Tested?

Tested end-to-end against a product instance. Verified that the before & after MCEs are the same except:

####  Columns
- `precision` for date-related types is set.
- `nullable` & `tag` are set.
-  Use upper case for `nativeType`. 

#### Table properties
- Missing Spark-related properties as they're not available via the `SHOW TBLPROPERTIES` command. 
- Use the property's raw values instead of JSON-encoded strings.   

#### Hierarchy:
- Set `createdAtSource`, `createdBy`, `lastUpdated`, `updatedBy` in `sourceInfo`
- Set `systemContacts` & `systemDescription`

### ☑️ Checks


metaphor/unity_catalog/extractor.py
```diff
@@ -331,79 +265,91 @@ def _init_dataset(self, table_info: TableInfo) -> Dataset:
...
+        field = build_schema_field(
+            column_name=column_info.column_name,
+            field_type=column_info.data_type,
+            description=column_info.comment,
+            nullable=column_info.is_nullable,
+            precision=safe_float(column_info.data_precision),
+        )
+
+        field.tags = [
+            f"{tag.key}={tag.value}" if tag.key else tag.value
```
usefulalgorithm, line 347: In Snowflake crawler there's `_stringify_system_tag` (in `metaphor/snowflake/utils.py`), move that method to a common utils files and use that?

metaphor/unity_catalog/extractor.py
```diff
@@ -784,51 +586,81 @@ def _init_volume(self, volume: VolumeInfo):
...
+                type=UnityCatalogVolumeType[volume.volume_type],
                 volume_files=[],
                 storage_location=volume.storage_location,
             ),
         )
         dataset.entity_upstream = EntityUpstream(source_entities=[])
 
-        dataset.system_tags = SystemTags(tags=[])
+        dataset.system_tags = SystemTags(
+            tags=[
```
usefulalgorithm, line 597: Good catch

metaphor/unity_catalog/queries.py
```diff
@@ -0,0 +1,719 @@
...
+}
+"""These are the operations that do not modify actual data."""
+
+
+def to_tags(tags: Optional[List[Dict]]) -> List[Tag]:
+    return (
+        [Tag(key=tag["tag_name"], value=tag["tag_value"]) for tag in tags]
+        if tags
+        else []
+    )
```
usefulalgorithm, line 45:
```suggestion
    return [Tag(key=tag["tag_name"], value=tag["tag_value"]) for tag in tags or []]
```

metaphor/unity_catalog/extractor.py
mars-lan: Sounds good. Will create a follow-up PR for that to avoid expanding the surface area of this PR.

usefulalgorithm (dismissed):

usefulalgorithm (approved):

PR #1026, opened 2024-10-31
[sc-29648] Use `init_virtual_view` to get or create the source in `depends_on`

usefulalgorithm:
### 🤔 Why?

If a source model isn't referenced in the run, when we parse the `depends_on` of the node that depends on the source model it will break.

Using `init_virtual_view` to get the model or create it when it does not exist yet solves the issue.

### 🤓 What?

`virtual_views[n]` -> `init_virtual_views(virtual_views, n, name_getter)`.

### 🧪 Tested?

Testing in progress.

### ☑️ Checks

### DIFF
--- a/metaphor/dbt/cloud/parser/common.py
+++ b/metaphor/dbt/cloud/parser/common.py
@@ -1,9 +1,15 @@
 import json
-from typing import Dict, List, Union
+from typing import Callable, Dict, List, Union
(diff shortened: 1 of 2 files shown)


mars-lan (dismissed):

alyiwang (approved):
LGTM

PR #1028, opened 2024-11-01
[sc-29650] NoneType is not convertible to int

usefulalgorithm:
### 🤔 Why?

I added `int()` to cast `parameters.get("numRows")` into an int to appease mypy, but it would cast a None into int which breaks the crawler.

### 🤓 What?

Remove `int()` cast for `parameters.get("numRows")`.

### 🧪 Tested?

Run crawler locally against our QA instance.

### ☑️ Checks


elic-eon (approved):

metaphor/glue/extractor.py
```diff
@@ -99,7 +99,7 @@ def _get_tables(self, database: str):
                 )
                 table_type = table.get("TableType")
                 parameters = table.get("Parameters")
-                row_count = int(parameters.get("numRows")) if parameters else 0
```
mars-lan, line 102: Use `safe_int()`? @usefulalgorithm

PR #1032, opened 2024-11-04
Add rate limit config for BigQuery crawler

elic-eon:
### 🤔 Why?

Getting HTTP 429 response while extracting query log.

### 🤓 What?

Tried adding a rate limiter to the actual API call but the API we used `list_entries` internally called the API endpoint, and we have no control over it. The solution is to sleep a while after one page of entries.

### 🧪 Tested?

Verified against prod and dev env 

### ☑️ Checks


mars-lan (commented):
A better approach is to handle 429 gracefully by sleeping for say 1s after the server returns 429.

elic-eon:
> A better approach is to handle 429 gracefully by sleeping for say 1s after the server returns 429.

We can't do that either, we are not able to get `page_token` from the API. unless we call the API directly

metaphor/bigquery/extractor.py
```diff
@@ -306,15 +308,28 @@ def _fetch_query_logs(self, project_id: str) -> Iterator[QueryLog]:
...
                 if log:
                     fetched_logs += 1
                     yield log
 
+            if count % self._query_log_fetch_size == 0:
+                current_time = time.time()
+                elapsed_time = current_time - last_time
+                wait_time = (60 / self._rate_limit) - elapsed_time
+                if wait_time > 0:
+                    time.sleep(wait_time)
```
usefulalgorithm, line 331: Don't really follow the logic here... what's the meaning of `wait_time` here?

- Why are we dividing 60 by `self._rate_limit`, which actually stands for max. requests per minute?
- If there are a lot of query entries, eventually `elapsed_time` will always be larger than `(60 / self._rate_limit)`, since `last_time` is never updated, and `wait_time` will never be above 0.

elic-eon: so every call should at least spend `60 secs / RPM` seconds

metaphor/bigquery/config.py
```diff
@@ -62,6 +64,9 @@ class BigQueryQueryLogConfig:
         default_factory=lambda: ProcessQueryConfig()
     )
 
+    # Maximum allowed requests per minute to the log entries API
+    rate_limit: int = DEFAULT_RATE_LIMIT
```
usefulalgorithm, line 68: Maybe just call this `max_requests_per_minute` so it's easier to understand.
```suggestion
    max_requests_per_minute: int = DEFAULT_MAX_REQUESTS_PER_MINUTE
```

mars-lan:
> > A better approach is to handle 429 gracefully by sleeping for say 1s after the server returns 429.
> 
> We can't do that either, we are not able to get `page_token` from the API. unless we call the API directly

I'm confused. Can't you repeat the call after X seconds using the previous page_token after receiving a 429?

elic-eon:
> > > A better approach is to handle 429 gracefully by sleeping for say 1s after the server returns 429.
> > 
> > 
> > We can't do that either, we are not able to get `page_token` from the API. unless we call the API directly
> 
> I'm confused. Can't you repeat the call after X seconds using the previous page_token after receiving a 429?

We can't get any `page_token` so we can't continue to get the remaining pages

usefulalgorithm (commented):
LGTM, if @mars-lan thinks it's good then go go go

mars-lan:
Chatted with @elic-eon offline and the approach makes sense. Please link to the doc for the default rate limit for posterity.

metaphor/bigquery/config.py
```diff
@@ -15,7 +15,7 @@
 # See https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/list
 DEFAULT_QUERY_LOG_FETCH_SIZE = 1000
 
-DEFAULT_MAX_REQUESTS_PER_MINUTE = 60
+DEFAULT_MAX_REQUESTS_PER_MINUTE = 59
```
mars-lan, line 18: Link to https://cloud.google.com/logging/quotas

mars-lan (approved):

PR #1035, opened 2024-11-06
BigQuery crawler filtering not working with [sc-29743]

alyiwang:
### 🤔 Why?

Turns out it's a misconfiguration caused by the documentation. Added more detail about the big query crawler `filter` config. 

Also discovered a bug about the snowflake filtering. if the `include` database contains wildcard, it is used as a plain text and causes all databases to be bypassed.

### 🤓 What?

- update readme of big query crawler filter
- fix snowflake `fetch_database`

### 🧪 Tested?

tested against big query and snowflake instances with wildcard in filter config, and verified the results.

### ☑️ Checks

### DIFF
--- a/metaphor/bigquery/README.md
+++ b/metaphor/bigquery/README.md
@@ -112,6 +112,7 @@ See [Output Config](../common/docs/output.md) for more information.
 #### Filtering
 
 See [Filter Config](../common/docs/filter.md) for more information on the optional `filter` config.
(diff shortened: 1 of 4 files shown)


mars-lan (approved):

usefulalgorithm (approved):

usefulalgorithm (changes requested):
Fix conflict and it's good

mars-lan:
Looks like the real issue is that BQ crawler doesn't support wildcard filter for projects?

alyiwang:
> Looks like the real issue is that BQ crawler doesn't support wildcard filter for projects?

The BQ crawler requires the project IDs to be explicitly listed in the `project_ids` config

usefulalgorithm (approved):
go go go

PR #1037, opened 2024-11-06
Fix OpenAPI hierarchy

elic-eon:
### 🤔 Why?

APIs without tags have empty hierarchy but they should belongs to top-level OpenAPI hierarchy

### 🤓 What?

- Fix `directories` logic

### 🧪 Tested?

Add unit tests.

### ☑️ Checks


usefulalgorithm (approved):
👍

PR #1040, opened 2024-11-12
[sc-29773] Handle tables without columns

usefulalgorithm:
### 🤔 Why?

Our crawler did not take into account for the case when `row["columns"]` is a `None`.

### 🤓 What?

- Fix NoneType not iterable error
- Refactor to make testing easier
- Added tests

### 🧪 Tested?

- Added unit test.
- Tested against stage environment.

### ☑️ Checks


elic-eon (approved):
LGTM

PR #1045, opened 2024-11-19
Fix rate limiting issues with Power BI admin APIs

mars-lan:
### 🤔 Why?

Most of the Power BI [Admin APIs](https://learn.microsoft.com/en-us/rest/api/power-bi/admin) have a rate limit of 200 requests/hour. We reach the limit quickly when crawling a large number of workspaces.

### 🤓 What?

Replace the following per-workspace API calls with global API calls. These global APIs appear to be equivalent and are able to scale to thousands of items without any issues.
- [GetDatasetsInGroupAsAdmin](https://learn.microsoft.com/en-us/rest/api/power-bi/admin/datasets-get-datasets-in-group-as-admin) => [GetDatasetsAsAdmin](https://learn.microsoft.com/en-us/rest/api/power-bi/admin/datasets-get-datasets-as-admin)
- [GetReportsInGroupAsAdmin](https://learn.microsoft.com/en-us/rest/api/power-bi/admin/reports-get-reports-in-group-as-admin) => [GetReportsAsAdmin](https://learn.microsoft.com/en-us/rest/api/power-bi/admin/reports-get-reports-as-admin)
- [GetDashboardsInGroupAsAdmin](https://learn.microsoft.com/en-us/rest/api/power-bi/admin/dashboards-get-dashboards-in-group-as-admin) => [GetDashboardsAsAdmin](https://learn.microsoft.com/en-us/rest/api/power-bi/admin/dashboards-get-dashboards-as-admin)

Unfortunately, there is no global version of [GetTilesAsAdmin](https://learn.microsoft.com/en-us/rest/api/power-bi/admin/dashboards-get-dashboards-as-admin), so we need to fallback to the non-admin version [Get Tiles In Group](https://learn.microsoft.com/en-us/rest/api/power-bi/dashboards/get-tiles-in-group). This would require the service principal to be added to each workspace, which is already a requirement for getting refresh schedule, data sources, etc.

### 🧪 Tested?

Verified the before/after MCE against production instances.

### ☑️ Checks

### DIFF
--- a/metaphor/power_bi/extractor.py
+++ b/metaphor/power_bi/extractor.py
@@ -63,6 +63,9 @@ from metaphor.power_bi.config import PowerBIRunConfig
 from metaphor.power_bi.graph_api_client import GraphApiClient
 from metaphor.power_bi.models import (
     PowerBIApp,
+    PowerBIDashboard,
+    PowerBIDataset,
+    PowerBIReport,
     PowerBISubscription,
     PowerBiSubscriptionUser,
     WorkspaceInfo,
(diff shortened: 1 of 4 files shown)


alyiwang (approved):
LGTM

PR #1049, opened 2024-12-12
Looker crawler is failed to load model [sc-30007]

alyiwang:
### 🤔 Why?

Some Looker model may have invalid connection and raise value error

### 🤓 What?

- add config to allow bypassing certain model files from parser

### 🧪 Tested?

Tested on customer environment 

### ☑️ Checks

### DIFF
--- a/metaphor/looker/lookml_parser.py
+++ b/metaphor/looker/lookml_parser.py
@@ -4,6 +4,7 @@ import logging
 import operator
 import os
 from dataclasses import dataclass
+from fnmatch import fnmatch
 from typing import Dict, List, Optional, Set, Tuple
 
 try:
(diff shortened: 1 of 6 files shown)


mars-lan (approved):

mars-lan (changes requested):
Prefer to fail loudly so the error is not hidden to the users.

metaphor/looker/lookml_parser.py
```diff
@@ -626,7 +636,12 @@ def parse_project(
     virtual_views = []
 
     for model_path in glob.glob(f"{base_dir}/**/*.model.lkml", recursive=True):
+        if _is_ignored_model_file(model_path, ignored_model_files):
```
mars-lan, line 639: Might be better using a path relative to `bsae_dir` instead of the absolute path since this is likely run in a CI environment with an unknown base directory?

alyiwang: Good point, updadted

mars-lan (approved):

PR #1050, opened 2024-12-20
Add pagination to Monte Carlo getMonitors call

mars-lan:
### 🤔 Why?

MC's GraphQL API places limit the max response size to 6 MB. The current non-paginated [getMonitors](https://apidocs.getmontecarlo.com/#query-getMonitors) call can exceed that limit when is a large number monitors configured.

### 🤓 What?

As titled.

### 🧪 Tested?

Verified that before & after output are identical against a production instance.

### ☑️ Checks

### DIFF
--- a/metaphor/monte_carlo/extractor.py
+++ b/metaphor/monte_carlo/extractor.py
@@ -1,5 +1,4 @@
 import re
-import traceback
 from typing import Collection, Dict, List
 
 from dateutil import parser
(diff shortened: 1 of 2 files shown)


metaphor/monte_carlo/extractor.py
```diff
@@ -102,11 +101,15 @@ def _fetch_monitors(self) -> None:
         See https://apidocs.getmontecarlo.com/#query-getMonitors
         """
 
-        try:
-            monitors = self._client(
+        offset = 0
+        limit = 200
```
rishimo, line 105: How is this size determined? Should this be configurable?

alyiwang (approved):
LGTM

metaphor/monte_carlo/extractor.py
mars-lan: It's a number small enough to avoid exceeding the size limit, but large enough to not cause significant slowdown. We could certainly make it configurable, but not sure users will want to mess with this value.

PR #1052, opened 2024-12-24
Rewrite dbt cloud crawler using discovery API

alyiwang:
### 🤔 Why?

Should use the dbt discovery API `environment` endpoint to fetch most of the metadata instead of `jobs`. This simplifies the steps and has more complete lineage info.

### 🤓 What?

- rewrite the whole dbt cloud crawler using discovery API `environment` endpoint, as well as the admin API to get all projects and environments 
- Update required configs, `job_ids` no longer used

### 🧪 Tested?

Tested against metaphor dbt instance. File diff with MCE generated using previous crawlers. The results are mostly the same.

> Known differences
- now able to retrieve dbt model columns
- `dbtModel.sourceModels` not longer filled as it's an deprecated field in favor of `entityUpstream`
- docsUrl no longer filled, as the previous format `https://cloud.getdbt.com/accounts/123/jobs/146/docs/#!/xxx` no longer supported by dbt
- `test.sql` not available currently, can get it from top-level `environment.tests` endpoint later on
- metrics label, dimensions, filters, timeGrains are not available right now, but `formula` is now available.

### ☑️ Checks


mars-lan (approved):
Great work! Thanks for the refactoring.

PR #1054, opened 2024-12-31
Fix various dbt cloud connector issues

mars-lan:
### 🤔 Why?

A few issues were observed when running https://github.com/MetaphorData/connectors/pull/1052 against more complex dbt projects.

### 🤓 What?

- Reduce the discovery API page size to avoid hitting HTTP 502 due to large responses.
- Make `connection_id` & `connection` fields optional, despite the latter being [documented as required](https://docs.getdbt.com/dbt-cloud/api-v3#/operations/List%20Projects).
- Ignore exceptions (e.g. missing `connection` field or unexpected connection type) thrown when parsing projects/environments to avoid crashing the entire crawler.

### 🧪 Tested?

Verified against a few complex production environments.

### ☑️ Checks

### DIFF
--- a/metaphor/dbt/cloud/extractor.py
+++ b/metaphor/dbt/cloud/extractor.py
@@ -68,7 +68,10 @@ class DbtCloudExtractor(BaseExtractor):
         projects = self._client.list_projects()
         for project in projects:
             if not self._project_ids or project.id in self._project_ids:
-                await self._extract_project(project)
+                try:
+                    await self._extract_project(project)
+                except Exception as e:
+                    logger.error(f"Error extracting project {project.id}: {e}")
 
         datasets = [d for d in self._datasets.values() if should_be_included(d)]
         views = [v for v in self._virtual_views.values() if should_be_included(v)]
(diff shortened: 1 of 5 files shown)


metaphor/dbt/cloud/extractor.py
```diff
@@ -68,7 +68,10 @@ async def extract(self) -> Collection[ENTITY_TYPES]:
         projects = self._client.list_projects()
         for project in projects:
             if not self._project_ids or project.id in self._project_ids:
-                await self._extract_project(project)
+                try:
+                    await self._extract_project(project)
+                except Exception as e:
```
alyiwang, line 73: Maybe we want to try catch at the environment level instead of at the higher project level

alyiwang (commented):
Otherwise lgtm

metaphor/dbt/cloud/extractor.py
mars-lan: All the errors I've observed so far are at project-level, e.g. missing `connection` info or unexpected connection type.

alyiwang (approved):
LGTM

PR #1059, opened 2025-01-09
QuickSight crawler unsupported dataset type [sc-30088]

alyiwang:
### 🤔 Why?

The quickSight DescribeDataSet API may throw error for some dataset type such as file formats (csv, s3 file, etc)

### 🤓 What?

- add try catch to not fail the entire crawler when DescribeDataSet error is thrown
- dump the list API response for debugging

### 🧪 Tested?

tested against Metaphor instance, but need to test on customer env to log the issue

### ☑️ Checks


mars-lan (approved):
Thanks for the quick fix.

PR #1067, opened 2025-01-16
update quicksight lineage

alyiwang:
### 🤔 Why?

The upstream snowflake URL sometimes contains the default region `us-west-2` that needs to be stripped to match the account name from Snowflake crawler.

Also in some cases, when the lineage parser needs to find an existing dataset, it used quicksight dataset_id but the dictionary used arn as key, should all use id instead.

### 🤓 What?

- update snowflake account to remove default region
- update MCE dictionary to use id as key instead of arn
- refactor lineage parsing
- in the Virtual View and Dashboard logical ID, use the Quicksight ID instead of ARN as name.

### 🧪 Tested?

tested on Metaphor instance

### ☑️ Checks

### DIFF
--- a/metaphor/common/snowflake.py
+++ b/metaphor/common/snowflake.py
@@ -1,8 +1,9 @@
 PRIVATE_LINK_SUFFIX = ".privatelink"
 SNOWFLAKE_HOST_SUFFIX = ".snowflakecomputing.com"
+SNOWFLAKE_DEFAULT_REGION = ".us-west-2"
 
 
-def normalize_snowflake_account(host: str) -> str:
+def normalize_snowflake_account(host: str, remove_default_region: bool = False) -> str:
     """
     Normalize different variations of Snowflake account.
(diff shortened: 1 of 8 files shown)


metaphor/common/snowflake.py
```diff
@@ -16,6 +17,9 @@ def normalize_snowflake_account(host: str) -> str:
 
     # Strip PrivateLink suffix
     if host.endswith(PRIVATE_LINK_SUFFIX):
-        return host[: -len(PRIVATE_LINK_SUFFIX)]
+        host = host[: -len(PRIVATE_LINK_SUFFIX)]
+
+    if remove_default_region and host.endswith(SNOWFLAKE_DEFAULT_REGION):
```
mars-lan, line 22: Add comment

alyiwang: added

mars-lan (approved):

PR #1071, opened 2025-02-03
Filter out non-active MC monitors

mars-lan:
### 🤔 Why?

Monte Carlo crawler has been incorrectly reporting non-active monitors as DQ errors, and not treating breached monitors as errors.

### 🤓 What?

- Filter out non-active monitors (i.e.  `consolidatedStatusTypes = ENABLED`) when calling [get_monitors](https://apidocs.getmontecarlo.com/#query-getMonitors).
- Use the `breached` field instead of `monitorStatus` field to determine if there's a DQ issue.
- Drop exception process logic since it's only for monitors that failed to execute (i.e. `consolidatedStatusTypes = ERROR`).

### 🧪 Tested?

### ☑️ Checks

### DIFF
--- a/metaphor/monte_carlo/config.py
+++ b/metaphor/monte_carlo/config.py
@@ -1,4 +1,3 @@
-from dataclasses import field
 from typing import List, Optional
 
 from pydantic.dataclasses import dataclass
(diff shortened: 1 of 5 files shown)


alyiwang (approved):
LGTM
Thanks for fixing!

